PR #30, opened 2023-02-02
Computation of gradient with respect to the input

David-Schnurr:
### Change Description
I added the functionality to compute a gradient with respect to X_train and X_test. Thereby, the gradient can only be computed if X_train and X_test do not contain nans.

### Standard Qs (please answer all)
**Did all tests pass?**
Yes, they passed on my Pc.

**Do you break behavior, if so why and did you communicate it?**
No behavior should be broken.

**If you did add dependencies, did you add them to `requirements.txt`?**
No new dependencies were added.

**Did you ping someone to review your PR?**
Yes, Samuel

Note: I think it would be a good idea to squash and merge, because I made many small commits.


tabpfn/scripts/transformer_prediction_interface.py
```diff
@@ -421,7 +455,8 @@ def get_preprocess(i):
                 #                        preprocess_input(eval_xs_, preprocess_transform='none')], -1)
                 #eval_xs_ = torch.flatten(torch.swapaxes(eval_xs_, -2, -1), -2)
             else:
-                eval_xs_ = preprocess_input(eval_xs_, preprocess_transform=preprocess_transform_configuration)
+                eval_xs_ = preprocess_input(eval_xs_, preprocess_transform=preprocess_transform_configuration,
```
SamuelGabriel: You can remove the `combine_preprocessing` and then I believe this to be the only call to pre_process_input, right? Then we can take the detach out of preprocess_input, have it here, and remove the `no_grad` argument. Could you do that?

tabpfn/scripts/transformer_prediction_interface.py
```diff
@@ -466,12 +501,13 @@ def get_preprocess(i):
...
         #output_ = predict(eval_xs, eval_ys, style_, preprocess_transform_)
-        if not average_logits:
+        if not average_logits and not return_logits:
+            # transforms every ensemble_configuration into a probability -> equal contribution of every configuration
             output_ = torch.nn.functional.softmax(output_, dim=-1)
         output = output_ if output is None else output + output_
 
-    output = output / len(ensemble_configurations)
-    if average_logits:
+    output /= len(ensemble_configurations)
```
SamuelGabriel: I saw this do an inlace operation before which breaks gradient flow for some PyTorch version. Here is a little blog post on this issue. https://lernapparat.de/pytorch-inplace

So probably most save to go back to the old line

tabpfn/utils.py
```diff
@@ -172,19 +171,30 @@ def nan_handling_missing_for_no_reason_value(set_value_to_nan=0.0):
...
+    value = torch.where(mask, x, torch.full_like(x, 0)).sum(dim=axis)
+    mean = value / num
+    mean_broadcast = torch.repeat_interleave(mean.unsqueeze(axis), x.shape[axis], dim=axis)
+    quadratic_difference_from_mean = torch.square(torch.where(mask, mean_broadcast - x, torch.full_like(x, 0)))
+    return mean, torch.sqrt(torch.sum(quadratic_difference_from_mean, dim=axis) / (num - 1))
+
 def torch_nanmean(x, axis=0, return_nanshare=False):
-    num = torch.where(torch.isnan(x), torch.full_like(x, 0), torch.full_like(x, 1)).sum(axis=axis)
-    value = torch.where(torch.isnan(x), torch.full_like(x, 0), x).sum(axis=axis)
+    num = torch.where(torch.isnan(x), torch.full_like(x, 0), torch.full_like(x, 1)).sum(dim=axis)
```
SamuelGabriel: dim is the standard keyword, why did it even work with axis before!? :D

tabpfn/utils.py
```diff
@@ -172,19 +171,30 @@ def nan_handling_missing_for_no_reason_value(set_value_to_nan=0.0):
...
+    """
+    Returns the mean and std of a torch tensor and only considers the elements, where the mask is true
+    """
+    num = torch.where(mask, torch.full_like(x, 1), torch.full_like(x, 0)).sum(dim=axis)
+    value = torch.where(mask, x, torch.full_like(x, 0)).sum(dim=axis)
+    mean = value / num
+    mean_broadcast = torch.repeat_interleave(mean.unsqueeze(axis), x.shape[axis], dim=axis)
+    quadratic_difference_from_mean = torch.square(torch.where(mask, mean_broadcast - x, torch.full_like(x, 0)))
+    return mean, torch.sqrt(torch.sum(quadratic_difference_from_mean, dim=axis) / (num - 1))
+
```
SamuelGabriel, line 197: Could this be split into two functions and then be re-used by torch_nanmean/nanstd?

tabpfn/utils.py
```diff
@@ -201,16 +211,16 @@ def normalize_data(data, normalize_positions=-1):
...
-    data_clean = data[:].clone()
+
     data_mean, data_std = torch_nanmean(data, axis=0), torch_nanstd(data, axis=0)
     cut_off = data_std * n_sigma
     lower, upper = data_mean - cut_off, data_mean + cut_off
 
-    data_clean[torch.logical_or(data_clean > upper, data_clean < lower)] = np.nan
-    data_mean, data_std = torch_nanmean(data_clean, axis=0), torch_nanstd(data_clean, axis=0)
+    mask = (data <= upper) & (data >= lower) & ~torch.isnan(data)
+    data_mean, data_std = torch_masked_mean_and_std(data, mask, axis=0)
```
SamuelGabriel: did you do a test that this does not change behavior? as this seems like a little bit of a dangerous change.

tabpfn/scripts/transformer_prediction_interface.py
```diff
@@ -167,8 +173,15 @@ def _validate_targets(self, y):
         return np.asarray(y, dtype=np.float64, order="C")
 
     def fit(self, X, y, overwrite_warning=False):
-        # Check that X and y have correct shape
-        X, y = check_X_y(X, y, force_all_finite=False)
+        """
+        Validates the training set and stores it.
+
+        If no_grad is set to false the function takes X as tensors and doesn't fully check the compatibility.
```
SamuelGabriel: This is hard to understand.
Maybe write something like this

If clf.no_grad:
    X,y should be of type np.array
else:
    X, y should both be of type torch.Tensors

tabpfn/scripts/transformer_prediction_interface.py
```diff
@@ -107,7 +107,8 @@ class TabPFNClassifier(BaseEstimator, ClassifierMixin):
 
     def __init__(self, device='cpu', base_path=pathlib.Path(__file__).parent.parent.resolve(), model_string='',
                  N_ensemble_configurations=3, combine_preprocessing=False, no_preprocess_mode=False,
-                 multiclass_decoder='permutation', feature_shift_decoder=True, only_inference=True, seed=0):
+                 multiclass_decoder='permutation', feature_shift_decoder=True, only_inference=True, seed=0,
+                 no_grad=True):
```
SamuelGabriel: we should add documentation here, if you feel good adding it, feel free to, if not, I am fine doing it after this PR.

SamuelGabriel (changes requested):
Looks pretty good in general. Only very minor, pretty easy to fix things I have to lament. 

It would be good to have some form to make sure the changes to the normalization code do not change the behavior of the TabPFN at all, as it would be a very big issue if this all of a sudden makes us worse or something.

tabpfn/scripts/transformer_prediction_interface.py
schnurrd: Yeah, sure. I will delete the if part with combine_preprocessing and remove all occurrences of it. This is then the only call to preprocess_input, and I will change it as you've described it.

tabpfn/utils.py
schnurrd: I have written a test (test_standardization) for the function remove_outliers (and indirectly the new nanmean and nanstd) that I can push with the changes and remove afterward. What I find interesting is that torch.sum and torch.nansum return slightly different values, which are sometimes off after around the sixth digit of the number. In the test, I currently have rewritten the nansum to the normal sum, so that the results are equal.

cyrilpic:
Hey, I just want to jump into the conversation and say that I have tested the code. I am able to use the gradient information and did not see any change in performance. Is there anything missing to merge it into the main branch? I'd be happy to help.

SamuelGabriel:
Merged and part of the pip package now :)

PR #33, opened 2023-03-20
Fix bug in is_classification

amueller:
Because of an issue in is_classification, auc_metric is not recognized as classification metric and invalid splits are generated on the validation set.

There is another implementation of ``is_classification`` that is correct, but this one is used.


amueller:
btw with that, the validation dataset still doesn't run, since the `flags` dataset can't be split apporopriately.

noahho:
Thank you very much Andreas!

PR #42, opened 2023-06-14
Fix typo in caching

amueller:
I think this is what you meant? Otherwise I don't see how the condition could be true.


noahho:
I didn't write this code but I do agree that this seems like the correct way to execute.

amueller:
where's the code from then?

noahho:
From @SamuelGabriel

PR #86, opened 2024-03-01
Update model_builder.py and transformer_prediction_interface.py for mixed precision training and inference

liuquangao:
Variable `train_mixed_precision` seems to have not been passed into the train function, which means that mixed precision training cannot be performed


liuquangao:
When using mixed precision inference, the output of the model is `torch. cuda HalfTensor`, this will result in some rows not summing up to 1 after calculating softmax, leading to Nan values when using `tabular_metrics.auc_metric(test_ys, prediction_)`
image

SamuelGabriel:
thanks a lot! this looks good, we had a renaming of this parameter at some point when uploading, maybe this happened due to that. Sorry for that!

PR #89, opened 2024-04-28
Updates to training notebook

carteakey:
Hi there!

While trying to reproduce the [training](https://github.com/carteakey/tabpfn-eval/tree/main) i thought of cleaning up some code, even though TabPFN 2.0 might be in the works (cant wait).

- Removed the duplicate model_config file, the one used in there in the model_configs.py inside scripts folder.
- Cleaned up the training notebook
     - removed unused imports
     - add the missing maximum_runtime variable in train function (removed total_training_in_s config as its not used apparently). Use the train_function for training instead of get_model.
     - updated some config to match the current classifier (pulled from `classifier.c`)

Feedback welcome!

Kartikey


LennartPurucker:
Heyho! We thank you for your contribution but will close this PR for now. 
Since we moved to TabPFN v2, the PR no longer represents the new code base. 

We would be happy to see similar community contributions for v2 in https://github.com/PriorLabs/tabpfn-extensions/tree/main/src/tabpfn_extensions

Issue #112, opened 2025-01-10
fix: joblib 1.4.0 `return_as` requirement


LennartPurucker:
The current code requires joblib 1.3.0 due to using `return_as` (https://github.com/joblib/joblib/blob/3aeca2025e62639f23e8aed013fb6505b3ce2eb8/CHANGES.rst#release-140----20240408). 

Can we use a workaround for <1.3.0? @eddiebergman 

Alternatively, require >=1.3.0 during the install. 

Rel. #101

LennartPurucker:
We shoud at least support >=1.2.0 following scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/main/pyproject.toml#L12C10-L12C17

LennartPurucker:
For a workaround, use `pip install -U joblib` for now.

eddiebergman:
It seems like for joblib < 1.4, there was a `return_generator: bool` option which returned a list if `False`, otherwise an `Iterator`. I guess the _easiest_ way to handle it is just a little try, except, but if you really wanted to be pedantic, I guess you could do a check on the joblib version.

See the title and changes of [this commit](https://github.com/joblib/joblib/pull/1463/commits/27629e01a0d1c7ab77cc60c9010a96fe0b666378)

LennartPurucker:
@noahho This will be fixed in the pip installable version. In the source code, it is already fixed. 

Leaving it open until we patched the pip version.

noahho:
Thanks a lot @LennartPurucker @eddiebergman 
We have the new pip version online

PR #114, opened 2025-01-10
fix: joblib <1.3 support

LennartPurucker:
Rel https://github.com/PriorLabs/TabPFN/issues/112


eddiebergman (approved):
Looks good to me, maybe document at `SUPPORTS_RETURN_AS` to reference scikit-learns minimum required which is `1.2`

PR #116, opened 2025-01-11
Specify numpy version requirement

clementzach:
When importing tabpfn, I got an error saying that some files were complied using numpy < 2.0 but my environment had numpy >2.0. This PR will ensure that the proper version of numpy is installed to reduce friction for future people importing the package


noahho:
Thanks a lot for looking to improve the package! Could you provide the error for me to see which package fails? I just installed numpy==2.0.0 and could run things on colab.

PR #117, opened 2025-01-11
Update README.md

viv-analytics:
Changed the import section in accordance with the colab example for the AutoTabPFNClassifier


noahho (approved):

noahho:
Thanks a lot for catching it!

PR #124, opened 2025-01-13
chore: update utils.py

eltociear:
ther -> there


noahho (commented):
Should be "the"

noahho:
Sorry, this change is too small in the codebase to merge! The comment continues and I think there is multiple smaller mistakes. Happy to look into larger corrections! :)

PR #126, opened 2025-01-13
fix `balance_probalities` argument

### LINKED ISSUE #122: `balance_probabilities` is maybe wrong?
The `balance_probabilities` argument from `TapPFNClassifier` is implemented like this:
```
        if self.balance_probabilities:
            class_prob_in_train = self.class_counts_ / self.class_counts_.sum()
            output = output * torch.Tensor(class_prob_in_train).to(self.device_)
            output = output / output.sum(dim=-1, keepdim=True)
```
I'm wondering whether `output = output * torch.Tensor(class_prob_in_train).to(self.device_)` shouldn't be `output = output / torch.Tensor(class_prob_in_train).to(self.device_)` to push models toward to minority class instead of the majority class. I may be confused about what this argument is for. @eddiebergman I think you implemented this, WTYT?

LeoGrin:
Fix #122

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -180,8 +180,10 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
             balance_probabilities:
                 Whether to balance the probabilities based on the class distribution
                 in the training data. This can help to improve predictive performance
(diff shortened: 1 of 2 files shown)


eddiebergman (approved):
LGTM! Thanks for finding and fixing this :)

PR #148, opened 2025-01-21
Accept dict as preprocessor config

LeoGrin:
Accept dict as input in addition to `PreprocessorConfig` objects (as is already done for `ModelInterfaceConfig`). Also move `ModelInterfaceConfig` to from `constants.py` to a new `config.py` to prevent circular imports.

This PR will make it easier to make AutoTabPFN compatible with the TabPFN API (see for instance https://github.com/PriorLabs/tabpfn-server/pull/135).


LennartPurucker (approved):
LGTM

PR #151, opened 2025-01-22
fix: make fingerprint feature deterministic

LennartPurucker:
Using Python's built-in hash is not consistent between runs due to salting. 

Thus, the fingerprint feature using the built-in hash is not deterministic, making the input data non-deterministic and making the predictions non-deterministic between runs. Hashlib's hash is consistent between runs and thus does not face this problem. 

This broke during the refactoring. The commit below is essentially reverting a commit that happened during the refactor. 

FYI: @eddiebergman

### DIFF
--- a/src/tabpfn/model/preprocessing.py
+++ b/src/tabpfn/model/preprocessing.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import warnings
 from abc import abstractmethod
 from collections import UserList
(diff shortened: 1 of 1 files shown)


LeoGrin (changes requested):
Thanks Lennart, LGTM! I would maybe just add a test that TabPFN is deterministic when setting the random state? I thought that scikit-learn checks would already check that but they were passing before. (I checked quickly and the only failing test `test_sklearn_compatible_estimator[TabPFNRegressor()-check_methods_sample_order_invariance]` still fails with this PR)

LennartPurucker:
We cannot test for this, as it requires separate Python runs (not just separate calls). 

And `test_sklearn_compatible_estimator[TabPFNRegressor()-check_methods_sample_order_invariance]` fails due to precision and not randomness.

LeoGrin:
Ag yeah ok the test does look hard to do! LGTM then!

LeoGrin (approved):

eddiebergman:
Ah damn, apologies. @LeoGrin here's the reference Lennart's refering to:
https://docs.python.org/3/reference/datamodel.html#object.__hash__

Scroll to the bottom of that function

PR #159, opened 2025-01-24
Add Python 3.12 PyPi tag

Innixma:
Seems that the Python 3.12 support isn't tagged in PyPi, so I added it.


noahho (approved):
Right, 3.12 should be added, thanks!

PR #165, opened 2025-01-31
enable ONNX export

AlexanderPfefferle:
modified the code to support ONNX export:
+ use custom implementation of torch.nanmean and torch.nansum everywhere
+ replaced usage of torch.Generator with code that isolates the rng in add_embeddings

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -112,8 +112,10 @@ jobs:
       - name: Install dependencies
         run: |
           uv pip install --system --no-deps .
-          uv pip install --system pytest
+          # onnx is required for onnx export tests
+          # we don't install all dev dependencies here for speed
           uv pip install --system -r requirements.txt
+          uv pip install --system pytest onnx
 
       - name: Initialize submodules
         run: git submodule update --init --recursive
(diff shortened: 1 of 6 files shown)


LeoGrin (approved):
LGTM, thanks a lot @AlexanderPfefferle 🥳🥳
The pre-commit error is not related to this PR, I've just fixed it on main.
For the regressor test, IIRC, the change is because `check_methods_subset_invariance` now also fail close to the 1e-7 threshold for lack of precision, correct?

AlexanderPfefferle:
Yes exactly, `check_methods_subset_invariance` is failing on the main branch right now with `Max absolute difference among violations: 3.199093e-06`.
I've added tests that check for exportability on CPU.

LeoGrin:
Thanks a lot @AlexanderPfefferle ! onnx should be added to the dependencies (but not the main ones) to make the tests green. (I can also do it if you add me to your fork).

LeoGrin:
I set the random state carefully to pass the regressor test for older versions, which is due to #175 and not to this pr.

LeoGrin:
Merging, thanks a lot @AlexanderPfefferle !

PR #172, opened 2025-02-05
Add an issue template + tabpfn.show_versions() function

LeoGrin:
Now add people to give a minimal reproducible example when submitting an issue to make our life easier (but this is not required) + some general guidelines inspired from sklearn. Also add the `tabpfn.show_versions()` function inspired from pytorch and sklearn to make it easy for users to share their setup. Currently I can't reproduce most of the bugs reported.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -7,7 +7,8 @@ body:
   attributes:
     value: >
       #### Before submitting a bug, please make sure the issue hasn't been already
-      addressed by searching through [the past issues](https://github.com/PriorLabs/TabPFN/issues).
+      addressed by searching through [the past issues](https://github.com/PriorLabs/TabPFN/issues)
+      and updating your tabpfn version by running `pip install tabpfn --upgrade`.
 - type: textarea
   attributes:
     label: Describe the bug
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,91 @@
+name: Bug Report
+description: Create a report to help us reproduce and correct the bug
+labels: [bug]
+
+body:
+- type: markdown
+  attributes:
+    value: >
+      #### Before submitting a bug, please make sure the issue hasn't been already
+      addressed by searching through [the past issues](https://github.com/PriorLabs/TabPFN/issues).
+- type: textarea
+  attributes:
+    label: Describe the bug
+    description: >
+      A clear and concise description of what the bug is.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Steps/Code to Reproduce
+    description: |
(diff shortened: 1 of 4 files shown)


src/tabpfn/debug_versions.py
```diff
@@ -0,0 +1,692 @@
...
+
+    for modname in deps:
+        try:
+            deps_info[modname] = get_version(modname)  # Use renamed function
+        except PackageNotFoundError:
+            deps_info[modname] = None
+    return deps_info
+
+
+def show_versions():
```
noahho: This doesn't seem to print anything or do anything with output?

src/tabpfn/debug_versions.py
```diff
@@ -0,0 +1,692 @@
...
+    _pretty_str(sys_info).replace(
+        "Versions of relevant libraries:\n"
+        "[pip3] No relevant packages\n"
+        "[conda] No relevant packages\n",
+        "",
+    )
+
+    # Print dependency versions in a clear section
+    deps_info = _get_deps_info()
+    for _pkg, _version in sorted(deps_info.items()):
```
noahho: this doesn't seem to print anything?

src/tabpfn/debug_versions.py
```diff
@@ -0,0 +1,692 @@
...
+
+def _get_deps_info():
+    """Overview of the installed version of main dependencies.
+
+    Returns:
+    -------
+    deps_info: dict
+        version information on relevant Python libraries
+    """
+    deps = [
```
noahho, line 632: could this be dynamically updated from the pyproject file or is that too complicated?

src/tabpfn/debug_versions.py
noahho: also, the function name isnt great even if from sklearn? Maybe could be "display_debug_info" ?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,91 @@
...
+    required: false
+- type: textarea
+  attributes:
+    label: Versions
+    render: shell
+    description: |
+      Please run the following and paste the output below.
+      ```python
+      import tabpfn; tabpfn.show_versions()
+      ```
```
noahho, line 86: +

noahho (changes requested):
This is super cool and will help us a lot!

src/tabpfn/debug_versions.py
LeoGrin: oops seems like it got removed while fixing the ruff complaints sorry, good catch!

src/tabpfn/debug_versions.py
LeoGrin: might be possible but doesn't seem straightforward to find the pytoml file in all cases for the user (while it's easy in the CI where we control everything)

.github/ISSUE_TEMPLATE/bug_report.yml
LeoGrin: We already have "Before submitting a bug, please make sure the issue hasn't been already
      addressed by searching through [the past issues](https://github.com/PriorLabs/TabPFN/issues)", do you want to replace it with that?

noahho: maybe change to by searching through [the past issues](https://github.com/PriorLabs/TabPFN/issues) and updating your tabpfn version by running `pip install tabpfn --upgrade`

src/tabpfn/debug_versions.py
noahho: okay lets not

Issue #111, opened 2025-01-10
Embeddings Functionality


noahho:
The embedding functionality was removed during the recent code refactor. Previously, the unsupervised model had a get_embedding() method that worked with the model's embedding method.

noahho:
Here are the most relevant connected pieces before the refactor:

1. Main embedding method:
```python
def get_embeddings(
    self, X: torch.Tensor, additional_y: dict = None
) -> torch.Tensor:
    """
    Get the embeddings for the input data `X`.

    Parameters:
        X (torch.Tensor): The input data tensor.
        additional_y (dict, optional): Additional labels to use during prediction.

    Returns:
        torch.Tensor: The computed embeddings.
    """
    return self.predict_full(
        X, additional_y=additional_y, get_additional_outputs=["test_embeddings"]
    )["test_embeddings"]
```

2. The predict_full path that processes embeddings:
```python
def predict_full(self, X, additional_y=None, get_additional_outputs=None) -> dict:
    """
    Predicts the target y given the input X.

    Parameters:
        X:
        additional_y: Additional inputs
        get_additional_outputs: Keys for additional outputs to return.

    Returns:
         (dict: The predictions, dict: Additional outputs)
    """
    X_full, y_full, additional_y, eval_position = self.predict_common_setup(
        X, additional_y_eval=additional_y
    )

    prediction, additional_outputs = self.transformer_predict(
        eval_xs=X_full,
        eval_ys=y_full, 
        eval_position=eval_position,
        additional_ys=additional_y,
        get_additional_outputs=get_additional_outputs,
        **get_params_from_config(self.c_processed_),
    )

    return {"proba": prediction, **additional_outputs}
```

3. The transformer_predict method that generates embeddings:
```python
def transformer_predict(
    self,
    eval_xs: torch.Tensor,  # shape (num_examples, [1], num_features)
    eval_ys: torch.Tensor,  # shape (num_examples, [1], [1])
    eval_position: int,
    bar_distribution: FullSupportBarDistribution | None = None,
    reweight_probs_based_on_train=False,
    additional_ys=None,
    cache_trainset_representations: bool = False,
    get_additional_outputs: list[str] = None,
) -> tuple[torch.Tensor, dict]:
    """
    Generates predictions from the transformer model.

    This method builds the ensemble configurations, applies preprocessing, runs the transformer
    model on the preprocessed data, and then aggregates the predictions from each configuration.
    """
    # ... initialization code ...

[...]

PR #178, opened 2025-02-10
Embeddings V2

KulikDM:
Hi,

This commit addresses  #111 
I have made two PRs where the functionality is slightly different, but they essentially both do the exact same thing.
Please pick which one you think is best for the package.
V2 adds the `get_embeddings` to each estimator.

Benefit is easier to find API, downside is more duplicate code.

Also added the appropriate tests.

Hope this helps!


src/tabpfn/regressor.py
```diff
@@ -707,6 +707,34 @@ def predict(  # noqa: C901, PLR0912
 
         return logit_to_output(output_type=output_type)
 
+    def get_embeddings(self, X: XType) -> np.ndarray:
```
iivalchev: just a minor observation, this method looks quite identical to the one in the classifier, perhaps duplication can be avoided an have a single method?

src/tabpfn/regressor.py
```diff
@@ -707,6 +707,34 @@ def predict(  # noqa: C901, PLR0912
...
+        """
+        check_is_fitted(self)
+
+        X = validate_X_predict(X, self)
+        X = _fix_dtypes(X, cat_indices=self.categorical_features_indices)
+        X = self.preprocessor_.transform(X)
+
+        embeddings: list[np.ndarray] = []
+
+        for output, config in self.executor_.iter_outputs(
```
iivalchev: Perhaps I am missing something but I don't see how you are getting the train/test embeddings for X. Shouldn't the `iter_outputs` method be passed a new param that would invoke the model with `only_return_standard_out=False`?

src/tabpfn/regressor.py
noahho: Yeah I agree, this does not get the embedding in the sense that it is the transformers intermediate representation but rather the class predictions or for regression bucket predictions of the model

noahho: The suggestion by iivalchev would be the way I would have implemented as well!

noahho:
Thanks so much for looking into this! Unfortunately the change does not actually retrieve the embeddings now, see the review comments.

KulikDM:
Oh shoot! 

Sorry, I saw in the `transformer.py` the code block
```
output_decoded["train_embeddings"] = train_encoder_out
output_decoded["test_embeddings"] = test_encoder_out
```

And traced it back to the regressor and classifier. 

Would adding the arg `only_return_standard_out=False` do the trick?
If so, I can make the adjustment to this version as well as updating the tests.

iivalchev:
> Oh shoot!
> 
> Sorry, I saw in the `transformer.py` the code block
> 
> ```
> output_decoded["train_embeddings"] = train_encoder_out
> output_decoded["test_embeddings"] = test_encoder_out
> ```
> 
> And traced it back to the regressor and classifier.
> 
> Would adding the arg `only_return_standard_out=False` do the trick? If so, I can make the adjustment to this version as well as updating the tests.

Looks that way, but you need to indicate this through the `tabpfn.inference.InferenceEngine.iter_outputs` with additional param to it, say `include_embbedings_in_output` wich defaults to False and only `get_embeddings` should set it to True. One more thing you should change how you handle what would then be returned from forward pass of the model as it would be a dict not just a single tensor.

KulikDM:
Hi @iivalchev 

Thanks for all the feedback!

The latest commit tries to incorporate all the suggestions mentioned above.
- Single function in `utils` that is called by the classifier and the regressor
- Pass `only_return_standard_out` as a new arg to the `inference.iter_outputs` (default set to `True` like before, embedding passes `False`)
- `iter_ouput` checks output type, and if type `dict`, it extracts the `test_embeddings` (not sure if this should be train was following #111)
- Tests have been updated 

Everything was working as expected on my side; hope this change is acceptable.

src/tabpfn/inference.py
```diff
@@ -494,4 +504,6 @@ def iter_outputs(
             # We'd rather just say unload from GPU, we already have it available on CPU.
             model = model.cpu()  # noqa: PLW2901
 
+            output = output['test_embeddings'] if isinstance(output, dict) else output
```
noahho: This change is a bit hacky for the application, ideally, we would retrieve the entire dict at this step and in the next step decide which entry to select. That would change the return type of this function.
Alternatively we could pass a parameter return_field = 'test_embeddings' which selects whcih field to select and which is linked to the only_return_standard_out. The first option would be cleaner though?

src/tabpfn/utils.py
```diff
@@ -40,6 +40,42 @@
 
 MAXINT_RANDOM_SEED = int(np.iinfo(np.int32).max)
 
+def _get_embeddings(model: TabPFNClassifier | TabPFNRegressor, X: XType) -> np.ndarray:
+    """
+    Get the embeddings for the input data `X`.
```
noahho: I am wondering if we should retrieve both train and test embeddings, make it configurable or document that we are only retrieving test and the other embeddings could be retrieved as well?

noahho:
Great! I think this should be working already, I just had some comments to make it more mantainable.

KulikDM:
Hi @noahho 

Should I make the changes to rather pass the dictionary to the `_get_embeddings` function and add an additional arg for the user to select which information they would like to retrieve?

KulikDM:
Hi @noahho 

Awesome, agreed! 

I have added the additional arg `data_source` to either select the test (default) or train embeddings in the `get_embeddings` function.

noahho:
Great! I just fixed the merge conflict and now CI should run.

KulikDM:
Hi @noahho 

Seems like it is failing for python 3.9 but not 3.12
The issue is an infinite showing up, not sure where or why.
The only thing that I can see that is different than the rest of the regressor tests is `n_estimators` is set to 3. Not sure if that has an impact on randomness. Just unlucky?

Your insights would be most appreciated :)

hzhz2020:
Dear Authors

Thanks for this great work! The performance in the paper is amazing and I have heard many good things about the work. I am eager to try it myself. I'm wondering do you have an expected timeline when this feature (obtaining embedding) would be available?

Thanks!

noahho:
PR looks great, thank you for the contribution @KulikDM !

PR #179, opened 2025-02-11
[Paused] Allow using ONNX compiled model in sklearn interface

### LINKED ISSUE #97: ONNX models
Do you support ONNX conversion?

LeoGrin:
Fix #97 

## Downside compared to PyTorch 

We don't do the memory estimation + reduction part, so it's easier to get memory error with large datasets (TODO)

(It seems that if if I export the onnx model on gpu I get less memory error afterward but not sure)

## Quick speed analysis:

For 1Kx10 on a T4 GPU node (seems to be quite hardware dependent):
- first fit: 5s for ONNX (we initialize the session), **0.4s for PyTorch**
- first predict: **~2s for ONNX**, 6s for Pytorch
- 2nd fit: **0.05s for ONNX**, 0.15s for PyTorch
- 2nd predict: 0.6-1.2s for ONNX (depending on where you export it, probably some more gains here?), **0.22s for PyTorch**

Seems like for smaller datasets ONNX becomes faster even for 2nd predict.
     
## Performance analysis on real datasets

Seemed good but need to rerun.

## TODOs

- [ ] Make memory reduction work on ONNX
- [ ] Allow to set the onnx session from outside to work with sklearn cloning
- [ ] Check more carefully that we haven't degraded performance (e.g on categoricals)
- [ ] Can we make ONNX model faster by exporting it with the right dummy input? How much speed vary by hardware?


reusyangyang:
Hi,does this model currently support training in a GPU environment and saving in ONNX format?

noahho:
@LeoGrin please close if stale

LeoGrin:
stale

PR #180, opened 2025-02-11
Use sklearn-compat to be compatible with all sklearn versions

LeoGrin:
Remove current annoying FutureWarnings.
Use (vendored) `sklearn-compat` to be compatible with different versions of sklearns. At the same time, create a new misc folder for this package and other misc files (like debug_versions)


LennartPurucker (approved):
LGTM!

PR #103, opened 2025-01-09
fix(windows): Ignore `os.sysconf` with windows

eddiebergman:
Fixes issue #100 by simply going with the default.

Should raise a specific issue to implement this properly

PR #184, opened 2025-02-13
Implement get_total_memory for windows

mert-kurttutan:
Addresses #103

### DIFF
--- a/src/tabpfn/utils.py
+++ b/src/tabpfn/utils.py
@@ -790,9 +790,6 @@ def _transform_borders_one(
     return logit_cancel_mask, descending_borders, borders_t
 
 
-
-
-
 # Terminology: Use memory to referent physical memory, swap for swap memory
 def get_total_memory_windows() -> float:
     """Get the total memory of the system for windows OS, using windows API.
@@ -800,10 +797,11 @@ def get_total_memory_windows() -> float:
     Returns:
         The total memory of the system in GB.
     """
+
     # ref: https://github.com/microsoft/windows-rs/blob/c9177f7a65c764c237a9aebbd3803de683bedaab/crates/tests/bindgen/src/fn_return_void_sys.rs#L12
     # ref: https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex
-    # this class is needed to load the memory status with GlobalMemoryStatusEx function using win32 API,
-    # for more details see microsoft docs link above
+    # this class is needed to load the memory status with GlobalMemoryStatusEx function
+    # using win32 API, for more details see microsoft docs link above
     class _MEMORYSTATUSEX(ctypes.Structure):
         _fields_: typing.ClassVar = [
             ("dwLength", ctypes.c_ulong),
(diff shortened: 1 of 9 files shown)


mert-kurttutan:
I will also add several test comparing against psutil results on windows workflows, with some several scenarios (e.g. multithreaded call to the get_total_memory function)

mert-kurttutan:
It might be a nitpick, to be super clear, but the terminology used in the comments is free_memory, but we were getting is total memory.
Probably, we should change variable name to total_memory to be consistent.

src/tabpfn/utils.py
```diff
@@ -786,3 +788,37 @@ def _transform_borders_one(
         )
 
     return logit_cancel_mask, descending_borders, borders_t
+
+
+# ref: https://github.com/microsoft/windows-rs/blob/c9177f7a65c764c237a9aebbd3803de683bedaab/crates/tests/bindgen/src/fn_return_void_sys.rs#L12
```
noahho: Could you please just add another comment explaining what this code is for? The name of the class could also reflect the meaning of it in the wider system. Might it even make sense to move this class inside the function to scope it clearly?

mert-kurttutan:
> It might be a nitpick, to be super clear, but the terminology used in the comments is free_memory, but we were getting is total memory. Probably, we should change variable name to total_memory to be consistent.

Just looked at the todo inside that function, this seems not resolved (whether to go for free or total, for example gpu code still uses free memory, cpus use total_memory). The naming of the variable is couple with this TODO, should be dealt with another PR.

mert-kurttutan:
Only thing to do is to add windows to os matrix in workflow. Is that OK with u?

noahho (approved):

noahho:
> The another thing to do is to add windows to os-matrix in workflow. Is that OK with u?
> Edit: We can run only the test for windows total memory function on these machine if we want

Sorry I didn't quite get this. "os-matrix in workflow" - where would this be?

mert-kurttutan:
> > The another thing to do is to add windows to os-matrix in workflow. Is that OK with u?
> > Edit: We can run only the test for windows total memory function on these machine if we want
> 
> Sorry I didn't quite get this. "os-matrix in workflow" - where would this be?
> 
> From my side looks good to merge

Sorry meant here: see matrix.os in https://github.com/PriorLabs/TabPFN/blob/main/.github/workflows/pull_request.yml

noahho:
> > > The another thing to do is to add windows to os-matrix in workflow. Is that OK with u?
> > > Edit: We can run only the test for windows total memory function on these machine if we want
> > 
> > 
> > Sorry I didn't quite get this. "os-matrix in workflow" - where would this be?
> > From my side looks good to merge
> 
> Sorry meant here: see matrix.os in https://github.com/PriorLabs/TabPFN/blob/main/.github/workflows/pull_request.yml

@LeoGrin

LeoGrin:
Thanks @mert-kurttutan for the PR! Yes adding windows to the CI tests would be great, thanks :) I think we can run all tests on windows.

mert-kurttutan:
Fails due to heredoc (which is valid only in shells for unix). 
An alternative approach is to use python file (located under scripts directory) than use it directly instead of heredoc tool. It is easier to read and more portable (unless there is some specific reason to keep using it).

mert-kurttutan:
Turns out minimal torch version has dependency on onnx with version that is buggy on windows. So, needed force downgrade of onnx version on that config.

LeoGrin:
Thanks a lot @mert-kurttutan for the fixes! I think I would rather skip the onnx compilation tests on Windows, as we don't really need them on all OS. Otherwise it looks great :)

LeoGrin:
Thanks a lot @mert-kurttutan for your contribution! 🙌🫡

Issue #138, opened 2025-01-15
TabPFN fails on text with NA


LeoGrin:
```
import pandas as pd
import numpy as np
from tabpfn import TabPFNClassifier
from sklearn.model_selection import train_test_split

# Create a sample dataset with text and NA values
data = {
    'text_feature': ['good product', 'bad service', pd.NA, 'excellent', 'average', pd.NA],
    'numeric_feature': [10, 5, 8, 15, 7, 12],
    'target': [1, 0, 1, 1, 0, 0]
}

# Create DataFrame
df = pd.DataFrame(data)

# Try to use the data directly without handling NA values
X = df[['text_feature', 'numeric_feature']]
y = df['target']

# Split the data
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.3, random_state=42)

# Initialize and try to train TabPFN
classifier = TabPFNClassifier(device='cpu')

# This will fail because TabPFN cannot handle pd.NA or string values directly
try:
    classifier.fit(X_train, y_train)
except Exception as e:
    print("TabPFN failed with error:\n", str(e))
```

gives
```
TabPFN failed with error:
 Encoders require their input to be uniformly strings or numbers. Got ['NAType', 'str']
 ```

kgovind0001:
@LeoGrin Thank you very much for sharing the file to reproduce the bug. There is mismatch in the data type which is required by function [_unique_python](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_encode.py) . 

[Pull Request](https://github.com/PriorLabs/TabPFN/pull/152) should fix this where we preprocess the X also before validating it.

LeoGrin:
Fixed by #242

PR #185, opened 2025-02-14
updated to accept NAN values in string columns: Fix for issue #138

kgovind0001:
@noahho Thank you very much for the feedback and guidance. This request should fix the [issue138](https://github.com/PriorLabs/TabPFN/issues/138) . I have made following corrections. 

1. Added the placement for NaN in string columns. Since the same processing needed to be applied in prediction, I had to add an attribute placeholder to the class. 
2. I have made similar changes to the regression also. 
3 tests have failed related to attribute. Please let me know if there is anything you would like to have another look. 

Thank you.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,10 @@ line-length = 88
 output-format = "full"
 src = ["src", "tests", "examples"]
 
+[tool.ruff.lint.mccabe]
+# Flag errors (`C901`) whenever the complexity level exceeds 5.
+max-complexity = 15
+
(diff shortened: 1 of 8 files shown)


noahho:
Hi @kgovind0001, thanks for your work on this!

The Ruff linting checks failed, which is blocking the PR. You can resolve most of the issues by running "ruff check . --fix"
This will automatically fix many formatting errors. Some issues might need manual adjustments, so you can check the output after running the command.

Let me know if you need any help!

PR #200, opened 2025-02-23
Fix dimension checking in MHA

mert-kurttutan:


mert-kurttutan:
This does not occur since all the MHA used have the property `d_v == d_k`

mert-kurttutan:
If we want to be super-rigorous about MHA module, it would be better to provide unit test that fuzz-tests with different inputs, weight dims, etc.

mert-kurttutan:
Alternatively, we could just get rid of code handling `d_v != d_k`, in which case the code becomes simpler but does not allow for `d_v != d_k`

LeoGrin:
Thanks @mert-kurttutan! I think the code assume `d_k == d_v` right now (for instance the other asserts above the one you fixed).

mert-kurttutan:
If you are referring to the `assert_tensor_shape` function, then they dont run in case `d_k != d_v` since those weights are set to none in which case `assert_tensor_shape` function returns early. So, I dont think they assume `d_k == d_v`, correct me If I am mistaken.
I also checked the error running some dummy code with different dimensions, these were only places where it gave error.

LeoGrin (approved):

LeoGrin:
Thanks @mert-kurttutan! I'm still a bit confused as to how `assert_tensor_shape(w_kv, [2, self._nhead_kv, self._d_k, self._input_size])` could work if `d_k != d_v`, but otherwise LGTM!

mert-kurttutan:
@LeoGrin `assert_tensor_shape` does not check the shapes (with early return statement) if the tensor is None. If `d_k != d_v`, `w_kv` is set to None, see [here](https://github.com/PriorLabs/TabPFN/blob/cfbbfe131d96817998a9f5754b183cab9f0e744a/src/tabpfn/model/multi_head_attention.py#L242), which leads to `assert_tensor_shape` not checking tensor shape of `w_kv`.

LeoGrin:
Ah yeah make sense, thanks!

PR #217, opened 2025-02-28
Add model consistency tests

noahho:
Added model consistency tests to verify TabPFN predictions remain consistent across code changes. Implementation verifies both ClassifierTabPFN and RegressorTabPFN outputs match reference values. Tests will fail if model predictions change, requiring explicit acknowledgment of changes.


LeoGrin (changes requested):
Would change to compare directly the predictions instead of the summary statistics, otherwise LGTM

LeoGrin (approved):
LGTM!
A few things:
- the new test README mentions "test_text_na_handling.py", which is not an actual test on GitHub
- in "should_run_consistency_test": "In CI environments, this will raise pytest.fail with instructions if the platform doesn't match the reference platform." I think this is not true anymore
- CI_PLATFORMS should be updated. We're now using python3.13 as the latest python version. (it would be nice to get directly the right version from pytoml but low priority maybe).

Except these this is ready to merge IMO 🥳

PR #240, opened 2025-03-12
Support python13

### LINKED ISSUE #234: pip defaults to older `tabpfn` version when you have python3.13.
Our current version doesn't support python3.13, but this was not specified for past tabpfn versions (though they probably don't support it either), so pip default to installing these older versions if you have python3.13 instead of saying that we don't support python3.13. This creates hard to understand issues like #203 where people don't realize they have an older `tabpfn` version

Solutions:
- support python3.13: we should do it either way, but we'll have the same issue when python3.14 comes out.
- yank all the older versions (incorrectly) supporting new python versions (Note: might need to change the instructions for old version `pip install tabpfn<2`)

LeoGrin:
Fix #234

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
   "einops>=0.2.0,<0.9",
   "huggingface-hub>=0.0.1,<1",
 ]
-requires-python = ">=3.9,<3.13"
+requires-python = ">=3.9"
 authors = [
   { name = "Noah Hollmann", email = "noah.hollmann@charite.de" },
   { name = "Samuel Müller", email = "muellesa@cs.uni-freiburg.de" },
@@ -46,6 +46,7 @@ classifiers = [
   'Programming Language :: Python :: 3.10',
   'Programming Language :: Python :: 3.11',
   'Programming Language :: Python :: 3.12',
+  'Programming Language :: Python :: 3.13',
 ]
 license = { file = "LICENSE" }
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -14,7 +14,7 @@ dependencies = [
   "einops>=0.2.0,<0.9",
   "huggingface-hub>=0.0.1,<1",
 ]
-requires-python = ">=3.9,<3.13"
+requires-python = ">=3.9"
```
noahho, line 17: Shouldn't this be <= 3.13 as we aren't testing newer versions?

noahho (approved):
Amazing! Just take a look at the requirements spec

pyproject.toml
LeoGrin: See some discussion in #234, I changed my mind and now think it makes sense not to upper bound the python version. WDYT?

noahho: ok, makes sense! Can we then test the package for version 3.14 as well or do you think that would fail?

noahho:
Feel free to directly merge. In case it makes sense, add 3.14 to the tested python versions.

noahho (approved):

pyproject.toml
mert-kurttutan: My main source of truth when it comes to python versions is: https://devguide.python.org/versions/
Also, the wheel support from torch (other libraries with C extension, which requires compilation if the wheel support is not present). 3.14 probably would require torch compilation, not really worth the effort.

noahho: okay and you would agree with Leo to keep the version number open anyways but not test it explicitly?

mert-kurttutan: Yeap, I would

PR #241, opened 2025-03-12
Add installation from the source

mert-kurttutan:
@LeoGrin 
In the same spirit as this: https://github.com/PriorLabs/tabpfn-extensions/pull/28

Should make it easier for end user to test very recent changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ Try our [Interactive Colab Tutorial](https://colab.research.google.com/drive/1SH
 # Simple installation
 pip install tabpfn
 
+# Installation from source
+pip install "tabpfn @ git+https://github.com/PriorLabs/TabPFN.git"
+
 # Local development installation
 git clone https://github.com/PriorLabs/TabPFN.git
 pip install -e "TabPFN[dev]"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +38,9 @@ Try our [Interactive Colab Tutorial](https://colab.research.google.com/drive/1SH
 # Simple installation
 pip install tabpfn
 
+# Installation from source
```
noahho: Can we make it clearer that we don't need to run all three commands, but it's either or. This is not your commit, but was an issue before. We could end the bash syntax, write "alternatively: ```bash" and go into next bash

mert-kurttutan: like this ?
to install from pip (official)
```bash
pip install tabpfn
```
or installation from source
```bash
pip install "tabpfn @ git+https://github.com/PriorLabs/TabPFN.git"
```
or local development installation
```bash

git clone https://github.com/PriorLabs/TabPFN.git
pip install -e "TabPFN[dev]"
```

noahho: yes exactly!

noahho: you agree this might be clearer (to a unix/pip beginner)?

mert-kurttutan: yeap

noahho:
Great, thank you!

PR #242, opened 2025-03-13
text-na-bug-fix

anuragg1209:
Hi, 

This Pull Request fixes bug #138 without requiring additional dependencies like `skrub` or `autogluon.features`.

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -52,6 +52,7 @@ from tabpfn.utils import (
     _fix_dtypes,
     _get_embeddings,
     _get_ordinal_encoder,
+    _process_text_na_dataframe,
     infer_categorical_features,
     infer_device_and_type,
     infer_random_state,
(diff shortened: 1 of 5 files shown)


noahho (approved):
Well done, this is great code and fixes the issue!

noahho:
Let's wait with the merge until we have the consistency check PR merged.

LeoGrin:
Thanks @anuragg1209, LGTM! As it would also be solved by using Skrub (or probably autogluon.features), some thoughts on whether we want to:
## Option 1: Use Skrub
### Pros:
Fix other issues like #163 and probably some other edge cases. I think it also enable polars dataframe support (to check).
### Cons:
- Force us to bump the minimum version for some package (but nothing too recent)
- One more dependency, and a dependency which is not so common, which can be an issue for people in companies where they only have access to a preinstalled list of packages (not sure, they do have to install tabpfn anyway so maybe it doesn't really make a difference?).

## Option 2: Vendor Skrub
Would fix the dependency issue, not sure how annoying this is.

## Option 3: Don't use Skrub, and do the few things we need ourselves.
I think we mostly need for now: NaNs in string fix (this PR), support datetime and bool dtypes, (polar support?).

WDYT? I think I'm leaning toward option 1 but not quite sure. In particular, I would like to understand better if the dependency will be an issue for some people.

PR #255, opened 2025-03-26
TabPFNRegressor preprocessing fails on bigger datasets fix

### LINKED ISSUE #169: TabPFNRegressor preprocessing fails on bigger datasets
See https://huggingface.co/Prior-Labs/TabPFN-v2-reg/discussions/2
It seems that QuantileTransformer fails on big datasets with message `The number of quantiles cannot be greater than the number of samples used`, which means TabPFN is unusable for these bigger datasets even with `ignore_pretraining_contraints=True`. Seems to only happen on regression? (not sure)

Krishnadubey1008:
This PR fixes #169 
chnages made- took n_quantiles=min(n_quantiles, 10_000) in `TabPFN/src/tabpfn/model/preprocessing.py`

### DIFF
--- a/src/tabpfn/model/preprocessing.py
+++ b/src/tabpfn/model/preprocessing.py
@@ -82,6 +82,57 @@ class KDITransformerWithNaN(KDITransformer):
         return X  # type: ignore
 
 
+class AdaptiveQuantileTransformer(QuantileTransformer):
+    """A QuantileTransformer that automatically adapts the 'n_quantiles' parameter
+    based on the number of samples provided during the 'fit' method.
+
+    This prevents errors that occur when the requested 'n_quantiles' is
+    greater than the number of available samples in the input data (X).
+    This situation can arises because we first initialize the transformer
+    based on total samples and then subsample.
+    """
+
+    def __init__(self, *, n_quantiles: int = 1000, **kwargs: Any) -> None:
+        # Store the user's desired n_quantiles to use as an upper bound
+        self._user_n_quantiles = n_quantiles
+        # Initialize parent with this, but it will be adapted in fit
+        super().__init__(n_quantiles=n_quantiles, **kwargs)
+
+    def fit(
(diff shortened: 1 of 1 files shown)


noahho:
Thanks so much for this change! Would yu be able to add a test for this change, i.e. one that tests if the preprocessing runs on datasets of > 10,000 samples. We can't run the inference step unfortunately as ofc, it would be way too slow. Only way to test the inference on lareg datasets was if we provided a tiny tabpfn checkpoint, a very small model that is randomly initialized but that would be a project in itself.

noahho:
Great, this looks really good. There seems to be a tiny ruff issue at this point. Do you know how to resolve?
"ruff check . --fix" with ruff version 0.8.6

noahho:
Ohh also something that copilot just caught:
The 'quantile_uni_coarse' transformer now caps n_quantiles to 10,000, yet the 'quantile_uni' transformer remains uncapped.

Krishnadubey1008:
i will fix it now

Krishnadubey1008:
@noahho I had  ruff check . --fix but still ruff linting test is failing

noahho:
The two open ones don't seem to be automatically fixable:
src/tabpfn/regressor.py:723:89: E501 Line too long (89 > 88)
tests/test_preprocessing.py:12:9: NPY002 Replace legacy `np.random.rand` call with `np.random.Generator`

An LLM will know how to fix number 2 and by deleting a character in src/tabpfn/regressor.py:723:89 you fix no1

Krishnadubey1008:
@noahho Please review

noahho:
Thanks a lot for continuing to work on this. It seems there were a few changes made for the linting that weren't right (such as adding ""). I'll look into the PR and fix those things, if you'd like me to.

Krishnadubey1008:
Yes ,sure

mohammadp1001:
Any update on this PR? #169 already solved?

Issue #269, opened 2025-04-07
Add Apple Silicon support (mps)


LeoGrin:
### Describe the workflow you want to enable

"mps" device should be accepted as a valid device and used. "auto" device should check if it's available.

### Describe your proposed solution

To determine. It might directly work.

### Impact

None

oscarkey:
hey @klemens-floege , was this fixed by https://github.com/PriorLabs/TabPFN/pull/405?

klemens-floege:
@oscarkey Yes, since commit: 660c98fe346b8bbe644a3fea48f9af61caae73b4 we now support MPS. :)

oscarkey:
amazing!

PR #270, opened 2025-04-07
Step towards adding MPS support.

anuragg1209:
Hi @LeoGrin & @noahho,

This PR is a step towards adding support for Apple Silicon (mps) #269 . I believe the mps support will not work directly. The two main issues with mps are that: 

1. Torch does not support autocast for mps yet. 
2. The MPS framework doesn't support float64 tensors. We have to use float32 for it to function properly. Due to this, we need to make changes in other areas. I would like to hear your thoughts on this. Please correct me if I am mistaken.

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -185,9 +185,13 @@ class InferenceEngineOnDemand(InferenceEngine):
                 y_train = y_train.type(self.force_inference_dtype)  # type: ignore  # noqa: PLW2901
 
             style = None
-
+            autocast_ctx = (
+                contextlib.nullcontext()
+                if device.type == "mps"
+                else torch.autocast(device.type, enabled=autocast)
+            )
             with (
-                torch.autocast(device.type, enabled=autocast),
+                autocast_ctx,
                 torch.inference_mode(),
             ):
                 output = self.model(
@@ -324,9 +328,13 @@ class InferenceEngineCachePreprocessing(InferenceEngine):
             )
 
             style = None
-
+            autocast_ctx = (
+                contextlib.nullcontext()
+                if device.type == "mps"
+                else torch.autocast(device.type, enabled=autocast)
+            )
             with (
-                torch.autocast(device.type, enabled=autocast),
+                autocast_ctx,
                 torch.inference_mode(),
             ):
                 output = self.model(
@@ -424,9 +432,13 @@ class InferenceEngineCacheKV(InferenceEngine):
             # We do not reset the peak memory for cache_kv mode
             # because the entire data has to be passed through the model
             # at once to generate the KV cache
-
+            autocast_ctx = (
+                contextlib.nullcontext()
+                if device.type == "mps"
+                else torch.autocast(device.type, enabled=autocast)
+            )
             with (
-                torch.autocast(device.type, enabled=autocast),
+                autocast_ctx,
                 torch.inference_mode(),
             ):
                 ens_model.forward(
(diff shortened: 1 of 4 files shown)


LeoGrin:
Thanks @anuragg1209!

> The MPS framework doesn't support float64 tensors. We have to use float32 for it to function properly. Due to this, we need to make changes in other areas. I would like to hear your thoughts on this. Please correct me if I am mistaken.

Aren't we defaulting to float32 for the pytorch part (using 64 for the sklearn part only?)? Not sure about that, but I think we can (should?) do that if we need to.

I see that the CI tests on macbook are failing with a memory error. It makes sense that we're using more memory if we're not using autocast, but it seems worrying as the test examples are very lightweight. Do you have an idea of what is the issue? Perhaps one of:
1) There is an issue with the memory estimation?
2) Our model/code takes too much memory for a standard mps setting
3) The CI hardware is unrealistically weak

noahho:
@anuragg1209

noahho:
@anuragg1209 to be able to merge this could you look into it again?

anuragg1209:
> Thanks @anuragg1209!
> 
> > The MPS framework doesn't support float64 tensors. We have to use float32 for it to function properly. Due to this, we need to make changes in other areas. I would like to hear your thoughts on this. Please correct me if I am mistaken.
> 
> Aren't we defaulting to float32 for the pytorch part (using 64 for the sklearn part only?)? Not sure about that, but I think we can (should?) do that if we need to.
> 
> I see that the CI tests on macbook are failing with a memory error. It makes sense that we're using more memory if we're not using autocast, but it seems worrying as the test examples are very lightweight. Do you have an idea of what is the issue? Perhaps one of:
> 
> 1. There is an issue with the memory estimation?
> 2. Our model/code takes too much memory for a standard mps setting
> 3. The CI hardware is unrealistically weak

Hi @LeoGrin, yes we are defaulting to float32 for pytorch part and float64 for sklearn and numpy part. 

The CI hardware for the MPS backend appears to be quite limited. This issue likely stems from constraints in the virtualized MPS environment provided by GitHub hosted macOS runners. While workarounds like switching runner versions or disabling MPS usage in CI exist, they aren't foolproof solutions.

Regarding memory estimation, we utilize PyTorch's default functionality, which relies on the Metal API to assess available memory. However, in virtualized or constrained environments like GitHub Actions, these estimations may not accurately reflect the actual available resources, leading to unexpected out-of-memory errors despite seemingly sufficient memory.

Additionally, I've noticed minor inconsistencies in performance metrics when using MPS compared to CPU on my local machine. (I am on Macbook Pro 2019 which has AMD GPU as MPS backend.)

anuragg1209:
> @anuragg1209 thanks a lot, this is great progress! So on your device this runs already? We might have to change the MPS device detection and make sure the MPS device allows for some memory? It seems on CI the device exists, but is effectively turned off making the tests fail?

Yes, it runs on my device! But the predictions are not consistent. Maybe the inconsistency could be due to the pretty old version of my MPI device. 

Yes, I think we might have to change the MPS device detection for CI. It seems to be turned off due to which the test fails showing the memory error.

liam-sbhoo:
@anuragg1209 any updates on this? :)
Would be great to have MPS supported.

dolchan:
Not sure if this is applicable to other Mac devices, but I just made a quick change in get_max_free_memory() in model.py so that it returns estimated GPU memory on MPS as follows, as suggested in the comments: 

            free_memory = torch.mps.recommended_max_memory() - torch.mps.current_allocated_memory()

when I tested the Basic Usage -> Classification, it seemed to work.  I am using MacBook Pro M2 Max.

PR #275, opened 2025-04-12
Rework allow_cpu_override to be usable without environment variables

LennartPurucker:
Currently, one can only use TabPFN on a CPU with more than 1000 samples by setting an environment variable.
That makes using TabPFN in pipelines very annoying. 

This PR also enables users to ignore this limit by setting `ignore_pretraining_limits=True`.

### DIFF
--- a/src/tabpfn/base.py
+++ b/src/tabpfn/base.py
@@ -245,15 +245,22 @@ def create_inference_engine(  # noqa: PLR0913
 
 
 def check_cpu_warning(
-    device: str | torch.device, X: np.ndarray | torch.Tensor | pd.DataFrame
+    device: str | torch.device,
+    X: np.ndarray | torch.Tensor | pd.DataFrame,
+    allow_cpu_override: bool = False,
 ) -> None:
     """Check if using CPU with large datasets and warn or error appropriately.
 
     Args:
         device: The torch device being used
         X: The input data (NumPy array, Pandas DataFrame, or Torch Tensor)
+        allow_cpu_override: If True, allow CPU usage with large datasets.
     """
-    allow_cpu_override = os.getenv("TABPFN_ALLOW_CPU_LARGE_DATASET", "0") == "1"
+    allow_cpu_override = allow_cpu_override or (os.getenv("TABPFN_ALLOW_CPU_LARGE_DATASET", "0") == "1")
+
+    if allow_cpu_override:
+        return
+
     device_mapped = infer_device_and_type(device)
 
     # Determine number of samples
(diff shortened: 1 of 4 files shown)


noahho:
Hi @LennartPurucker,

Thanks for providing this PR! The functionality definitely makes sense, and making the CPU override configurable via a parameter instead of relying solely on an environment variable is a welcome improvement, especially for use in pipelines.

Could you clarify if your pipeline use case involves regularly running TabPFN on CPU for datasets significantly larger than the 1000 sample limit? That would be quite slow but is still desired? This might mean we want to also change the limit parameters, if people still find it useful below. The limit should be set at a point where people wouldn't really want to use TabPFN.

Regarding the parameter name ignore_pretraining_limits=True: While I understand the desire to set this directly on the model, the naming feels convoluted, implying a different scope than just bypassing the specific CPU performance warning for larger datasets.

However, considering the practicalities, renaming the parameter now could be disruptive for any potential early adopters, and adding another separate parameter just for the CPU override might unnecessarily complicate the API.

Therefore, sticking with the current name ignore_pretraining_limits seems reasonable. The key action then would be to ensure the main documentation for this parameter is very explicit and updated (classifier.py and regressor.py). It should clearly state that setting it True also allows users to proceed with CPU execution even when the sample size exceeds 1000.

Would you agree with all of this? @LennartPurucker @LeoGrin

LennartPurucker:
Right now, it is used in a benchmarked pipeline where the system does not have access to a GPU but enough CPUs even to run something slow. 

I think the use case is not necessarily realistic, and yes, it is quite slow. @Innixma might have a different perspective. 
Consumers or practitioners likely have a GPU nowadays (or at least a Mac). 

>  adding another separate parameter just for the CPU override might unnecessarily complicate the API.

I fully agree. I also started with a separate parameter, but that was too much. Moreover, `ignore_pretraining_limits` seemed like a catch-all for our recommended default limits. One could rename it to `ignore_limits` in the future, or argue in the documentation that the CPU limit is also an artifact of the pretraining due to the architecture used during pretraining. 

Or we can move it into a parameter under `inference_config`, which we can fill out with more details.

LeoGrin:
I agree that using the same parameter `ignore_pretraining_limits` is a good idea. `ignore_limits` or `ignore_size_limits` would be a better name, but yeah not super sure if it's worth changing, especially if this is a rare usecase.

noahho:
@LennartPurucker would you look into changing docstring in classifier and regressor to make this PR ready to merge?

LennartPurucker:
I already added the following as part of the PR:

```
                - If `True`, the model will not raise an error if the input data is
                  outside the pre-training range. Also allows to use the model with
                  more than 1000 samples on CPU.
```

@noahho do you want it to be more explicit or in a specific framing?

noahho:
@LennartPurucker - sorry I overlooked this, this looks great - thank you!

PR #278, opened 2025-04-16
Fix missing comma in pyproject.toml

BrownianNotion:
`pip install -e ".[dev]"` currently yields the following error due to a missing comma on line 20. Error was introduced in commit 233e20b7fa2d0515eddd978bf7a75fba6523f497
```
tomllib.TOMLDecodeError: Unclosed array (at line 20, column 3)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
 ]
 requires-python = ">=3.9"
 authors = [
-  { name = "Noah Hollmann" }
+  { name = "Noah Hollmann" },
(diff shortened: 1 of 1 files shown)


LennartPurucker:
@BrownianNotion Thank you for noticing and fixing this!

noahho:
Of course @BrownianNotion @LennartPurucker thank you so much. Big learning for me here 🫣

noahho:
I just added branch protection without admin bypass so we won't break things again. Thanks again for the check and quick fix!

PR #282, opened 2025-04-28
Use hash based versioning for actions

mert-kurttutan:
Using commit hashes is more secure as it locks to a specific version, preventing new (potentially malicious) version from running. This would prevent for instance: https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           src: "./src ./tests"
           version: 0.8.6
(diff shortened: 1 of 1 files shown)


mert-kurttutan:
Also, use ruff action from astral as the old one is archived.

mert-kurttutan:
@LeoGrin pinging. I think this is especially important after changed-files incident I linked above.

noahho:
Dear @mert-kurttutan, thanks so much for looking into this! Apologies for slow time getting back to you - LGTM.

mert-kurttutan:
No worries. This seems to be the starndard move after the incident I linked. Happy to make it secure.

Issue #287, opened 2025-05-06
Potential issue with bar distribution calculation


marcosfelt:
https://github.com/PriorLabs/TabPFN/blob/e8744e461dbd092d82389f6351f5f9cd5789d9d4/src/tabpfn/utils.py#L568-L568

Should this code be:
```python
share_of_bucket_left = (ys - borders[y_buckets]) / bucket_widths[y_buckets]
```

iivalchev:
I believe you are right! If you look at: https://github.com/PriorLabs/TabPFN/blob/e8744e461dbd092d82389f6351f5f9cd5789d9d4/src/tabpfn/model/bar_distribution.py#L85 
It has the expected parentheses By the looks of it it affects the `TabPFNRegressor` only. There is some duplication between `utils.py` and `bar_distribution.py` perhaps using `cdf` method from `BarDistribution` will be better maintenance-wise.

noahho:
Amazing you spotted this and @iivalchev was able to verify! I agree this looks like a mistake.

marcosfelt:
Thanks for the fix!

PR #289, opened 2025-05-08
fix share of bucket calc in bar dist cdf

iivalchev:
fix for #287

@noahho fyi

@marcosfelt spotted this


noahho:
Thanks a lot for quickly verifying and making this PR. Curious to see how it improves results.

Issue #268, opened 2025-04-07
Add TPU support


LeoGrin:
### Describe the workflow you want to enable

On platforms like Kaggle, the available GPU hardware (e.g., P100, T4) is often dated and heavily utilized. In contrast, powerful TPUs are frequently underused. Enabling TabPFN to run on TPUs would offer a compelling and high-performance alternative for users, particularly within the Kaggle community.

### Describe your proposed solution

Allow to use TPUs as a device, instead of just CPUs and GPUs. TPUs should also be detected if device is "auto".
Ideally, installation of `torch_xla` should be optional I think.

Unfortunately while this enables TPU usage, the actual utilization is very low. More optimization is needed.

### Impact

None

noahho:
PR by @MagnusBuehler looked into this https://github.com/PriorLabs/TabPFN/pull/424. However speed improvements seem marginal compared to CPU, more work would be needed to make our architecture efficient on TPU.

PR #291, opened 2025-05-13
feat: Implement TPU support

demoncoder-crypto:
- Adds TPU device detection and XLA optional dependency.
- Modifies attention mechanisms for CUDA-specific optimizations.
- Files applied manually.

Addresses #268

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "tabpfn"
-version = "2.0.9"
+version = "2.0.9" # Or the version from your clean clone
 dependencies = [
   "torch>=2.1,<3",
   "scikit-learn>=1.2.0,<1.7",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -319,26 +322,26 @@ typeCheckingMode = "strict"
 
 strictListInference = true
 strictSetInference = true
-strictDictionaryInference = false
+strictDictionaryInference = false # Changed to false as per original
 reportImportCycles = false
 reportMissingSuperCall = true
-reportMissingTypeArgument = false
+reportMissingTypeArgument = false # Changed to false as per original
```
noahho, line 328: strange comment, please remove lots of these dev comments

noahho:
Hey @demoncoder-crypto! Thanks for opening a PR 🤗 
This is really impactful, but there is a lot of changes that are unrelated to the PR. This includes left-over comments from dev (maybe from LLM agent?, e.g. in the pyproject.toml which shouldn't be changed right now), and lots of renamings. If tehre is work on code quality that should be separate from the change for TPU compatability.
Would be great to have this cleaned up and scoped down a bit!

PR #312, opened 2025-05-22
Add S3 fallback test and clean download docs

noahho:
## Summary
- delete unused `download_all_tabpfn_models.sh` script
- remove reference to that script in the README
- fix `get_fallback_urls` implementation
- add `tests/test_download_fallbacks.py` to check S3 fallback logic

## Testing
- `pre-commit run --files tests/test_download_fallbacks.py src/tabpfn/model/loading.py README.md`
- `PYTHONPATH=$PWD/src pytest -q tests/test_download_fallbacks.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,9 +228,9 @@ This script will download the main classifier and regressor models, as well as a
 
 **Manual Download**
 
-1. Download the model files manually from HuggingFace:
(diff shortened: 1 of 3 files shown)


MagnusBuehler:
@noahho  The logic looks good to me. I have two small suggestions:

- `urllib.request.urlopen(...)` without a `timeout` parameter can hang indefinitely if the server doesn’t respond, since the default timeout is set to `None`.
- It looks like the code was added to `src/tabpfn/model/loading.py`, which is deprecated. It should possibly be placed in `src/tabpfn/model_loading.py` instead.

PR #325, opened 2025-06-02
Add AdaptiveQuantileTransformer everywhere

noahho:


iivalchev:
@noahho perhaps the `AdaptiveQuantileTransformer` should handle the subsample being smaller than the number of quantiles issue.

noahho:
@iivalchev exactly that was the idea! This one would automatically adapt the quantiles if too large, but it was added in all places previoulsy. It seems newer sklearn version would handle on their own.

Issue #284, opened 2025-05-05
TabPFNClassifier predictions fail if inference precision is set to `torch.float16`


b-zhou:
### Describe the bug

Setting `inference_precision` to `torch.float16` results in `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::Half`.

### Steps/Code to Reproduce

```py
from sklearn.datasets import make_classification
from tabpfn import TabPFNClassifier
import torch

# Generate synthetic classification data
X, y = make_classification(n_samples=100, n_features=10, random_state=42)

# Initialize and train TabPFN
classifier = TabPFNClassifier(n_estimators=4, inference_precision=torch.float16)
classifier.fit(X, y)

# Make predictions
y_pred = classifier.predict(X)
```

### Expected Results

No error is thrown. Or if this torch dtype is indeed not supported, a more informative error message would be much appreciated.

### Actual Results

```
Traceback (most recent call last):
  File "/home/bszh/tabpfn/dtype_error.py", line 13, in <module>
    y_pred = classifier.predict(X)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/bszh/.virtualenvs/r-tabpfn/lib/python3.12/site-packages/tabpfn/classifier.py", line 531, in predict
    proba = self.predict_proba(X)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/bszh/.virtualenvs/r-tabpfn/lib/python3.12/site-packages/tabpfn/classifier.py", line 554, in predict_proba
    for output, config in self.executor_.iter_outputs(
  File "/home/bszh/.virtualenvs/r-tabpfn/lib/python3.12/site-packages/tabpfn/inference.py", line 332, in iter_outputs
    output = self.model(
             ^^^^^^^^^^^
  File "/home/bszh/.virtualenvs/r-tabpfn/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bszh/.virtualenvs/r-tabpfn/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #335, opened 2025-06-24
Fix Inference Precision Bug

noahho:
## Motivation and Context

This pull request addresses a critical bug that caused crashes when using `TabPFNClassifier` or `TabPFNRegressor` with `inference_precision=torch.float16`. The issue manifested in two stages:

1. A `RuntimeError` due to a data type mismatch between the input tensor and the model's internal `float16` weights.

2. After fixing the initial crash, a `ValueError` would occur due to `NaN` values being generated. This was caused by numeric instability in the normalization steps, specifically when a feature was constant (`std=0`) or had only a single data point. Division by zero in these edge cases produced `inf` values, which were then converted to `NaN`s in subsequent encoding steps.

This PR implements a robust, multi-layered fix within `src/tabpfn/model/encoders.py` to ensure that low-precision inference is stable and reliable. The core changes are:

* Handling division-by-zero in `torch_nanstd` when only one sample is present.

* Handling division-by-zero in `normalize_data` when a feature is constant.

* Explicitly casting the input tensor's `dtype` in `LinearInputEncoderStep` to match the layer's weights.

These changes resolve the bug for both the classifier and the regressor without altering the public-facing API.

## Public API Changes

* \[x] No Public API changes

* \[ ] Yes, Public API changes (Details below)

## How Has This Been Tested?

The fix has been thoroughly validated by adding `torch.float16` as a parameterized option to the existing test suites for both `TabPFNClassifier` (`tests/test_classifier_interface.py`) and `TabPFNRegressor` (`tests/test_regressor_interface.py`).

The tests, which previously failed with `RuntimeError` and `ValueError` for `torch.float16`, now pass successfully across all configurations, confirming that the numeric instability and `dtype` mismatch issues have been resolved.

## Checklist

* \[x] The changes have been tested locally.

* \[ ] Documentation has been updated (if the public API or usage changes).

* \[x] A entry has been added to `CHANGELOG.md` (if relevant for users).

* \[x] The code follows the project's style guidelines.

* \[x] I have considered the impact of these changes on the public API.

Fixed https://github.com/PriorLabs/TabPFN/issues/284

### DIFF
--- a/src/tabpfn/base.py
+++ b/src/tabpfn/base.py
@@ -192,6 +192,14 @@ def determine_precision(
         byte_size:
             The byte size per element for the chosen precision.
     """
+    # PyTorch CPU kernels do not support float16. We must error out.
+    if device_.type == "cpu" and inference_precision == torch.float16:
+        raise RuntimeError(
+            "Half-precision (torch.float16) inference is not supported on CPU. "
+            "Please use a GPU, or set inference_precision to 'auto', "
+            "torch.bfloat16, torch.float32, or torch.float64."
+        )
+
     if inference_precision in ["autocast", "auto"]:
         use_autocast_ = infer_fp16_inference_mode(
             device=device_,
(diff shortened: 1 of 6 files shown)


LeoGrin:
bugbot run

noahho:
@LeoGrin normalization could also have effect on model training, possibly making it more stable now? Exploding gradients / nans could be a cause of unstable normalization in our encoders

LeoGrin:
bugbot run

LeoGrin:
bugbot run

LeoGrin (approved):
LGTM

PR #352, opened 2025-06-26
Add executor save/load

noahho:
## Summary
- implement `save_fitted_tabpfn_model` and `load_fitted_tabpfn_model`
- use helper utilities in `TabPFNClassifier`/`TabPFNRegressor`
- document saving fitted models in the README and example script
- export helpers from the package and test round‑trip persistence

## Testing
- `pre-commit run --files src/tabpfn/model/loading.py src/tabpfn/classifier.py src/tabpfn/regressor.py src/tabpfn/__init__.py README.md examples/save_and_load_model.py tests/test_save_load_fitted_model.py`
- `pytest -q tests/test_save_load_fitted_model.py`

------
https://chatgpt.com/codex/tasks/task_b_685c5da037fc83338f5ae3d822d595e0


README.md
```diff
@@ -240,11 +240,37 @@ This script will download the main classifier and regressor models, as well as a
...
+# Train the regressor on GPU
+reg = TabPFNRegressor(device="cuda")
+reg.fit(X_train, y_train)
+save_fitted_tabpfn_model(reg, "my_reg.tabpfn_fit")
+
+# Later or on a CPU-only machine
+reg_cpu = load_fitted_tabpfn_model("my_reg.tabpfn_fit", device="cpu")
+```
+
+To store just the foundation model weights (without a fitted estimator) use
```
LeoGrin, line 269: why do we need this?

src/tabpfn/classifier.py
```diff
@@ -828,3 +826,41 @@ def get_embeddings(
             np.ndarray: The computed embeddings for each fitted estimator.
         """
         return _get_embeddings(self, X, data_source)
+
+    def save_fit_state(self, path: Path | str) -> None:
```
LeoGrin, line 830: why use this function instead of directly save_fitted_tabpfn_model?

src/tabpfn/model/loading.py
```diff
@@ -718,3 +738,126 @@ def save_tabpfn_model(model: nn.Module, save_path: Path | str) -> None:
...
+    """Persist a fitted TabPFN estimator to ``path``.
+
+    This stores the initialization parameters and the fitted state, but crucially
+    omits the large foundation model weights for efficiency.
+    """
+    if not hasattr(estimator, "executor_"):
+        raise RuntimeError("Estimator must be fitted before saving.")
+
+    path = Path(path)
+    if path.suffix != ".tabpfn_fit":
```
LeoGrin, line 753: this seems a bit strict no?

LeoGrin: ah I see in the rest of the code that it's actually needed. I still think it's a bit harsh and we could handle any filename but maybe that's fine

src/tabpfn/regressor.py
```diff
@@ -1015,6 +1018,44 @@ def get_embeddings(
         """Gets the embeddings for the input data `X`."""
         return _get_embeddings(self, X, data_source)
 
+    def save_fit_state(self, path: Path | str) -> None:
```
LeoGrin, line 1021: same comment

src/tabpfn/inference.py
```diff
@@ -100,6 +102,28 @@ def use_torch_inference_mode(self, *, use_inference: bool):
             "This inference engine does not support torch.inference_mode changes."
         )
 
+    def save_state(self, path: str | Path) -> None:
```
LeoGrin: name could be more informative

src/tabpfn/model/loading.py
```diff
@@ -718,3 +738,126 @@ def save_tabpfn_model(model: nn.Module, save_path: Path | str) -> None:
...
+        params = estimator.get_params(deep=False)
+        params = {
+            k: (str(v) if isinstance(v, torch.dtype) else v) for k, v in params.items()
+        }
+        params["__class_name__"] = estimator.__class__.__name__
+        with (tmp / "init_params.json").open("w") as f:
+            json.dump(params, f)
+
+        # 2. Automatically save all scikit-learn fitted attributes
+        fitted_attrs = {
```
LeoGrin, line 772: would this contain estimator.model_?

src/tabpfn/model/loading.py
```diff
@@ -718,3 +738,126 @@ def save_tabpfn_model(model: nn.Module, save_path: Path | str) -> None:
 
     # Save the checkpoint
     torch.save(checkpoint, save_path)
+
+
+def save_fitted_tabpfn_model(estimator: BaseEstimator, path: Path | str) -> None:
```
LeoGrin, line 743: not sure if it's better but what would you think about instead 1) setting .model_  to None 2) pickling 3) when loading, putting the model back? (basically the same thing you're doing but relying on pickle for everything instead of saving the params yourself).

LeoGrin (changes requested):

README.md
LeoGrin: (this=save_tabpfn_model)

LeoGrin:
bugbot run

src/tabpfn/model/loading.py
noahho: I tried to avoid pickling as it's problematic with version upgrades and inter system compatability. When one workflow / machine saves a model it won't be restoreable with another version. Also pickle is pretty unsafe. With this workflow we can control and potentially enable cross version compatability.

src/tabpfn/model/loading.py
noahho: Yeah I think they will always have this filename unless it's the worng file

src/tabpfn/regressor.py
noahho: You're right

src/tabpfn/model/loading.py
noahho:
```
# Attributes that are handled separately or should not be saved.
blacklist = {"model_", "executor_", "config_", "device_"}

# ... (omitted code) ...

# 2. Automatically save all scikit-learn fitted attributes
fitted_attrs = {
    key: value
    for key, value in vars(estimator).items()
    if key.endswith("_") and key not in blacklist
}
joblib.dump(fitted_attrs, tmp / "fitted_attrs.joblib")
```
--> No

README.md
noahho: I'm unsure if you use this in the fomo fitting code? Sometimes we want to save the model I guess, e.g. fientuning model too?

src/tabpfn/inference.py
noahho: It's a pretty common name for what the function does, tried standard naming

LeoGrin: I was thinking save_state_expect_model_weights or something but that's ok

src/tabpfn/model/loading.py
LeoGrin: nitpick: yeah ok, maybe then it makes sense to add the suffix ourselves, instead of forcing the user to add it?

README.md
LeoGrin: hum yeah not 100% sure we want to remove the function, but probably make sense to remove this from the readme right?

src/tabpfn/model/loading.py
LeoGrin: ah yeah missed that!

src/tabpfn/model/loading.py
LeoGrin: sounds good!

LeoGrin (approved):
LGTM! Added a few comments in the threads but they are quite minro

src/tabpfn/inference.py
noahho: done

PR #372, opened 2025-07-06
Consolidate `os.environ.get` via `pydantic-settings`

muneebshahid:
## Motivation and Context

This PR introduces a centralized settings management system for TabPFN using `pydantic-settings`, replacing scattered environment variable usage with a structured configuration approach. 

### New Features
- **Settings Module** (`src/tabpfn/settings.py`): 
  - Centralized configuration management using Pydantic
  - Environment variable support with `TABPFN_` prefix
  - Type-safe settings with validation
  - Support for `.env` files

### Key Configurations Added
- `TABPFN_MODEL_CACHE_DIR`: Custom directory for model caching
- `TABPFN_ALLOW_CPU_LARGE_DATASET`: Override CPU limitations for large datasets
- `PYTORCH_CUDA_ALLOC_CONF`: PyTorch CUDA memory settings
- Testing settings: `FORCE_CONSISTENCY_TESTS`, `CI`

### Code Updates
- Replaced direct `os.environ` calls with settings instance
- Updated model loading to use `settings.tabpfn.model_cache_dir`
- Migrated test configurations to use the settings system

### Dependencies
- Added `pydantic-settings>=2.10.1`
- Added `eval-type-backport>=0.2.2` (for Python 3.9 compatibility)
- Updated `typing_extensions>=4.12.0`

## Breaking Changes
None - all existing environment variables continue to work as before.

## Future Work

This PR can be followed by refactoring system/tool level variables in a separate settings class as well e.g.

  - System variables: `CONDA_EXE`, `SYSTEMROOT`, `CUDA_PATH`, `CUDNN_LIBRARY`, `PROGRAMFILES`
  - PyTorch-specific: `CUDA_MODULE_LOADING`, `PIP_DISABLE_PIP_VERSION_CHECK`

---

## Public API Changes

-   [x] No Public API changes

Note: While this PR adds support for .env files, all environment variables already existed and continue to work exactly as before. This is purely an internal refactoring that adds the convenience of .env file support.

---

## How Has This Been Tested?

- Existing tests pass without modification, confirming backward compatibility
- All previously existing environment variables continue to work as before:
    - `TABPFN_MODEL_CACHE_DIR`
    - `TABPFN_ALLOW_CPU_LARGE_DATASET`
    - `TABPFN_PYTORCH_CUDA_ALLOC_CONF`
    - `FORCE_CONSISTENCY_TESTS`
    - `CI`
- Added `.env` file support tested manually
- Verified settings are properly loaded from both environment variables and .env files
- Type checking passes with new Pydantic models

---

## Checklist

---


src/tabpfn/settings.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        if (
+            isinstance(v, str)
+            and not v.strip()
+            and info.field_name
+            and (field_info := cls.model_fields.get(info.field_name))
+            and field_info.default is not PydanticUndefined
+        ):
+            return field_info.default
+        return v
```
muneebshahid: This is purely here to keep the previous behavior as is i.e. empty strings result in None or default values. 

We can also simply remove this and stick with default pydantic settings behavior which is to try and use empty string or raise exception if not possible.

src/tabpfn/misc/debug_versions.py
```diff
@@ -606,7 +607,7 @@ def _pretty_str(envinfo):
     return env_info_fmt.format(**mutable_dict)
 
 
-def _get_pretty_env_info():
+def _get_pretty_env_info():  # Not used anywhere??
```
muneebshahid: May be we remove this method?

pyproject.toml
```diff
@@ -8,11 +8,13 @@ version = "2.0.9"
...
   "scikit-learn>=1.2.0,<1.7",
-  "typing_extensions>=4.4.0",
+  "typing_extensions>=4.12.0",
   "scipy>=1.11.1,<2",
   "pandas>=1.4.0,<3",
   "einops>=0.2.0,<0.9",
-  "huggingface-hub>=0.0.1,<1"
+  "huggingface-hub>=0.0.1,<1",
+  "pydantic-settings>=2.10.1",
+  "eval-type-backport>=0.2.2",
```
oscarkey: Could you add a note that this is only required for Python 3.9, so we remember to remove it later?

src/tabpfn/misc/debug_versions.py
oscarkey: yes sounds good, please remove!

src/tabpfn/model/memory.py
```diff
@@ -11,8 +11,10 @@
 import numpy as np
 import torch
 
+from tabpfn.settings import settings
+
 # TODO(eddiebergman): Make this an option
```
oscarkey:
```suggestion
```

src/tabpfn/model/memory.py
```diff
@@ -11,8 +11,10 @@
...
-os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "max_split_size_mb:512"
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = settings.tabpfn.pytorch_cuda_alloc_conf
```
oscarkey: Another option would be to check if PYTORCH_CUDA_ALLOC_CONF, and set it to the default if it's not set. This would mean people could use PYTORCH_CUDA_ALLOC_CONF as they usually do. I don't know if this is more or less confusing than the current approach aha, what do you think?

As this a new setting/behaviour, it could also be added in a follow-up PR, if you'd like to get this one merged :)

oscarkey:
bugbot run

oscarkey:
bugbot run 🤞🏻

src/tabpfn/settings.py
oscarkey: I think we can safely change the behaviour to the pydantic settings default. While it's technically a breaking change, I think it's very unlikely anyone is relying on this behaviour? It will make things quite a bit simpler.

noahho:
@muneebshahid would be amazing to take another look if you find time! :-)

muneebshahid:
@noahho @oscarkey Thank you both for the feedback. Sorry for the delay. I'll have a look either today or tomorrow and patch it up.  🚀

noahho:
No rush Muneeb, just wondered as this would be cool to have and sometimes things tend to just be forgotten!

src/tabpfn/misc/debug_versions.py
muneebshahid: Removed

src/tabpfn/settings.py
muneebshahid: Removed

pyproject.toml
muneebshahid: Added

src/tabpfn/model/memory.py
muneebshahid: Good catch.

In order to simplify things and keep the behavior as before I have instead introduced `PytorchSettings`.

https://github.com/PriorLabs/TabPFN/blob/1056aab7869c60eebe102b60ab4ae20c7257f74c/src/tabpfn/settings.py#L38-L45

The neat thing is you don't need to set `os.environ["PYTORCH_CUDA_ALLOC_CONF"]` to the default value anymore, pydantic will do this for you.

You can verify this via running

```python
from tabpfn.settings import settings
import os

print(os.environ["PYTORCH_CUDA_ALLOC_CONF"])
```

---

Meaning the following can be removed.

https://github.com/PriorLabs/TabPFN/blob/1056aab7869c60eebe102b60ab4ae20c7257f74c/src/tabpfn/model/memory.py#L16

But you need to have imported settings at least once i.e. if the settings is not imported to at least once in the path, the environment variable won't be set. Although i do see it being imported in `base.py` i'll leave it up to you if we should remove it or not.

pyproject.toml
```diff
@@ -64,7 +66,7 @@ source = "https://github.com/priorlabs/tabpfn"
 dev = [
   # Lint/format
   "pre-commit",
-  "ruff",
+  "ruff==0.8.6",
```
muneebshahid, line 69: The project is formatted via ruff `0.8.6` it seems. Else my local setup installs the latest one which formats very differently

oscarkey: Ah thank you! We actually have an internal change which improves this, but we haven't pushed it yet, so this is a good fix.

src/tabpfn/misc/debug_versions.py
```diff
@@ -391,6 +392,8 @@ def _get_pip_packages(run_lambda, patterns=None):
         run_lambda,
         [sys.executable, "-mpip", "list", "--format=freeze"],
     )
+    if out is None:
```
oscarkey, line 395: Could you add a comment about when/why this happens?

README.md
```diff
@@ -245,6 +245,25 @@ A: Try the following:
...
 
+**Q: What environment variables can I use to configure TabPFN?**
+A: TabPFN uses Pydantic settings for configuration, supporting environment variables and `.env` files:
+
+**Model Configuration:**
+- `TABPFN_MODEL_CACHE_DIR`: Custom directory for caching downloaded TabPFN models (default: platform-specific user cache directory)
+- `TABPFN_ALLOW_CPU_LARGE_DATASET`: Allow running TabPFN on CPU with large datasets (>1000 samples). Set to `true` to override the CPU limitation. Note: This will be very slow!
+
+**PyTorch Settings:**
+- `PYTORCH_CUDA_ALLOC_CONF`: PyTorch CUDA memory allocation configuration to optimize GPU memory usage (default: `max_split_size_mb:512`)
```
oscarkey: Could you add a link to https://docs.pytorch.org/docs/stable/notes/cuda.html#optimizing-memory-usage-with-pytorch-cuda-alloc-conf , so people can reference this for more information ?

oscarkey:
bugbot run

pyproject.toml
oscarkey: thank you!

src/tabpfn/model/memory.py
oscarkey: ah this is great, thank you!
I think it's best to keep this line for now, to make it explicit?

src/tabpfn/misc/debug_versions.py
oscarkey: nice!

oscarkey:
bugbot run

oscarkey (approved):

src/tabpfn/model/memory.py
muneebshahid: Yea i agree. I'd leave it as is for now.

README.md
muneebshahid: Done

src/tabpfn/misc/debug_versions.py
muneebshahid: Added

noahho (approved):
Thanks you so much @muneebshahid this looks great!

PR #403, opened 2025-07-21
Fix the class_permutation issue when predicting with differentiable_input

benraha:
## Motivation and Context
While trying to wrap a KV-Cached model with ONNX, I came across the option `differentiable_input`.
However, when trying to predict with this option, I got the following error:

```
            # Process the config_list (which is now guaranteed to be a list)
            if config_list is not None:
                output_batch = []
                for i, batch_config in enumerate(config_list):
                    assert isinstance(batch_config, ClassifierEnsembleConfig)
                    # make sure the processed_output num_classes are the same.
>                   if len(batch_config.class_permutation) != self.n_classes_:
E                   TypeError: object of type 'NoneType' has no len()

src/tabpfn/classifier.py:842: TypeError
```

This PR fixes the issue, when class_permutation is `None`, no class permutation exist and we can return only the relevant classes.
---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?
Added a test
---

## Checklist

---


benraha:
@noahho @LeoGrin, I would love your feedback on that, and please let me know if you'd like me to follow Gemini's suggestion.

src/tabpfn/classifier.py
```diff
@@ -838,15 +838,18 @@ def forward(  # noqa: C901, PLR0912
                 output_batch = []
                 for i, batch_config in enumerate(config_list):
                     assert isinstance(batch_config, ClassifierEnsembleConfig)
-                    # make sure the processed_output num_classes are the same.
-                    if len(batch_config.class_permutation) != self.n_classes_:
-                        use_perm = np.arange(self.n_classes_)
-                        use_perm[: len(batch_config.class_permutation)] = (
-                            batch_config.class_permutation
-                        )
+                    if batch_config.class_permutation is None:
```
noahho: Could you add a comment on when the class permutation is set to None?

noahho:
@benraha thanks a lot! I don't think Geminis comment is relevant here and I resolved it.

src/tabpfn/classifier.py
benraha: Added :) Also, removed the config_list check for None, since it can't happen (and never happened since the process would fail since output_all wasn't defined

src/tabpfn/classifier.py
```diff
@@ -838,15 +838,18 @@ def forward(  # noqa: C901, PLR0912
...
+                        output_batch.append(processed_output[:, i, : self.n_classes_])
                     else:
```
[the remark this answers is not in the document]
benraha: Looks fine to me ...

noahho:
@benraha could you please sign the CLA?

benraha:
@noahho I have already, twice!! signed again now ... can you check?

noahho:
Hi @benraha absolutely, I see it's signed. Quickly figuring out whats going on, thanks for the patience :-)

noahho:
@benraha I'm a bit worried about the change in the reference predictions. We are very careful when changing these. The difference seems quite small, was the change necessary to pass the test?  If not, could you remove that file from the PR please.

benraha:
Since we are using the differentiable_input, I didn't want to compare it to the regular version, but rather to itself (because the preprocessing is mostly off) ...  I think I'll add a new reference file; what do you think?

noahho:
This now looks great to me, thank you so much, will be merged after tests pass.

noahho (approved):

PR #411, opened 2025-07-24
Use FunctionTransformer in the standard way

benraha:
The standard way to create a sklearn transformer that does nothing is to pass None in the constructor to FunctionTransformer. When converting sklearn pipelines to ONNX format, this is also important as arbitrary Python functions are not supported, so that's also a step in that direction.

## Motivation and Context

---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?
Locally
---

## Checklist

-   [] Documentation has been updated (if the public API or usage changes).
-   [] An entry has been added to `CHANGELOG.md` (if relevant for users).

---

### DIFF
--- a/src/tabpfn/model/preprocessing.py
+++ b/src/tabpfn/model/preprocessing.py
@@ -743,12 +743,14 @@ class ReshapeFeatureDistributionsStep(FeaturePreprocessingTransformerStep):
                     ),
                     (
                         "ordinal",
+                        # default FunctionTransformer yields the identity function
                         FunctionTransformer(),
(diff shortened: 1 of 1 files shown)


benraha:
@noahho Hi again, I've signed the CLA once again. I'm not sure why it is not showing up. Would love your review!

benraha:
@noahho I get this ruff error, but I can't figure out the problem, and running the pre-commit or ruff check directly works... Can you help?

noahho:
@benraha just fixed ruff, I also broke it with my commit before :-)

PR #421, opened 2025-07-31
Push internal changes to the public repository.

oscarkey:
Main changes:
* Move tabpfn.model -> tabpfn.architectures.base
* Change PerFeatureTransformer.forward() interface:
  * Remove single_eval_pos: this is now inferred from the length of the targets
  * Only accept a single argument format: `def forward(x, y, **kwargs)`

## Public API Changes
Yes, Public API changes:
* tabpfn.model has moved to tabpfn.architectures.base, but stub files are left at the old locations for backwards compatibility.
* The type hint of `TabPFNRegressor.config_` and `TabPFNClassifier.config_` has changed from `ModelConfig` to `ArchitectureConfig`, but the runtime type remains unchanged as `ModelConfig`.


noahho:
LGTM!

Issue #422, opened 2025-08-01
infer_categorical_features crashes if input dataframe contains nan and str values for a feature


LennartPurucker:
This [code](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/utils.py#L501) crashes if one tries to pass a feature as input that contains str and nan values. 

The reason is that `np.unique(col)` does not work if `col=pd.Series([np.nan, "NA"])`:

```python
import numpy as np
import pandas as pd

np.unique(pd.Series([np.nan, "NA"]))
```

As the function indicates, it should only get numpy arrays, but it seems that is not the case, and thus, it crashes.

LennartPurucker:
This assumes the data is not ordinal encoded first, but [by the code later](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/classifier.py#L514-L517), and one uses numpy 2.1.3.

LennartPurucker:
Potential solution:

```python
for ix, col in enumerate(X.T):
    nan_vals = pd.Series(col).isna()
    nan_offset = 1 if any(nan_vals) else 0
    if ix in maybe_categoricals:
        if (len(np.unique(col[~nan_vals])) + nan_offset)  <= max_unique_for_category:
            indices.append(ix)
    elif (
        large_enough_x_to_infer_categorical
        and (len(np.unique(col[~nan_vals])) + nan_offset)   < min_unique_for_numerical
    ):
        indices.append(ix)
```

noahho:
Thanks a lot for flagging this and providing a solution! I just created a PR: https://github.com/PriorLabs/TabPFN/pull/432 but didn't manage to add a test case yet, if you feel like it, please improve this :-)

klemens-floege:
Since we merged [PR #432](https://github.com/PriorLabs/TabPFN/pull/432), I mark this issue as resolved. :)

PR #432, opened 2025-08-03
Fix infer cat nans

noahho:
## Motivation and Context

This pull request resolves \[Issue #422\] by implementing \[briefly describe the core change, e.g., "a more efficient data serialization method"\]. The primary motivation is to \[explain the high-level goal, e.g., "reduce application startup time" or "fix a critical bug causing data corruption"\]. This change will improve \[mention the benefit, e.g., "system performance and user experience"\].

* * *

## Public API Changes

-   \[x\] No Public API changes
-   \[ \] Yes, Public API changes (Details below)
    
* * *

## How Has This Been Tested?

These changes have been thoroughly tested to ensure correctness and prevent regressions.

* * *

## Checklist

-   \[x\] The changes have been tested locally.
-   \[x\] Documentation has been updated (if the public API or usage changes).
-   \[x\] A entry has been added to `CHANGELOG.md` (if relevant for users).
-   \[x\] The code follows the project's style guidelines.
-   \[x\] I have considered the impact of these changes on the public API.


src/tabpfn/utils.py
```diff
@@ -536,12 +536,20 @@ def infer_categorical_features(
...
 
     for ix, col in enumerate(X.T):
+        # Calculate total distinct values once, treating NaN as a category.
+        try:
+            s = pd.Series(col)
+            # counts NaN/None as a category
+            num_distinct = s.nunique(dropna=False)
+        except TypeError as e:
+            # e.g. "unhashable type: 'dict'" when object arrays contain dicts
+            raise TypeError("argument must be a string or a number") from e
```
priorphil: This error message might be hard to parse - IIUC, it's a check that the column only contains strings or numbers? If so, maybe `TypeError("Columns must only contain strings or numbers")` would be better?

tests/test_utils.py
```diff
@@ -43,3 +46,63 @@ def get_memory() -> None:
         t.join()
     psutil_result = psutil.virtual_memory().total / 1e9
     assert all(result == psutil_result for result in results)
+
+
+def test_infer_categorical_with_str_and_nan_provided_included():
```
priorphil, line 51: Love tests! Can we add one with a dictionary in a column to verify that this raises an error?

src/tabpfn/utils.py
klemens-floege: Done

tests/test_utils.py
klemens-floege: Added this: 


```
def test_infer_categorical_with_dict_raises_error():
    X = np.array([[{"a": 1}], [{"b": 2}]], dtype=object)
    with pytest.raises(TypeError):
        infer_categorical_features(
            X,
            provided=None,
            min_samples_for_inference=0,
            max_unique_for_category=2,
            min_unique_for_numerical=2,
        )
```

priorphil: image

PR #440, opened 2025-08-06
Adding GPU testing to our CI pipeline

noahho:
### Motivation and Context

To ensure `TabPFN` is robust, it's crucial to validate its performance on both CPU and GPU hardware. The previous CI setup only performed tests on CPUs. 

Overall test time has not increased while GPU tests are only run if CPU gating tests passed. Workflow file complexity has increased.

This series of changes addresses that gap by:
1.  **Adding a Dedicated GPU Test Job:** A new workflow job now runs on a GPU-enabled runner (`ubuntu-22.04-4core-gpu`) to test CUDA functionality. This runner is a custom runner hosted by us through GitHub which we get through our enterprise GitHub tier.
2.  **Implementing a Gated Workflow:** To manage costs, the expensive GPU job is gated. It will only run after a prerequisite CPU job (Ubuntu with Python 3.13) passes successfully. That CPU test is the fastest one so we want to only depend on it.
3.  **Using device more consistently in tests:** The test suite has been made more device-aware. We now use an environment variable (`TABPFN_EXCLUDE_DEVICES`) to ensure that the GPU runner only executes CUDA-relevant tests, avoiding redundant CPU test runs.  Hardcoded `device="cpu"` parameters within the tests have been updated to `device="auto"` or use a dynamic utility function. This allows the same test suite to run seamlessly across different hardware backends.

**Tradeoff in implementation:** To make the GPU run only depend on ubuntu finishing (and not slower mac and windows), I had to copy the system into a separate workflow which meant duplication. Couldn't find another way though.

* * *

### Public API Changes
-   \[x\] No Public API changes
-   \[ \] Yes, Public API changes (Details below)
* * *

### How Has This Been Tested?

CI runs.

Floating point precision tolerances in one test were slightly adjusted to account for minor numerical differences between CPU and GPU calculations, which is an expected part of this cross-device validation.

* * *

### Checklist

-   \[x\] The changes have been tested locally.
-   \[ \] Documentation has been updated (if the public API or usage changes).
-   \[ \] A entry has been added to CHANGELOG.md (if relevant for users).
-   \[x\] The code follows the project's style guidelines.
-   \[x\] I have considered the impact of these changes on the public API.


tests/test_classifier_interface.py
```diff
@@ -229,15 +222,7 @@ def test_predict_logits_and_consistency(
         # softmax to the averaged logits will NOT match predict_proba.
```
priorphil, line 222: Nit: just remove the else branch?

.github/workflows/pull_request.yml
```diff
@@ -34,9 +36,7 @@ jobs:
           - os: windows-latest
             python-version: "3.9"
             dependency-set: minimum
-          - os: ubuntu-latest
-            python-version: "3.13"
-            dependency-set: maximum
+          # REMOVED ubuntu-latest 3.13 here which is executed in separate workflow
```
oscarkey:
```suggestion
          # ubuntu-latest 3.13 maximum is not included here because it is executed
          # as a separate workflow below. This is because we wish to gate the GPU
          # workflow on ubuntu-latest 3.13 maximum, which requires it to be a
          # separate workflow.
```

.github/workflows/pull_request.yml
```diff
@@ -80,6 +77,88 @@ jobs:
...
+      - name: Initialize submodules
+        run: git submodule update --init --recursive
+
+      - name: Run Tests (exclude MPS params)
+        env:
+          TABPFN_EXCLUDE_DEVICES: "mps"
+        run: pytest tests/
+
+  # -------------------------------------------------------------------
+  # GPU: depends on the Ubuntu 3.13 job above
```
oscarkey:
```suggestion
  # GPU: To save compute we only want to execute this workflow once
  # a CPU workflow has passed. Hence, this workflow depends on the
  # ubuntu-latest 3.13 workflow above.
```

oscarkey (approved):
cool!

PR #437, opened 2025-08-05
Run tests with all models #433

martino-vic:
## Motivation and Context

- See issue #433.
- After running tests locally, the file `tests/reference_predictions/inconsistency_test_predictions.json` is being modified. I haven't included those changes in this PR.
---

## Public API Changes

-   [x] No Public API changes
-   [ ] Yes, Public API changes (Details below)

---

## How Has This Been Tested?
- Running all tests with pytest locally.
---

## Checklist

-   [x] The changes have been tested locally.
-   [ ] Documentation has been updated (if the public API or usage changes).
-   [ ] A entry has been added to `CHANGELOG.md` (if relevant for users).
-   [x] The code follows the project's style guidelines.
-   [x] I have considered the impact of these changes on the public API.

---

PR #449, opened 2025-08-13
Test all models

noahho:
## Motivation and Context

This patch expands our test coverage to run against **all available TabPFN models** while keeping CI time reasonable.
-   Introduce `ModelSource` in tests to **programmatically fetch model filenames**.
-   Restructure parameter grids so we **fully exercise one “primary” model** (broad grid) and run a **fast “smoke” sweep** across the remaining models.
-   Reduce reliance on hardcoded `"auto"` model selection inside tests to make failures attributable to specific model artifacts.

This addresses gaps identified in “Run tests with all models” and aligns with feedback that initialization and interface tests should validate **each** shipped model.
* * *
## Public API Changes
-    **No Public API changes**
(Only test code is modified.)
* * *

## How Has This Been Tested?
-   Confirmed the new grids keep test duration practical on CPU-only environments.

* * *

## Checklist
-    The changes have been tested locally.
-    Documentation updated: _N/A (tests only)_.
-    `CHANGELOG.md` entry: _N/A (tests only)_.
-    Code follows project style (ruff format & ruff check).
-    Considered impact on public API: _None (tests only)_.

* * *

## Implementation Notes
-   Built:
    -   `_full_grid`: exhaustive combos **only** for the first model path.
    -   `_smoke_grid`: a single, fast combo for **each remaining** model path.
-   Combined as `all_combinations = list(_full_grid) + list(_smoke_grid)`.

* * *
## Prior Work / Acknowledgements

Big thanks to the work in **PR #437 “Run tests with all models” (by @martino-vic)** which motivated this direction and highlighted the need to validate every shipped model. That PR is **good prior art but currently not mergeable** (pending checks/CLA and minor integration issues). This patch folds in the workable pieces (programmatic model discovery + targeted grid split) and aligns them with the current test suite and style rules.


LeoGrin:
bugbot run

tests/test_classifier_interface.py
```diff
@@ -47,19 +49,38 @@
 inference_precision_methods = ["auto", "autocast", torch.float64, torch.float16]
 remove_outliers_stds = [None, 12]
 estimators = [1, 2]
+model_paths = ModelSource.get_classifier_v2().filenames
+
+primary_model = model_paths[0]
+other_models = model_paths[1:]
```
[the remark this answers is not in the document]
noahho: I don't think that's an issue. We should have always some models to run

tests/test_classifier_interface.py
```diff
@@ -47,19 +49,38 @@
 inference_precision_methods = ["auto", "autocast", torch.float64, torch.float16]
 remove_outliers_stds = [None, 12]
 estimators = [1, 2]
+model_paths = ModelSource.get_classifier_v2().filenames
+
+primary_model = model_paths[0]
```
LeoGrin: the first one is the one we get by default right? is it safer to check it has the expected name so the test behavior doesn't change if the ordering changes? Not sure

tests/test_classifier_interface.py
```diff
@@ -743,7 +767,7 @@ def test_initialize_model_variables_classifier_sets_required_attributes() -> Non
     assert norm_criterion is None, "norm_criterion should be None for classifier"
 
     # 2) Test the sklearn-style wrapper on TabPFNClassifier
-    classifier = TabPFNClassifier(model_path="auto", device="cpu", random_state=42)
+    classifier = TabPFNClassifier(device="cpu", random_state=42)
```
LeoGrin, line 770: should it be "model_path=model_path" here or am I missing somethign?

LeoGrin (approved):
LGTM, just two comments

tests/test_classifier_interface.py
LeoGrin: same for regression

LeoGrin:
not sure about the failing test, probably just easier to go over the precision limit by random change if we test more models right?

tests/test_classifier_interface.py
noahho: Here it just uses the default model, it's a specific test that we aren't doing for all checkpoints. Only the basic tests above are run with the model_path parameterized to save time

PR #451, opened 2025-08-14
Enable type annotation rules in the Ruff linter.

oscarkey:
Enable the ANN rules, but exclude some overly strict ones, in line with our other repositories.

I added type annotations where they were missing. I also:
- ignore ANN for src/tabpfn/architectures/base/preprocessing.py, due to a lot of errors
- delete a class and a method that were not used in any PriorLabs repository
- add noqa in some cases where the type was not obvious, to avoid getting it wrong

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -10,7 +10,7 @@ from collections.abc import Iterator, Sequence
 from copy import deepcopy
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, NoReturn
 from typing_extensions import override
 
 import joblib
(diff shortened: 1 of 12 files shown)


src/tabpfn/inference.py
```diff
@@ -83,7 +83,7 @@ def iter_outputs(
         """
         ...
 
-    def use_torch_inference_mode(self, *, use_inference: bool):
+    def use_torch_inference_mode(self, *, use_inference: bool) -> NoReturn:
```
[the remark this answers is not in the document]
oscarkey: hmmm interesting. Gemini seems correct to me, but the NoReturn annotation was actually added automatically by `ruff check --fix`. I will change it to `None`.

tests/test_consistency.py
```diff
@@ -325,7 +325,7 @@ def save_platform_metadata(cls):
             json.dump(metadata, f, indent=2)
 
     @classmethod
-    def load_platform_metadata(cls):
+    def load_platform_metadata(cls) -> dict:
```
priorphil: should we do `dict[str, Any]`?

priorphil (approved):
Thanks for adding these!

tests/test_consistency.py
oscarkey: Seems this function wasn't called anywhere, so I deleted it instead.

PR #459, opened 2025-08-20
Changed the CPU attention mechanism to use PyTorch implementation

benraha:
## Motivation and Context

This pull request resolves a performance bottleneck on CPU. Previously, the attention dot product utilized a naive `einsum` implementation, which was significantly slower than the optimized PyTorch version. By replacing the naive version with PyTorch's version, we have achieved a **65% improvement in CPU inference speed.**

The naive implementation is still there for PyTorch versions earlier than 2.0. Since the project's minimum requirement is version 2.1, this code is now dead. Let me know if you want me to clean it up.

---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?
On my local Mac - for inference of 500 records, with and without KV cache, the results are consistent.

The local initialisation looks like this:

```
classifier = TabPFNClassifier(
    model_path="tabpfn-v2-classifier.ckpt",
    n_estimators=8,
    device="cpu",
    random_state=42,
    fit_mode="fit_with_cache", # both with and without
    memory_saving_mode=False,
)
```
The memory saving mode is turned off to get pure results.
---

## Checklist

---

### DIFF
--- a/src/tabpfn/architectures/base/attention/full_attention.py
+++ b/src/tabpfn/architectures/base/attention/full_attention.py
@@ -534,6 +534,7 @@ class MultiHeadAttention(Attention):
         assert q is not None
         assert k is not None
         assert v is not None
+
         batch_size, seqlen_q, nhead, d_k = q.shape
         _, seqlen_kv, nhead_kv, d_v = v.shape
         share_kv_across_n_heads = nhead // nhead_kv
@@ -547,11 +548,9 @@ class MultiHeadAttention(Attention):
         )
 
         # this string comparison is reliable, as it does not compare to a subversion
-        TORCH_2_ATTENTION_POSSIBLE = (
-            torch.__version__ >= "2" and torch.cuda.is_available()
-        )
+        TORCH_2_ATTENTION_POSSIBLE = torch.__version__ >= "2"
         USE_TORCH_2_GQA = False
-        if TORCH_2_ATTENTION_POSSIBLE:
+        if TORCH_2_ATTENTION_POSSIBLE and torch.cuda.is_available():
             # check whether torch.nn.functional.scaled_dot_product_attention has a
             # kwarg enable_gqa
             # Check if enable_gqa is supported by trying to call the function with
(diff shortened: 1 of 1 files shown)


src/tabpfn/architectures/base/attention/full_attention.py
```diff
@@ -563,17 +562,13 @@ def compute_attention_heads(  # noqa: C901, PLR0912
                     torch.empty(1, 1, 1, 1),
```
benraha: Since we are already here, this can be solved much cleaner using `inspect` ... I can do it if you want ...

priorphil: That'd be awesome, thanks! :)

src/tabpfn/architectures/base/attention/full_attention.py
```diff
@@ -547,11 +548,9 @@ def compute_attention_heads(  # noqa: C901, PLR0912
         )
 
         # this string comparison is reliable, as it does not compare to a subversion
-        TORCH_2_ATTENTION_POSSIBLE = (
-            torch.__version__ >= "2" and torch.cuda.is_available()
-        )
+        TORCH_2_ATTENTION_POSSIBLE = torch.__version__ >= "2"
```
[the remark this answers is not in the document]
priorphil: I'd be up for adding this as well, since string comparison really isn't reliable :)

priorphil (approved):
Thanks a lot for this PR and the cleanup, looks a lot nicer!

benraha:
@priorphil made the changes, but using `inspect` didn't work, so I changed it to check the torch version instead ... looks like this capability [was added in torch 2.5](https://github.com/pytorch/pytorch/commit/8bc5ef563eab08bfa9e48c6a546aba994ab7828c#diff-c358c214f663ba0c8b9c6846fbe0042fa29494cf02fe4714a17dcd0d268b035b) (also verified in the docs). Let me know what you think.

src/tabpfn/architectures/base/attention/full_attention.py
```diff
@@ -28,6 +28,38 @@
 except (ModuleNotFoundError, ImportError):
     HAVE_FLASH_ATTN = False
 
+TORCH_VERSION = torch.__version__.split(".")
+
+TORCH_2_ATTENTION_POSSIBLE = int(TORCH_VERSION[0]) >= 2
+
+
+def _check_torch_gqa_support() -> bool:
```
priorphil: Nit: Let's rename the function to directly indicate what the returned bool means
```suggestion
def _gqa_is_supported() -> bool:
```

src/tabpfn/architectures/base/attention/full_attention.py
```diff
@@ -28,6 +28,38 @@
...
+
+    # Check compute capability only if CUDA is available
+    # We need compute capability >= 8.0 for efficient GQA
+    device = torch.cuda.current_device()
+    nvidia_compute_capability = torch.cuda.get_device_capability(device)
+    return nvidia_compute_capability[0] >= 8
+
+
+# Cache the GQA support check at module level
+USE_TORCH_2_GQA = _check_torch_gqa_support()
```
priorphil:
```suggestion
USE_TORCH_2_GQA = _gqa_is_supported()
```

benraha:
@priorphil committed the suggestions. I'm a big fan of correct names! (Which reminds me of my confusion when I read "broadcast_kv_across_heads" doesn't broadcast anything, but I don't like changing parts that I'm not touching directly ;-)

priorphil:
Yeah, we're also slowly working on improving the code base in the areas we're touching, but as you might guess this can take a while ;)

priorphil:
Thanks again for the contribution!

PR #468, opened 2025-08-26
Improve column selection in RemoveEmptyFeaturesEncoderStep

benraha:
## Motivation and Context

This PR improves the encoder step that removes non-empty columns (RemoveEmptyFeaturesEncoderStep). The `select_features`  function was generating 1500 steps in an ONNX graph for the model, which were 30% of the total steps in the graph for a use case with 50 columns.

The updated approach is optimized for the more common scenario where no features are constant. When constant features do exist, the new logic removes them rather than rebuilding the entire tensor, significantly reducing the number of graph nodes.

Additionally, a comment in the code suggests this step is a no-op, but it demonstrably removes empty features. The comment has been updated with a question mark to highlight this discrepancy and request clarification.

---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?

Locally, with datasets that contain constant features.

---

## Checklist

---

### DIFF
--- a/src/tabpfn/architectures/base/encoders.py
+++ b/src/tabpfn/architectures/base/encoders.py
@@ -233,25 +233,26 @@ def select_features(x: torch.Tensor, sel: torch.Tensor) -> torch.Tensor:
         The shape is (sequence_length, batch_size, total_features) if batch_size is greater than 1.
     """
     B, total_features = sel.shape
-    sequence_length = x.shape[0]
+
+    # Do nothing if we need to select all of the features
+    if torch.all(sel):
+        return x
 
     # If B == 1, we don't need to append zeros, as the number of features don't need to be fixed.
     if B == 1:
         return x[:, :, sel[0]]
 
-    new_x = torch.zeros(
-        (sequence_length, B, total_features),
-        device=x.device,
-        dtype=x.dtype,
-    )
+    new_x = x.detach().clone()
 
     # For each batch, compute the number of selected features.
     sel_counts = sel.sum(dim=-1)  # shape: (B,)
 
     for b in range(B):
         s = int(sel_counts[b])
-        if s > 0:
-            new_x[:, b, :s] = x[:, b, sel[b]]
+        if s != total_features:
+            if s > 0:
+                new_x[:, b, :s] = x[:, b, sel[b]]
+            new_x[:, b, s:] = 0
 
     return new_x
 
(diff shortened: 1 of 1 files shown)


src/tabpfn/architectures/base/encoders.py
```diff
@@ -233,25 +233,26 @@ def select_features(x: torch.Tensor, sel: torch.Tensor) -> torch.Tensor:
...
     # If B == 1, we don't need to append zeros, as the number of features don't need to be fixed.
     if B == 1:
         return x[:, :, sel[0]]
 
-    new_x = torch.zeros(
-        (sequence_length, B, total_features),
-        device=x.device,
-        dtype=x.dtype,
-    )
+    new_x = x.detach().clone()
```
priorphil, line 245: What's the reason for this change? My intuition would say filling the tensor with zeros should be faster than copying the exact values :)

benraha: But we are copying the values anyway - I just changed that so we're only moving what needs to be moved and not the entire tensor

benraha:
@priorphil also, please review the comment change - I can change it to whatever you think, but now it's just incorrect and misleading ...

src/tabpfn/architectures/base/encoders.py
priorphil: Yes, but with this we're copying it twice!

priorphil:
Yeah, that comment is really confusing. I'd say let's remove the whole ```Was changed to NOT DO ANYTHING(?), the removal of empty features now
    done elsewhere, but the saved model still needs this encoder step.
    TODO: REMOVE.``` part and just keep the first line summary.
:)

Also, feel free to add an entry to the Changelog.md!

src/tabpfn/architectures/base/encoders.py
benraha: I see what you mean, but if we agree that having a constant feature is not a common scenario, then the old way would have copied almost the entire tensor in a python loop, and now we copy all in one batch and only modify a small part ... want me to benchmark it?

benraha:
@priorphil Agree on the comment, and added this (and my previous PR) to the changelog.

priorphil:
Thanks! Please also see my comment on copying twice

src/tabpfn/architectures/base/encoders.py
benraha: Was curious already :) For this code:

```
x = torch.randn(300, 20, 2)
x[:,12,1] = 1.34
```

The new version runs in 29.3 μs vs 170 μs, using %%timeit

priorphil: Aaah, sorry my bad, I missed the `if s != total_features:` bit - I only registered the torch.all part above as new.
Let's keep your version then!

priorphil:
Merging, thanks again for the contribution!

benraha:
@priorphil Just occurred to me that we don't need to copy at all! I'll submit another PR soon :)

PR #477, opened 2025-09-01
Declare joblib dependency.

oscarkey:
We depend on joblib, see e.g. `tabpfn.preprocessing`, but currently do not declare this (things still work because scikit-learn depends on it).

Specify >=1.2.0, the same as scikit-learn, so the previous minimum version is maintained.


priorphil (approved):
Should we then import joblib directly instead of going the indirection via sklearn?

oscarkey:
oh I didn't even notice that! Updated

PR #484, opened 2025-09-04
[WIP] Evaluate the estimators in parallel.

oscarkey:


oscarkey:
Hey Brendan, I'd love to get your overall thoughts on this draft before I finish it up.

brendan-priorlabs:
@oscarkey, would love to. I can give this a proper pass tomorrow if that works!

src/tabpfn/parallel_evaluate.py
```diff
@@ -0,0 +1,135 @@
...
+
+def parallel_evaluate(
+    devices: Sequence[torch.device],
+    functions: Iterable[Callable[[torch.device], R]],
+) -> Generator[R]:
+    """Evaluate the given functions in parallel across `devices`.
+
+    The function evaluations are parallelised using Python threads, so this will only
+    result in a speed-up if the functions do not hold the global interpreter lock. It
+    works well for functions that spend most of their time executing GPU kernels.
```
brendan-priorlabs: Yep, I buy this explanation. It looks like `_evaluate_estimator` almost immediately calls a Pytorch module so I wouldn't expect there to be much CPU overhead typically, but please do correct my intuition!

src/tabpfn/parallel_evaluate.py
```diff
@@ -0,0 +1,135 @@
...
+    devices: Sequence[torch.device],
+    functions: Iterable[Callable[[torch.device], R]],
+) -> Generator[R]:
+    # Create our own context to avoid conflicts with the caller's.
+    mp_context = multiprocessing.get_context(method="spawn")
+
+    time.time()
+    with mp_context.Manager() as manager:
+        time.time()
+        free_devices = manager.Queue(maxsize=len(devices))
```
brendan-priorlabs: Some LLM discussion has led me to think that using a `Manager.Queue` might be preventing the use of shared memory for moving tensors around which would certainly explain some overhead. What happens with `mp_context.Queue` instead? Not a super helpful issue, but maybe a breadcrumb https://github.com/pytorch/pytorch/issues/30401

src/tabpfn/parallel_evaluate.py
```diff
@@ -0,0 +1,135 @@
...
+    all_devices: Sequence[torch.device],
+    free_devices: queue.Queue[int],
+    function: Callable[[torch.device], R],
+) -> R:
+    device_index = free_devices.get(block=True)
+    try:
+        device = all_devices[device_index]
+        # We use a separate stream per thread so that threads can execute kernels in
+        # parallel.
+        with torch.cuda.stream(torch.cuda.Stream(device=device)):
```
brendan-priorlabs: Question for you, @oscarkey, when asking GPT 5 about this code, one of things it gave me was:

> Missing device context
> You create a stream on device, but if function(device) does ops without explicitly setting the device, they may run on whatever the current thread’s device is (often 0). Wrap with torch.cuda.device(device) so both device and stream are set.

I think this might be needlessly paranoid in that we'd get device mismatch errors if we were somehow messing this up, but I'm wondering if you see any more wisdom in that reply than I do...

src/tabpfn/inference.py
```diff
@@ -211,49 +211,32 @@ def iter_outputs(
...
-                    X_full,
-                    y_train,
-                    only_return_standard_out=only_return_standard_out,
-                    categorical_inds=batched_cat_ix,
-                )
-
-            output = output if isinstance(output, dict) else output.squeeze(1)
-
-            yield output, config
+        outputs = parallel_evaluate(devices, estimator_functions)
```
brendan-priorlabs: So we typically have 4 workers and 8 units of work (estimators), right? We could think a bit harder about moving some computation from the units of work to the workers, e.g., the module copies, but I'm not sure that's worth it given the similar size.

src/tabpfn/parallel_evaluate.py
```diff
@@ -0,0 +1,135 @@
...
+        free_devices.put(device_index, block=False)
+
+    with ThreadPool(processes=len(devices)) as pool:
+        async_results = [
+            pool.apply_async(
+                # We deepcopy the input function to avoid threads mutating each others
+                # state. For example, if the function uses an nn.Module this allows
+                # different threads to move the module to different devices.
+                _execute_function_in_thread,
+                (devices, free_devices, deepcopy(func)),
```
brendan-priorlabs: Worth calling out that `deepcopy` might not be totally foolproof for complicated objects, but I think any issues would come out in testing. So just an fyi in case you start seeing mysterious bugs later

brendan-priorlabs (commented):
Hey @oscarkey, thanks for putting this together! Very clean and readable. I left a few minor comments, but this looks solid overall. The only that _might_ be major is the one on `Manager.Queue`. Keep me posted!

src/tabpfn/parallel_evaluate.py
oscarkey: we may as well! I will fix.

src/tabpfn/parallel_evaluate.py
oscarkey: ahhh I misunderstood that a Manager was necessary in order to share the queue between processes, but I now see that you don't need one. Which is great, because it will also mean we have to wait for fewer processes to start.

src/tabpfn/parallel_evaluate.py
oscarkey: For small datasets our model can unfortunately still spend most of it's time in python code launching kernels on the gpu, and I think it would hold the gil during this time. But for big datasets we spend almost all our time in the flash attention kernel, so it's not a problem.

src/tabpfn/parallel_evaluate.py
oscarkey: ah good point! I'll add a note about this to the pydoc too.

src/tabpfn/parallel_evaluate.py
brendan-priorlabs: The shared memory part might really help speed things up (if it's real!), so lmk what happens!

src/tabpfn/classifier.py
```diff
@@ -146,7 +146,7 @@ def __init__(  # noqa: PLR0913
         balance_probabilities: bool = False,
         average_before_softmax: bool = False,
         model_path: str | Path | Literal["auto"] = "auto",
-        device: str | torch.device | Literal["auto"] = "auto",
+        device: Device | Sequence[Device] | Literal["auto"] = "auto",
```
brendan-priorlabs: I like that you kept backwards compatibility here, btw!

src/tabpfn/parallel_evaluate.py
brendan-priorlabs: Interesting! Did you happen to benchmark the multi-processing code too or was it so slow it wasn't worth it? (Looked at https://www.notion.so/priorlabs/Fit-estimators-in-parallel-25b5be1f3b49802ea0d5eafff785f5d8, but didn't see it.) I'm imagining possibly dispatching to multi-processing specifically for small datasets. Though maybe startup cost would be too high... More an idea for future work

src/tabpfn/parallel_evaluate.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+            for func in functions
+        ]
+        for async_result in async_results:
```
brendan-priorlabs: I just read your old notes that mentioned the straggler problem. We might be able to do something about that actually. Use `imap_unordered` like:

```
    for result in pool.imap_unordered(fn, items, chunksize=1):
        yield result
```

That'll return the results as they're completed. You might have to do something like return the index of the estimator (if the order matters and we're assuming it doesn't change). I think the utility of this will also depend on how much work we subsequently do before "joining" and operating on the results of all the estimators together. Still, worth a shot!

src/tabpfn/inference.py
```diff
@@ -458,69 +441,39 @@
...
-                    model=self.model,
-                    X=X_full,
-                    cache_kv=False,
-                    device=device,
-                    dtype_byte_size=self.dtype_byte_size,
-                    safety_factor=1.2,  # TODO(Arjun): make customizable
-                )
-            else:
-                pass
+            self.model.type(self.force_inference_dtype)
```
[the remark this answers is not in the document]
oscarkey: This is not true, `.type()` modifies the module in place: https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.type

src/tabpfn/parallel_evaluate.py
oscarkey: Good morning! I never ran the full benchmark, but from my testing I saw 4-8 seconds of startup time per process (unsure if they can start in parallel). As this makes multiprocessing difficult to use for small datasets, and for large datasets the multithreading seems to be efficient, maybe it's okay just to stick with multithreading for now? And return to this if we want better performance later.

src/tabpfn/parallel_evaluate.py
oscarkey: Because of the long startup time for the processes, I was thinking about leaving this for future work? Happy to do it now though too of course!

src/tabpfn/inference.py
oscarkey: I refactored a bit and the module copying now happens in the workers. I suspect nothing else is a bottleneck, but I haven't profiled.

src/tabpfn/parallel_evaluate.py
oscarkey: I had a look and unfortunately the results are almost immediately stacked into a tensor after being returned :(

oscarkey:
This is now ready for a full review!

src/tabpfn/parallel_evaluate.py
brendan-priorlabs: Mornin'! I think that's totally fine if you're already getting the speedup you need. Please view my ideas on this PR more as "ideas to try" than performance improvement demands ;)

src/tabpfn/parallel_evaluate.py
brendan-priorlabs: Future works sounds fine! I think the way around this would be for the classifier and regressor to accept an optional process pool and in Vertex AI we would have long-lived process pools. So this certainly seems fairly surmountable. I might add a TODO in the code or list this as an idea in Notion/Linear somewhere so we have it in the back of our minds

src/tabpfn/parallel_evaluate.py
brendan-priorlabs: Appreciate you checking!

src/tabpfn/parallel_evaluate.py
oscarkey: oooh yes that would be very cool!

src/tabpfn/parallel_execute.py
```diff
@@ -0,0 +1,106 @@
...
+from collections.abc import Generator, Iterable, Sequence
+from multiprocessing.pool import ThreadPool
+from typing import Generic, Protocol, TypeVar
+
+import torch
+
+R_co = TypeVar("R_co", covariant=True)
+
+
+class Function(Protocol, Generic[R_co]):
```
brendan-priorlabs: nit: Might want a slightly more specific name

brendan-priorlabs (approved):
LGTM! Thanks, Oscar!

LeoGrin (approved):
Looks great to me!

src/tabpfn/parallel_execute.py
oscarkey: done!

Issue #416, opened 2025-07-28
Tabpfn Regressor not loading on cpu even after setting device to cpu


thebinarycode:
### Describe the bug

using both Classifier and Regressor model for a usecase
tabpfn: 2.1.0
torch: 2.6.0
python: 3.11

After fitting on training data, the model weights are saved via save_fitted_tabpfn_model.
The saved the weights are loaded using load_fitted_tabpfn_model

Classifier was loading on gpu even by passing device="cpu" to load_fitted_tabpfn_model. 
it loads on cpu by passing arg device=torch.device("cpu")
but Regressor seems to ignore the argument and attempts to load on cuda

File "python3.11/site-packages/tabpfn/model/loading.py", line 855, in load_fitted_tabpfn_model
    fitted_attrs = joblib.load(tmp / "fitted_attrs.joblib")
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/joblib/numpy_pickle.py", line 658, in load
    obj = _unpickle(fobj, filename, mmap_mode)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/joblib/numpy_pickle.py", line 577, in _unpickle
    obj = unpickler.load()
          ^^^^^^^^^^^^^^^^
  File "/python3.11/pickle.py", line 1213, in load
    dispatch[key[0]](self)
  File "python3.11/pickle.py", line 1590, in load_reduce
    stack[-1] = func(*args)
                ^^^^^^^^^^^
  File "python3.11/site-packages/torch/storage.py", line 533, in _load_from_bytes
    return torch.load(io.BytesIO(b), weights_only=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/torch/serialization.py", line 1495, in load
    return _legacy_load(
           ^^^^^^^^^^^^^
  File "python3.11/site-packages/torch/serialization.py", line 1754, in _legacy_load
    result = unpickler.load()
             ^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/torch/serialization.py", line 1682, in persistent_load
    obj = restore_location(obj, location)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/torch/serialization.py", line 693, in default_restore_location
[...]

PR #486, opened 2025-09-05
Resolved issue saving & loading the predictor (Regressor/Classifier) model on different devices (cuda / cpu)

MagnusBuehler:
## Motivation and Context
Issue [#416](https://github.com/PriorLabs/TabPFN/issues/416) reported a crash when a model saved on a CUDA device was later loaded on a CPU-only machine, due to tensors being serialized as CUDA tensors. This has been fixed by ensuring all tensor-like elements of the predictor are saved as CPU tensors, allowing safe loading on any device. The desired device can be specified (or automatically inferred) at loading time. This adds a negligible one-time overhead (~±10 ms, tested over 10 runs). Corresponding tests now cover all combinations of fitting, saving, and loading devices.
 Further I added uv.lock to the .gitignore file and renamed a function in "src/tabpfn/inference.py" from `save_state_expect_model_weights` to `save_state_except_model_weights`, as it makes more sense with what the function is doing. 

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
On a cuda-environment & locally. Also added corresponding pytests.
---

## Checklist

---

### DIFF
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -661,7 +661,7 @@ def save_fitted_tabpfn_model(estimator: BaseEstimator, path: Path | str) -> None
             for key, value in vars(estimator).items()
             if key.endswith("_") and key not in blacklist
         }
-        # move all tensors to "cpu" before loading, so if fitted & saved on cuda-device
+        # move all tensors to "cpu" before saving, so if fitted & saved on cuda-device
         # and loading on cpu-device does not throw
         # "RuntimeError: Attempting to deserialize object on a CUDA device..."
         fitted_attrs = {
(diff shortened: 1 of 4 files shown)


src/tabpfn/model_loading.py
```diff
@@ -661,10 +661,18 @@ def save_fitted_tabpfn_model(estimator: BaseEstimator, path: Path | str) -> None
             for key, value in vars(estimator).items()
             if key.endswith("_") and key not in blacklist
         }
+        # move all tensors to "cpu" before loading, so if fitted & saved on cuda-device
```
noahho:
```suggestion
        # move all tensors to "cpu" before saving, so if fitted & saved on cuda-device
```

noahho (approved):
LGTM, thanks for the work. Please only review the comment made.

PR #488, opened 2025-09-07
QKV calculation improvements in attention mechanism

benraha:
## Motivation and Context

This PR has two changes:
1. Replace costly einsum with simple matrix multiplication - this version is faster in both PyTorch and ONNX. IMO, we do miss some of the clarity, but since this happens in every forward pass, it is worth it.
2. Add a fast-path for the `broadcast_kv_across_heads` function - since the use case where we don't have to change anything happens 50% of the time, this simple fix has great performance benefits.

---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?

Locally.

---

## Checklist

---

### DIFF
--- a/src/tabpfn/architectures/base/attention/full_attention.py
+++ b/src/tabpfn/architectures/base/attention/full_attention.py
@@ -434,13 +434,10 @@ class MultiHeadAttention(Attention):
             batch_shape = x.shape[:-1]  # [..., seq_len]
             j, nhead, d_k, input_size = self._w_qkv.shape
 
-            # [..., seq_len, input_size] -> [batch_flat * seq_len, input_size]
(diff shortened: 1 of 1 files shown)


benraha:
@priorphil can you please review? :)

priorphil:
Thanks! Could you share some of the benchmarks (including hardware, datasets shapes, dtype, timings) you ran for each of these changes so I can get a feeling for the magnitude of the speedup? :)

benraha:
Sure! I wrap the model in ONNX, and I measure a 15% reduction in the number of nodes in the graph and ~10% faster inference.

I work on CPU inference, with a fit of 1000 rows, running inference on a few lines at a time.

PR #492, opened 2025-09-08
Forward cache_trainset_representation to load_model.

priorphil:
This also allows to load models that don't support caching yet and works for models that might assign this to a different attribute (previously we set to cache_trainset_representation=True when loading, just to overwrite the attribute immediately after).


LeoGrin (approved):
LGTM!

PR #496, opened 2025-09-09
Select multiple gpus for inference, if available. Don't use them yet.

oscarkey:
If the specified device is "auto" and multiple cuda gpus are available, then select all the gpus. This PR doesn't actually implement multi-gpu inference. Instead, all the inference engines just use the first.

Public API changes:
- `TabPFNClassifier/Regressor.device_` has been renamed to `.devices_`, and is now a tuple rather than a single device.
- The `device` argument of `TabPFNClassifier/Regressor.__init__()` is unchanged, as is the `.device` property. Multiple gpus will only be used when this is set to "auto". An alternative would be to update this to `devices`. I kept it like this as most users probably only want a single device, it avoids changing the api, and we don't actually support multi-device inference in this PR. But maybe it is confusing to have both `device` and `devices_`. What do you think?

Notes:
- `TabPFNRegressor.znorm_space_bardist_` is always placed on the first device, as we're not aiming to parallelise this portion of inference
- `infer_devices()` no longer accepts `device=None`: I couldn't find anywhere using this

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -69,7 +69,7 @@ class InferenceEngine(ABC):
         self,
         X: np.ndarray,
         *,
-        device: torch.device,
+        devices: Sequence[torch.device],
         autocast: bool,
     ) -> Iterator[tuple[torch.Tensor, EnsembleConfig]]:
         """Iterate over the outputs of the model.
@@ -78,7 +78,7 @@ class InferenceEngine(ABC):
 
         Args:
             X: The input data to make predictions on.
-            device: The device to run the model on.
+            devices: The devices to run the model on.
             autocast: Whether to use torch.autocast during inference.
         """
         ...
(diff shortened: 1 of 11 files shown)


src/tabpfn/inference.py
```diff
@@ -194,10 +194,13 @@ def iter_outputs(
         self,
         X: np.ndarray,
         *,
-        device: torch.device,
+        devices: Sequence[torch.device],
         autocast: bool,
         only_return_standard_out: bool = True,
     ) -> Iterator[tuple[torch.Tensor | dict, EnsembleConfig]]:
+        # This engine currently only supports one device, so just take the first.
+        device = devices[0]
```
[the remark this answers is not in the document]
oscarkey: I think the current simple approach is better.

oscarkey:
@noahho are you happy with these public api changes?

tests/test_utils.py
```diff
@@ -118,3 +120,77 @@ def test_infer_categorical_with_dict_raises_error():
...
+def test__infer_devices__auto__cuda_and_mps_not_available__selects_cpu(
+    mocker: MagicMock,
+) -> None:
+    mocker.patch("torch.cuda").is_available.return_value = False
+    mocker.patch("torch.backends.mps").is_available.return_value = False
+    assert infer_devices(devices="auto") == (torch.device("cpu"),)
+
+
+def test__infer_devices__auto__single_cuda_gpu_available__selects_it(
+    mocker: MagicMock,
```
noahho, line 134: Nice tests! Just flagging we do have GPUs available for testing, but I think this does it too :-)

oscarkey: I was wondering about this, but we'd only be able to test cpu and 1 gpu, so I decided it was better to test all the device combinations in a consistent way.

PR #497, opened 2025-09-09
Add note to softmax_temperature for the classifier

brendan-priorlabs:
Address possible confusion from the fact that we expose a temperature setting, but use simply `argmax` in `predict`. See https://discord.com/channels/1285598202732482621/1412644840667086939/1414888423965196288


LeoGrin (approved):
LGTM thanks! I also don't know what "This is only applied when predicting during a post-processing step." but fine to leave it.

brendan-priorlabs:
Thanks, Leo! I'll leave refining that latter point for later, but lmk if you think it deserves attention

PR #498, opened 2025-09-09
Add consistency checks for the low_memory and fit_with_cache modes.

oscarkey:
The existing consistency checks use the default fit_preprocessors mode. Duplicate the "tiny classifier" test to check all three modes. Ideally this test file would be simplified and refactored e.g. to use pytest parametrize(), but leave it as it is for now.

Creating more of these tests adds friction, but I think it is worth it to test all the fit modes. In particular, the existing, flakey tests for the fit modes were disabled in
https://github.com/PriorLabs/TabPFN/pull/481 and
https://github.com/PriorLabs/TabPFN/pull/421 .


LeoGrin (approved):
LGTM thanks! Can we also do it for regression while we're at it if it's quick?

oscarkey:
oh great idea. done!

PR #499, opened 2025-09-09
Set up methods for anonymous usage analytics

safaricd:
## Motivation and Context

This PR introduces fully anonymous usage analytics to help us understand how the library is being used and guide future improvements.

Telemetry is GDPR-compliant and designed with data minimization in mind — no personal data, code, inputs, or outputs are ever collected.

By default, analytics are enabled in an opt-in, transparent manner, and users can easily disable them at any time via the `TABPFN_DISABLE_TELEMETRY=1` environment variable.

## Public API Changes

-   [x] No Public API changes

---

## Checklist

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ authors = [
   { name = "Oscar Key" },
   { name = "Felix Birkel" },
   { name = "Philipp Jund" },
-  { name = "Felix Jabloski" },
   { name = "Brendan Roof" },
   { name = "Dominik Safaric" },
   { name = "Benjamin Jaeger" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
   # Once Python 3.10 is the minimum version, this can be removed.
   "eval-type-backport>=0.2.2",
   "joblib>=1.2.0",
+  "tabpfn-common-utils[telemetry-interactive]>=0.1.7"
 ]
 requires-python = ">=3.9"
 authors = [
@@ -35,12 +36,15 @@ authors = [
   { name = "Eddie Bergman" },
   # Prior Labs Contributors
   { name = "Leo Grinsztajn" },
-  { name = "Felix Jabloski"},
-  { name = "Klemens Flöge"},
-  { name = "Oscar Key"},
-  { name = "Felix Birkel"},
-  { name = "Philipp Jund"},
-
+  { name = "Felix Jabloski" },
+  { name = "Klemens Flöge" },
+  { name = "Oscar Key" },
+  { name = "Felix Birkel" },
+  { name = "Philipp Jund" },
+  { name = "Felix Jabloski" },
+  { name = "Brendan Roof" },
+  { name = "Dominik Safaric" },
+  { name = "Benjamin Jaeger" }
 ]
 
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
   # Once Python 3.10 is the minimum version, this can be removed.
   "eval-type-backport>=0.2.2",
   "joblib>=1.2.0",
-  "tabpfn-common-utils[telemetry-interactive]>=0.1.7"
+  "tabpfn-common-utils[telemetry-interactive]>=0.1.8"
 ]
 requires-python = ">=3.9"
 authors = [
--- a/scripts/generate_dependencies.py
+++ b/scripts/generate_dependencies.py
@@ -14,7 +14,9 @@ from pathlib import Path
 
 def parse_dependency_lines(content: str) -> list[str]:
     """Finds and cleans the dependency lines from the pyproject.toml content."""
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -35,12 +36,15 @@ authors = [
...
-
+  { name = "Felix Jabloski" },
+  { name = "Klemens Flöge" },
+  { name = "Oscar Key" },
+  { name = "Felix Birkel" },
+  { name = "Philipp Jund" },
+  { name = "Felix Jabloski" },
+  { name = "Brendan Roof" },
+  { name = "Dominik Safaric" },
+  { name = "Benjamin Jaeger" }
```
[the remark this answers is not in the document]
noahho: Please change :-)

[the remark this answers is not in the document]
safaricd: Thank you for the feedback, de-duplicated the list. 😊

noahho (changes requested):
When installing I get an error:
```
warning: The package `tabpfn-common-utils==0.1.7` does not have an extra named `telemetry-interactive`
```

scripts/generate_dependencies.py
```diff
@@ -62,20 +64,24 @@ def main() -> None:
     if args.mode == "maximum":
         for dep in deps:
             # Check for maximum version constraint
-            max_version_match = re.search(r'([^>=<\s]+).*?<\s*([^,\s"\']+)', dep)
+            # Updated regex to handle package extras like [telemetry-interactive]
```
oscarkey: I guess "Updated" only makes sense in the context of this PR. So we should either say "Be careful to handle package extras like [telemetry-interactive]" or nothing at all?

src/tabpfn/classifier.py
```diff
@@ -689,6 +696,7 @@ def _raw_predict(self, X: XType, *, return_logits: bool) -> torch.Tensor:
 
         return self.forward(X, use_inference_mode=True, return_logits=return_logits)
 
+    @track_model_call(model_method="predict", param_names=["X"])
```
oscarkey, line 699: As this calls `predict_proba().` we will end up with two logs. I guess this can be filtered out when looking at the data, but would it make analysis easier to create a `_predict_proba()` function and call this from both `predict()` and `predict_proba()`, thus avoiding the double log?

safaricd:
> When installing I get an error:
> 
> ```
> warning: The package `tabpfn-common-utils==0.1.7` does not have an extra named `telemetry-interactive`
> ```

@noahho fix has been applied and pushed, thanks for reporting. It was due to the `generate_dependencies.py` script we use for parsing and generating `requirements.txt` from `pyproject.toml`.

src/tabpfn/classifier.py
safaricd: Very good catch, thanks @oscarkey 🤗 Wrapped the function into `_predict_proba`, please have a look.

scripts/generate_dependencies.py
safaricd: Agree with the comment - changed the wording.

src/tabpfn/classifier.py
```diff
@@ -722,9 +731,24 @@ def predict_logits(self, X: XType) -> np.ndarray:
         return logits_tensor.float().detach().cpu().numpy()
 
     @config_context(transform_output="default")  # type: ignore
+    @track_model_call(model_method="predict", param_names=["X"])
     def predict_proba(self, X: XType) -> np.ndarray:
         """Predict the probabilities of the classes for the provided input samples.
 
+        This is a wrapper around the `_predict_proba` method.
+
```
oscarkey, line 739:
```suggestion
```
I suggest we don't need this as it doesn't add anything to the public documentation, and can very quickly be observed from reading the method implementation.

oscarkey (approved):
LGTM!

src/tabpfn/classifier.py
```diff
@@ -698,14 +706,17 @@ def predict(self, X: XType) -> np.ndarray:
         Returns:
             The predicted class labels as a NumPy array.
         """
-        proba = self.predict_proba(X)
+        with config_context(transform_output="default"):
```
brendan-priorlabs: I haven't looked into this deeply, but it seems like we should decorate with `config_context` like we do with the other methods. Currently it's only being used as a context manager in tests. If that's wrong for some reason, could you please add a comment to the code explaining the special case?

brendan-priorlabs (approved):
LGTM. Thanks, Dominik

src/tabpfn/classifier.py
safaricd: @brendan-priorlabs It's more of a style question - double-decorating both `predict_proba` and `_predict_proba` (called by `predict_proba`) is redundant but harmless. This would essentially mean that sklearn will push another identical config context, which is not necessary - e.g. user calling `predict_proba`.

brendan-priorlabs: Single decorate `_predict_proba`?

safaricd: @brendan-priorlabs that works too. Just pushed to change.

noahho (approved):

PR #503, opened 2025-09-11
refactor preprocessors

bejaeger:
## Motivation and Context

All things related to preprocessing lived in one large file which made it hard to navigate and polluted the namespace. With this change, things are much cleaner. Each preprocessing step gets a separate file and everything is moved to a top level folder called `preprocessors`. This makes it easier to add different processors.

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

tests, full forward loop.

---

## Checklist

---


noahho (commented):
Nice work! I'm suprised the Quantile change didn’t affect predictions at all - is that right (I don't see you had to update consistency files)? 
Could you add an entry to the changelog file please: once refactoring, then the speed improvement but (if appropriate) flag it as potential change to behavior at e.g. large X size.

bejaeger:
Thanks for the quick review, added the inconsistency file (only miniscule changes) and an entry to the changelog.

noahho (approved):
LGTM!

PR #505, opened 2025-09-12
Increase versioning from 2.1.4 to 2.2.0

safaricd:
## Motivation and Context

- Increase versioning for `tabpfn` from `2.1.4` to `2.2.0`

## Public API Changes

-   [x] No Public API changes

---

## Checklist

---


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.2.0] - 2025-09-12
+
+### Added
+
+### Changed
+- @safaricd Increase versioning from `2.1.4` to `2.2.0`
+
 ## [2.1.4] - 2025-09-11
```
noahho:
```suggestion
## [2.1.4] - 2025-09-11 - yanked
```

README.md
```diff
@@ -336,7 +336,7 @@ The data is used exclusively to help us understand how the library is being used
 We only collect high-level, non-identifying information such as:
 - Package version
 - Operating system & Python version
-- General feature usage
+- How often fit and inference are called, including simple metadata like the dimensionality of the input (:warning: never the data itself)
```
noahho:
```suggestion
- How often fit and inference are called, including simple metadata like the dimensionality of the input and which tabpfn functionalities were used (:warning: never the data itself)
```

noahho (commented):
Thanks! Please take a look at the suggestions before merging

README.md
safaricd: @noahho thanks for the suggestion.

We already describe `how often fit and inference are called`, so `which tabpfn functionalities were used` feels a bit redundant because we only keep track of fit and inference. WDYT?

noahho: I'd keep it, we measure the regression, classification, extensions etc

noahho: We could also make this more explicit

safaricd: @noahho I see, and good point. I made the wording even more explicit:

> - How often fit and inference are called, including simple metadata like the dimensionality of the input and the type of task (e.g., classification vs. regression) (:warning: never the data itself)

Please let me know if there is anything else to change. Btw, these are the only metadata we collect - dimensionality and type of a task.

noahho (approved):
Thanks!

CHANGELOG.md
safaricd: This was changed.

PR #516, opened 2025-09-16
Add global seed fixture for reproducible tests

MagnusBuehler:
Defines a session-scoped fixture in conftest.py that sets seeds for torch, numpy, and random to ensure deterministic test behavior across sessions.

## Motivation and Context
Multiple tests rely on randomly sampled data using random, numpy, and torch. Without a fixed seed, this can lead to non-deterministic and flaky test results. Adding a global seed fixture improves consistent and reproducible test behavior across runs.

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
On a local machine the randomly sampled data across test runs are consistent.
---

## Checklist

---

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,9 +10,8 @@ from __future__ import annotations
 import os
 import random
 
-import pytest
-
 import numpy as np
+import pytest
 import torch
 
 


noahho (approved):
I think this is a very positive change, hope I'm not overlooking anything here. Thanks a lot!

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
+# conftest.py
```
oscarkey:
```suggestion
```

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+
+import random
+
+import numpy as np
+import pytest
+import torch
+
+
+@pytest.fixture(autouse=True, scope="session")
+def set_global_seed():
```
oscarkey:
```suggestion
def set_global_seed() -> None:
```

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+from __future__ import annotations
+
+import random
+
+import numpy as np
+import pytest
+import torch
+
+
+@pytest.fixture(autouse=True, scope="session")
```
oscarkey: Should the scope instead be "function", so the seeds are set before every test? If I understand correctly, this will just set the seed once per test session, which could still result in unpredictable behaviour if the order of the tests is changed (e.g. due to adding/removing tests, or using pytest-xdist to run them in parallel).

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+def set_global_seed():
+    seed = 42
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
```
oscarkey:
```suggestion
```
I think torch.manual_seed() will set the seed for all devices, see https://discuss.pytorch.org/t/difference-between-torch-manual-seed-and-torch-cuda-manual-seed/13848/5, so we don't need the separate call. Otherwise, we would also need to add one for MPS.

tests/conftest.py
noahho: ohh yes taht's a very good and subtle point 🙌

tests/conftest.py
MagnusBuehler: agreed.

tests/conftest.py
MagnusBuehler: added suggested return type.

tests/conftest.py
MagnusBuehler: @oscarkey I removed `torch.cuda.manual_seed_all()`

tests/conftest.py
MagnusBuehler: @oscarkey Good catch. I'll set `scope="function"`, such that execution order etc does not matter. Ruff gives a warning, due to `"function"` being the default value, which I disabled ( noqa: PT003), as I think it makes sense to explicitly still set the scope to `"function"` for clarity.

oscarkey: makes sense to me! this rule seems a bit ott

oscarkey (approved):

PR #517, opened 2025-09-17
If device="auto" and multiple GPUs are present, only select the first.

oscarkey:
We're seeing prediction errors when multiple GPUs are used. Switch to only using multiple GPUs if explicitly enabled, while we debug.

Tested manually on a machine with 2 GPUs and a dataset exhibiting the issue:
- current main: inference does not work
- this PR with device=auto: inference works
- this PR with device=["cuda:0","cuda:1"]: inference does not work

Also, update the device docstring. Don't include multi-gpu inference for now, until we've fixed it.


noahho (approved):
LGTM barring geminis comment

PR #518, opened 2025-09-17
Try enabling MPS devices on the CI.

oscarkey:


oscarkey:
swapping reviewer because of ooo

PR #526, opened 2025-09-23
Re-enable onnx tests.

oscarkey:
Onnx now works on Python 3.13.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,10 +68,7 @@ jobs:
       - name: Install dependencies
         run: |
           uv pip install --system --resolution ${{ matrix.dependency-set }} .
-          uv pip install --system pytest pytest-mock psutil
-          if [[ "${{ matrix.python-version }}" != "3.13" ]]; then
-            uv pip install --system onnx
-          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -72,10 +72,7 @@ jobs:
         run: |
           uv pip install --system --no-deps .
           uv pip install --system -r requirements.txt
-          uv pip install --system pytest pytest-mock psutil
-          if [[ "${{ matrix.python-version }}" != "3.13" ]]; then
-            uv pip install --system onnx
-          fi
+          uv pip install --system pytest pytest-mock psutil onnx
```
brendan-priorlabs, line 71: `onnx` sounds like a proper dependency as opposed to something test-only like, e.g., pytest-mock here. Any reason not to put it in `pyproject.toml`? You could make an optional section `test_only` or something if the full set of [`dev` dependencies i](https://github.com/PriorLabs/TabPFN/blob/21344388579ae628ecf7120475adc7b9a0742aa2/pyproject.toml#L74)s too much

oscarkey: Yes that makes sense, I'll make a follow-up PR with a minimum set of test dependencies.

PR #534, opened 2025-10-02
Bump max Sklearn version to < 1.8

klemens-floege:


LeoGrin (changes requested):
I think we could do <1.8. We kind of moved from stricter restrictions to broader ones at some point because the former was quite annoying to manage

klemens-floege:
Fair enough, updated the .toml

LeoGrin (approved):

LeoGrin:
Thanks!

Issue #530, opened 2025-09-30
Bug: Column Reordering by ColumnTransformer Causes Incorrect Missing Value Handling and Feature Attribution


jinyueyuan:
### Describe the bug

Hi team,
There's a bug in the TabPFN preprocessing pipeline that affects both missing value handling and feature interpretation. The issue stems from ColumnTransformer's automatic column reordering behavior, which breaks the assumption that original column indices remain valid after transformation. 

**Problem Overview**
The OrdinalEncoder.transform will reorders columns by placing categorical columns first (i.e., dtype_include=["category",   "string"] in the code), followed by numerical columns. However, several parts of the codebase still use original column indices, leading to incorrect data processing. 

**Affected Components**
1. Missing Value Handling (utils.py:process_text_na_dataframe:632-638)
2. SHAP Feature Attribution: (classifier.py:_raw_predict)
3. Potential model performance impact when evaluating TabPFN model performance on dataset with lots of categorical features or missing values.
  

**Code Detail Information**
E.g., below is the source code for process_text_na_dataframe, "X_encoded" returned by OrdinalEncoder(ord_encoder) is reordered with categorical columns first.  "placeholder_mask" is calculated based on original order (X), but applied to new order (X_encoded) for missing value manipulation.

```
def process_text_na_dataframe(
    X: pd.DataFrame,
    placeholder: str = NA_PLACEHOLDER,
    ord_encoder: ColumnTransformer | None = None,
    *,
    fit_encoder: bool = False,
) -> np.ndarray:
    """Convert `X` to float64, replacing NA with NaN in string cells.

    If `ord_encoder` is not None, then it will be used to encode `X` before the
    conversion to float64.

    Note that this function sometimes mutates its input.
    """
    string_cols = X.select_dtypes(include=["string", "object"]).columns
    if len(string_cols) > 0:
        X[string_cols] = X[string_cols].fillna(placeholder)

    if fit_encoder and ord_encoder is not None:
        X_encoded = ord_encoder.fit_transform(X)
    elif ord_encoder is not None:
        X_encoded = ord_encoder.transform(X)
    else:
        X_encoded = X

    string_cols_ix = [X.columns.get_loc(col) for col in string_cols]
    placeholder_mask = X[string_cols] == placeholder
    X_encoded[:, string_cols_ix] = np.where(
        placeholder_mask,
        np.nan,
        X_encoded[:, string_cols_ix],
    )
    return X_encoded.astype(np.float64)
```

[...]

PR #539, opened 2025-10-07
Fix column reordering in process_text_na_dataframe  with nan and missing values due to Columntransformer

MagnusBuehler:
## Motivation and Context

- fixing bug mentioned in _https://github.com/PriorLabs/TabPFN/issues/530_
- columntransformer does not keep the feature order
- therefore, this change computes a mapping from the new ordering (not what we keep) to the original column order
- added tests to check different use-cases with nan and missing values

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

Locally. Implemented a test, which runs different combinations of nan and missing values. 

## Checklist

---

### DIFF
--- a/src/tabpfn/utils.py
+++ b/src/tabpfn/utils.py
@@ -627,11 +627,29 @@ def process_text_na_dataframe(
     else:
         X_encoded = X
 
+    if ord_encoder is not None:
+        # The ColumnTransformer will reorder the columns as following:
+        # - categorical columns (in the order they appear in the dataframe)
+        # - numerical columns (in the order they appear in the dataframe)
+        # We need to revert this to the original order.
+        # assumes the column transformer has one OrdinalEncoder, which is
+        # what we get when creating it using get_ordinal_encoder()
+        cat_cols = next(
+            columns
+            for _, transformer, columns in ord_encoder.transformers_
+            if isinstance(transformer, OrdinalEncoder)
+        )
+        num_cols = [c for c in X.columns if c not in cat_cols]
+        col_to_pos = {c: i for i, c in enumerate(cat_cols + num_cols)}
+        X_encoded = X_encoded[:, [col_to_pos[c] for c in X.columns]]
+
     string_cols_ix = [X.columns.get_loc(col) for col in string_cols]
     placeholder_mask = X[string_cols] == placeholder
     X_encoded[:, string_cols_ix] = np.where(
         placeholder_mask,
         np.nan,
+        # iloc because get_loc returns the position of the column
+        # in-case X_encoded has string column names
         X_encoded[:, string_cols_ix],
     )
     return X_encoded.astype(np.float64)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -6,10 +6,22 @@ import os
 from unittest.mock import MagicMock
 
 import numpy as np
+import pandas as pd
 import pytest
 import torch
+from sklearn.preprocessing import LabelEncoder
 
-from tabpfn.utils import infer_categorical_features, infer_devices
+from tabpfn import TabPFNClassifier
+from tabpfn.config import ModelInterfaceConfig
+from tabpfn.constants import NA_PLACEHOLDER
+from tabpfn.utils import (
+    fix_dtypes,
+    get_ordinal_encoder,
+    infer_categorical_features,
+    infer_devices,
+    process_text_na_dataframe,
+    validate_Xy_fit,
+)
 
 
 def test_internal_windows_total_memory():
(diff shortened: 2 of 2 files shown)


src/tabpfn/utils.py
```diff
@@ -627,6 +627,22 @@ def process_text_na_dataframe(
     else:
         X_encoded = X
 
+    if ord_encoder is not None:
+        # The ColumnTransformer will reorder the columns as following:
+        # - categorical columns (in the order they appear in the dataframe)
+        # - numerical columns (in the order they appear in the dataframe)
+        # We need to revert this to the original order.
+        # assumes the column transformer has one OrdinalEncoder, which is
+        # what we get when creating it using get_ordinal_encoder()
```
bejaeger: I would prefer an assert rather than a comment?

src/tabpfn/utils.py
```diff
@@ -627,6 +627,22 @@ def process_text_na_dataframe(
...
+        cat_cols = next(
+            columns
+            for _, transformer, columns in ord_encoder.transformers_
+            if isinstance(transformer, OrdinalEncoder)
+        )
+        num_cols = [c for c in X.columns if c not in cat_cols]
+        col_to_pos = {c: i for i, c in enumerate(cat_cols + num_cols)}
+        X_encoded = X_encoded[:, [col_to_pos[c] for c in X.columns]]
+
```
bejaeger: Hm...this looks a bit hacky and is not easy to generalize to other places.
I think it would be great to have something like a `OrderPreservingColumnTransformer` class that handles that for us and which we instantiate in `get_ordinal_encoder()`. Then we can also switch to this class in other places where we might not treat this properly. What do you think?

tests/test_utils.py
```diff
@@ -192,3 +204,131 @@ def test__infer_devices__device_selected_twice__raises() -> None:
...
+    assert X_out.shape[1] == X.shape[1]
+
+    # Ensure no extra features
+    assert X_out.shape[1] == len(X.columns)
+
+    # Column order: verify string/object columns
+    string_cols = X.select_dtypes(include=["object", "string"]).columns
+    for col in string_cols:
+        col_idx = X.columns.get_loc(col)
+        mask = X[col] == NA_PLACEHOLDER
```
bejaeger: I'm a fan of explicit testing because it is also a great way of documenting. 
Could you add another array of `expected_results` similar to `test_cases` with exactly the `X_out` we expect?
You can maybe remove one or two examples to test because some seem to test similar scenarios.

tests/test_utils.py
```diff
@@ -192,3 +204,131 @@ def test__infer_devices__device_selected_twice__raises() -> None:
...
+    # Infer categorical features
+    cls.inferred_categorical_indices_ = infer_categorical_features(
+        X=X,
+        provided=getattr(cls, "categorical_features_indices", None),
+        min_samples_for_inference=ModelInterfaceConfig.MIN_NUMBER_SAMPLES_FOR_CATEGORICAL_INFERENCE,
+        max_unique_for_category=ModelInterfaceConfig.MAX_UNIQUE_FOR_CATEGORICAL_FEATURES,
+        min_unique_for_numerical=ModelInterfaceConfig.MIN_UNIQUE_FOR_NUMERICAL_FEATURES,
+    )
+
+    # Fix dtypes
```
bejaeger: Could you remove all the comments that are a bit redundant because they just repeat what's written in actual code?

src/tabpfn/preprocessors/preprocessing_helpers.py
```diff
@@ -166,6 +180,79 @@ def transform(self, X: np.ndarray) -> TransformResult:
...
+
+                # map original columns to indices in the transformed array
+                transformed_columns = col_subset + [
+                    c for c in original_columns if c not in col_subset
+                ]
+                indices = [transformed_columns.index(c) for c in original_columns]
+                X = X.iloc[:, indices] if isinstance(X, pd.DataFrame) else X[:, indices]
+                # Once we restored the original columns, for one coder we can break out
+                # of the loop
+                break
```
bejaeger: What if there are 3 transformers, each of them processing only a subset of columns?
I don't think we have that case anywhere but I think we should change this class to either support the very general case or make sure we don't use it when there are more than 2 transformers (raise ValueError during init).

bejaeger (commented):
Thanks @MagnusBuehler , looks great!
I added another comment which would be good to address.

Then another thing, could you add tests specifically for the new Transformer? Would make it easy to ensure it works for different scenarios.

tests/test_utils.py
```diff
@@ -192,3 +204,180 @@ def test__infer_devices__device_selected_twice__raises() -> None:
...
+        X,
+        placeholder=NA_PLACEHOLDER,
+        ord_encoder=ord_encoder,
+        fit_encoder=True,
+    )
+
+    assert X_out.shape[0] == ground_truth.shape[0]
+    assert X_out.shape[1] == ground_truth.shape[1]
+
+    for col_name in X.columns:
```
bejaeger, line 348: Why don't compare the full `X_out` with the ground truth? am I missing something?

src/tabpfn/utils.py
```diff
@@ -408,7 +415,7 @@ def get_ordinal_encoder(
     # using a regex on the type of the column, and using `object`, `"object"` and
     # `np.object` will not pick up strings.
     to_convert = ["category", "string"]
-    return ColumnTransformer(
+    return OrderPreservingColumnTransformer(
```
noahho: Can we add a larger comment on why we need the order preserving transformer here? Also a note that this might be not necessary if we switch to fully detecting the categoricals based on data type

src/tabpfn/preprocessors/preprocessing_helpers.py
MagnusBuehler: @bejaeger You are right, this should be handled. One issue is that we can only maintain the original column order if all transformers inside the `ColumnTransformer` are subtypes of `OneToOneFeatureMixin` (i.e., have a one-to-one correspondence between input and output features) and all inner transformers process disjoint subsets of features. If two inner transformers (for example, `StandardScaler` and `PowerTransformer`) are applied to the same feature, the `ColumnTransformer` will not pipe them sequentially. Instead, it will create one output feature with `StandardScaler` applied and another output feature with `PowerTransformer` applied. This increases the total number of output features by one, which would require updating `inferred_categorical_variables` accordingly. The same applies to other non-`OneToOneFeatureMixin` transformers, such as `OneHotEncoder`. 
I think as a quick fix raising a `ValueError` during init would work for now.

tests/test_utils.py
MagnusBuehler: I do not think we can compare the full `X_out` with the ground truth if we use the `get_ordinal_encoder` function, which does not pass a specific `categories` parameter to control for the exact encoding.
 For example,  for a column with entries `["cold", "warm", "cold", "hot"]`, the encoder could map these to `[0, 1, 0, 2]`, `[1, 0, 1, 2]`, or other permutations depending on the internal ordering of categories. 
There is no guarantee which numeric label corresponds to which category. Meanwhile, the ground truth represents one specific manual mapping, such as `[2, 1, 2, 0]`. As a result, even if the encoding is technically correct, it may not match the ground truth exactly.

bejaeger: Hm...okay, but isn't the `OrdinalEncoder` deterministic in mapping objects to ints? Meaning you can setup a test case for it?

bejaeger (approved):
Cool! Thanks for working on this and implementing all the changes.
LGTM let's get this in :)

LeoGrin:
@MagnusBuehler is this ready to merge?

src/tabpfn/preprocessors/preprocessing_helpers.py
```diff
@@ -166,6 +185,121 @@ def transform(self, X: np.ndarray) -> TransformResult:
...
+        **kwargs : additional keyword arguments
+            Passed to sklearn.compose.ColumnTransformer.
+        """
+        super().__init__(transformers=transformers, **kwargs)
+
+        # Check if there is a single transformer, of subtype OneToOneFeatureMixin
+        assert all(
+            isinstance(t, OneToOneFeatureMixin)
+            for name, t, _ in transformers
+            if name != "remaineder"
```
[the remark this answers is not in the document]
LeoGrin: +1

src/tabpfn/utils.py
MagnusBuehler: done

src/tabpfn/preprocessors/preprocessing_helpers.py
```diff
@@ -7,10 +7,29 @@
...
+
+import pandas as pd
+from sklearn.base import (
+    BaseEstimator,
+    check_is_fitted,
+)
+from sklearn.compose import ColumnTransformer, make_column_selector
+from sklearn.preprocessing import FunctionTransformer, OrdinalEncoder
+
+if TYPE_CHECKING:
```
noahho: Can this be merged with above TYPE_CHECKING?

noahho: the import section just be cleaned up a bit (reorder)

noahho (approved):
LGTM, @MagnusBuehler please merge if ready

src/tabpfn/preprocessors/preprocessing_helpers.py
MagnusBuehler: good catch! I organized the imports.

PR #543, opened 2025-10-10
Fix potential race condition in InferenceEngineOnDemand.

oscarkey:
If the number of workers used in `fit_preprocessings()` is >1, then the outputs of `fit_preprocessings()` could end up in a different order to the input configs. This would mean the configs could end up being paired with the wrong outputs when returned from `iter_outputs()`.

I don't think this bug could happen in practice because the `n_workers` parameter is not connected, and we always use one worker: https://github.com/PriorLabs/TabPFN/blob/7373d9dd3b13397ebfdba1262baa3ebca5952e61/src/tabpfn/preprocessing.py#L739
It's reasonably likely we might want to use multiple workers in the near future, so I don't remove n_workers.

I updated the tests to catch this race condition, and verified that the test works by disabling the fix, setting the number of workers to >1, and observing that the test fails.

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -108,22 +108,9 @@ def test__cache_preprocessing__result_equal_in_serial_and_in_parallel() -> None:
         )
     )
 
-    def find_seq_output(config: EnsembleConfig) -> Tensor | dict:
-        """Find the sequential output corresponding to the given config.
-
-        The configs are not hashable, so we have to resort to this search method.
-        """
-        for output, trial_config in outputs_sequential:
-            if trial_config == config:
-                return output
-
-        return pytest.fail(
-            f"Parallel config was not found in sequential configs: {config}"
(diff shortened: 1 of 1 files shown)


tests/test_inference.py
```diff
@@ -147,11 +164,54 @@ def test__on_demand__result_equal_in_serial_and_in_parallel() -> None:
             X_test, devices=[torch.device("cpu"), torch.device("cpu")], autocast=False
         )
     )
+
+    def find_seq_output(config: EnsembleConfig) -> Tensor | dict:
```
brendan-priorlabs: Would be nice to deduplicate here. Could change to `find_seq_output(config: EnsembleConfig, outputs_sequential: List)` and move outside the test

oscarkey: ah nice idea, done!

PR #549, opened 2025-10-14
Update Ruff to 0.14.

oscarkey:
- Run `ruff format` and `ruff check --fix`
- New rule PLC0415 requires imports to be at the top level: move some imports and add noqa to others.
- Move ignores of two entire files (which are copied without changes from other repositories) to the pyproject.toml

### DIFF
--- a/tests/test_save_load_fitted_model.py
+++ b/tests/test_save_load_fitted_model.py
@@ -144,7 +144,7 @@ def test_save_load_happy_path(
 
 
 # --- Error Handling Tests ---
-def test_saving_unfitted_model_raises_error(regression_data, tmp_path):
+def test_saving_unfitted_model_raises_error(tmp_path: Path) -> None:
     """Tests that saving an unfitted model raises a RuntimeError."""
     model = TabPFNRegressor()
     with pytest.raises(RuntimeError, match="Estimator must be fitted before saving"):


src/tabpfn/architectures/base/bar_distribution.py
```diff
@@ -802,7 +803,7 @@ def get_bucket_limits(
             assert full_range[1] >= ys.max()
             full_range = torch.tensor(full_range)  # type: ignore
 
-        ys_sorted, ys_order = ys.sort(0)  # type: ignore
+        ys_sorted, _ys_order = ys.sort(0)  # type: ignore
```
LeoGrin:
```suggestion
        ys_sorted, _ = ys.sort(0)  # type: ignore
```

tests/test_save_load_fitted_model.py
```diff
@@ -146,7 +146,7 @@ def test_save_load_happy_path(
 # --- Error Handling Tests ---
 def test_saving_unfitted_model_raises_error(regression_data, tmp_path):
     """Tests that saving an unfitted model raises a RuntimeError."""
-    X, y = regression_data
+    _X, _y = regression_data
```
LeoGrin:
```suggestion
```

PR #553, opened 2025-10-16
Add KV cache example for fast predictions

noahho:
Implement KV cache example


examples/kv_cache_fast_prediction.py
```diff
@@ -0,0 +1,54 @@
...
+    n_features=20,
+    random_state=42,
+    n_classes=2
+)
+
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.05, random_state=42, stratify=y
+)
+
+def bench(clf, name: str):
```
oscarkey:
```suggestion
def bench(clf: TabPFNClassifier, name: str):
```

PR #554, opened 2025-10-16
Explicitly disable Dynamo onnx exports to fix CI.

oscarkey:
torch 2.9 came out and changed dynamo to be the default onnx export mode (see https://docs.pytorch.org/docs/stable/onnx.html#functions). We don't work with the dynamo export, and this requires the onnxscript dependency, which we don't have, so the CI broke. Instead, set `dynamo=False` so it uses the torchscript export, which does work.

onnx export is not supported for users (we just have the tests to avoid our wip support regressing), so this was just a ci problem.


noahho:
@oscarkey Failing tests unfortunately!

oscarkey:
@noahho turns out the problem was they switched to the dynamo export by default (which then required onnxscript). So now I just force the old export method.

PR #557, opened 2025-10-17
Allow to run Classifier and Regressor with multiple models

bejaeger:
## Motivation and Context

This introduces the option to run `TabPFNClassifier` and `TabPFNRegressor` with multiple models.

---

## Public API Changes

-   [ ] No Public API changes

There are no breaking changes, but the `model_path` argument can also be a list of paths now.

---

## Checklist

---


examples/finetune_regressor.py
```diff
@@ -158,7 +158,7 @@ def main() -> None:
 
     # Optimizer must be created AFTER get_preprocessed_datasets, which initializes the model
     optimizer = Adam(
-        regressor.model_.parameters(), lr=config["finetuning"]["learning_rate"]
+        regressor.models_[0].parameters(), lr=config["finetuning"]["learning_rate"]
```
oscarkey: Is it better here to add all of the models' parameters to the optimizer? In case we change the default config of the classifier/regressor to use multiple models. It is a bit more confusing though, hmmm

src/tabpfn/base.py
```diff
@@ -121,54 +115,76 @@ def initialize_tabpfn_model(
...
-        download = True
+        return [model_path.model], [model_path.config], model_path.norm_criterion
+
+    if isinstance(model_path, ClassifierModelSpecs) and which == "classifier":
+        return [model_path.model], [model_path.config], None
+
+    if (
+        isinstance(model_path, list)
+        and len(model_path) > 0
+        and isinstance(model_path[0], RegressorModelSpecs)
```
oscarkey: May as well check that every element is a RegressorModelSpecs (or ClassifierModelSpecs)?

src/tabpfn/inference.py
```diff
@@ -202,39 +201,46 @@ def iter_outputs(
...
+        itr = fit_preprocessing(
+            configs=self.ensemble_configs,
+            X_train=self.X_train,
+            y_train=self.y_train,
+            random_state=rng,
+            cat_ix=self.cat_ix,
+            n_workers=self.n_workers,
+            parallel_mode="as-ready",
         )
+        ensemble_configs, preprocessors, X_trains, y_trains, cat_ixs = list(zip(*itr))
```
oscarkey: Can we avoid doing the conversion to the list? Because this means the preprocessing will no longer be overlapped with the model execution.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Allow to run classifier and regressor with multiple models [#557](https://github.com/PriorLabs/TabPFN/pull/557).
```
oscarkey: Worth highlighting that `.model_` is now `.models_`? As it's part of the public api

oscarkey (changes requested):
lgtm!
Have you done a search on github to check if any of our other packages are using the `.model_` property? Unfortunately, atm the demo notebook installs main of both tabpfn and tabpfn-extensions, so we might break that.

examples/finetune_regressor.py
bejaeger: I made the scripts more conservative and just raise an error when the regressor uses multiple models.

src/tabpfn/base.py
bejaeger: Done!

src/tabpfn/inference.py
bejaeger: Yes, very good point, I missed that! It was a bit more tricky and required some code but it should be good now.

CHANGELOG.md
bejaeger: As discussed, `.model_` works when only a single model is used, so I don't mention it explicitely here

bejaeger:
Thanks for the review @oscarkey , it's ready for another round.

examples/finetune_regressor.py
oscarkey: oh nice, that's a good idea

src/tabpfn/classifier.py
```diff
@@ -406,13 +416,40 @@ class in Fine-Tuning. The fit_from_preprocessed() function sets this
...
+        if not hasattr(self, "models_"):
+            raise ValueError(
+                "The model has not been initialized yet. Please initialize the model "
+                "before using the `model_` property."
+            )
+        if len(self.models_) > 1:
+            raise ValueError(
+                "The `model_` property is not supported when multiple models are used. "
+                "Use `models_` instead."
+            )
```
oscarkey: query usage search

oscarkey (approved):

PR #561, opened 2025-10-22
Check for incompatible licenses

brendan-priorlabs:
ref ENG-229

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,8 @@ jobs:
         run: |
           licensecheck \
             --requirements-paths pyproject.toml \
-            --only-licenses APACHE MIT BSD ISC PYTHON \
+            --only-licenses APACHE MIT BSD ISC PYTHON UNLICENSE \
+            --ignore-packages certifi "tabpfn*" "nvidia*" \
             --show-only-failing \
             -0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,25 @@ jobs:
         run: uv pip install --system --resolution ${{ matrix.dependency-set }} ".[ci]"
         shell: bash
 
+      - name: Find who depends on pyobjc-framework-metal (mac only)
+        run: |
+          uv pip install --system pipdeptree
+          pipdeptree -r -p nvidia-cublas-cu12
+          pipdeptree -r -p nvidia-cuda-cupti-cu12
+          pipdeptree -r -p nvidia-cuda-nvrtc-cu12
+          pipdeptree -r -p nvidia-cuda-runtime-cu12
+          pipdeptree -r -p nvidia-cudnn-cu12
+          pipdeptree -r -p nvidia-cufft-cu12
+          pipdeptree -r -p nvidia-cufile-cu12
+          pipdeptree -r -p nvidia-curand-cu12
+          pipdeptree -r -p nvidia-cusolver-cu12
+          pipdeptree -r -p nvidia-cusparse-cu12
+          pipdeptree -r -p nvidia-cusparselt-cu12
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -69,6 +69,21 @@ jobs:
         run: uv pip install --system --resolution ${{ matrix.dependency-set }} ".[ci]"
         shell: bash
 
+      # Ignore notes:
+      # einops: Appears to be MIT. https://github.com/arogozhnikov/einops/blob/361b11e87da94ead4bd09de636c5dbed73e0e3e0/LICENSE
+      # certifi: MPL. https://github.com/certifi/python-certifi/blob/fb14ac49a976b1695d84b1ac1307276a20b3aac9/LICENSE
+      # tabpfn*: Our packages
+      # TODO: Add note on nvidia*.
```
brendan-priorlabs: @simo-prior, did we ever put together that central list of licenses/packages? Ideally I'd have something a bit more concrete for certifi and nvidia.

oscarkey (approved):
lgtm aside form nvidia comment!

.github/workflows/ci.yml
noahho: Is it possible to add MPL to the allowed licenses instead of ignoring above? Gemini suggests the below approach for it:

```
Step 1: Update pyproject.toml
Add a [tool.licensecheck] section to your pyproject.toml file. This will define all your rules, including the specific exception for certifi.

Ini, TOML

[tool.pytest.ini_options]
# ... (your existing pytest config) ...

[tool.licensecheck]
# This is your main list of allowed licenses
only_licenses = [
    "APACHE",
    "MIT",
    "BSD",
    "ISC",
    "PYTHON",
    "UNLICENSE",
]

# These packages are still ignored completely
ignore_packages = [
    "einops",
    "nvidia*",
    "tabpfn*",
]

# This is the new rule:
# Allow 'certifi' ONLY if its license is 'MPL' (or 'MPL-2.0')
allow_packages = [
    "certifi (MPL)",
]
Note: The license string must be exact. Run licensecheck locally once. If it reports the license as "MPL-2.0", you should use "certifi (MPL-2.0)" instead.

Step 2: Simplify .github/workflows/ci.yml
Now that all the rules are in pyproject.toml, you can make your CI step much simpler. licensecheck will automatically find and use that configuration.

YAML

      - name: 'Check for forbidden licenses'
        shell: bash
        run: |
          licensecheck \
            --requirements-paths pyproject.toml \
            --show-only-failing \
            -0
This change achieves your goal:

certifi is now explicitly allowed, but only with the MPL license.

MPL is not allowed for any other package.

nvidia* and others are still ignored as you intended.

Your CI file is cleaner.
```

.github/workflows/ci.yml
```diff
@@ -111,6 +145,40 @@ jobs:
         run: uv pip install --system --resolution highest ".[ci]"
         shell: bash
 
+      - name: Find who depends on pyobjc-framework-metal (mac only)
```
noahho: Why are we doing this @simo-prior ? Could we add a note?

brendan-priorlabs: This was me making sure I understood the dependency chain. Will be removed in the final PR.

brendan-priorlabs: Fixed

.github/workflows/ci.yml
brendan-priorlabs: Nice, thanks! This worked well except `allow_packages` appeared to be a hallucination, so I just had to ignore `certifi` like the others

brendan-priorlabs:
@noahho, this should be ready for another pass

noahho (approved):
LGTM, thanks for giving this another pass

PR #562, opened 2025-10-22
Fix build error on Python 3.9

brendan-priorlabs:
Our mac builds with Python 3.9 were all failing because an indirect dep stopped publishing packages for 3.9. See https://github.com/PriorLabs/TabPFN/actions/runs/18689219131/job/53367028472?pr=539

This fixes it by pinning away for the old builds.


bejaeger (approved):
Thanks!

PR #576, opened 2025-10-29
Load the inference config from the checkpoint, if present.

oscarkey:
- If the config is not present, guess the model version and load an appropriate default instead. This will be updated to include the v2.5 default.
- Use Pydantic to parse the InferenceConfig. There is a slight risk that parsing might fail at runtime for something it didn't before, but I couldn't see any problems in our repositories.
- I added some unit tests for the loading code, but overall this is covered by `test_..._interface.py`, which tries out all the different checkpoints.


src/tabpfn/classifier.py
```diff
@@ -1080,3 +1075,13 @@ def load_from_fit_state(
...
+
+
+def _get_preprocessor_configs(
+    inference_config: InferenceConfig,
+) -> list[PreprocessorConfig]:
+    if inference_config.PREPROCESS_TRANSFORMS == "v2_default":
+        return v2_classifier_preprocessor_configs()
+    if inference_config.PREPROCESS_TRANSFORMS is None:
+        return default_classifier_preprocessor_configs()
+    return inference_config.PREPROCESS_TRANSFORMS
```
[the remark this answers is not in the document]
oscarkey: it's not identical.

src/tabpfn/base.py
```diff
@@ -194,6 +219,15 @@ def initialize_tabpfn_model(
     )
 
 
+def _assert_inference_configs_equal(
+    model_specs: list[ClassifierModelSpecs] | list[RegressorModelSpecs],
+) -> None:
+    if not all(
+        spec.inference_config == model_specs[0].inference_config for spec in model_specs
+    ):
+        raise ValueError("All models must have the same inference config")
```
bejaeger, line 230: I think that's definitely the way to go for now. Just flagging that we might have cases in the future where we want different `inference_config`s for different models.

src/tabpfn/base.py
```diff
@@ -485,15 +519,15 @@ def initialize_model_variables_helper(
         dtype, and rng is a NumPy random Generator for use during inference.
     """
     static_seed, rng = infer_random_state(calling_instance.random_state)
-    models, configs, norm_or_bardist = initialize_tabpfn_model(
+    models, arch_configs, maybe_bardist, inference_config_ = initialize_tabpfn_model(
```
bejaeger: why trailing `_` in `inference_config_` already?

src/tabpfn/inference_config.py
```diff
@@ -160,40 +168,23 @@ class InferenceConfig:
...
-                for config in inference_config_.PREPROCESS_TRANSFORMS
-            ]
-
-        return inference_config_
+        if user_config is None:
+            return deepcopy(self)
+        if isinstance(user_config, InferenceConfig):
+            return deepcopy(user_config)
+        if isinstance(user_config, dict):
+            return dataclasses.replace(self, **user_config)
```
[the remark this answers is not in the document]
bejaeger: sounds like this is an issue?

src/tabpfn/inference_config.py
```diff
@@ -160,40 +168,23 @@ class InferenceConfig:
...
+        raise ValueError(f"Unknown {user_config=} passed to model.")
```
bejaeger:
```suggestion
        raise ValueError(f"{user_config=}\nUnknown user config provided, see config above.")
```
Would be a bit easier to read.

src/tabpfn/model_loading.py
```diff
@@ -600,14 +624,35 @@ def load_model(
...
+    """
+    if "inference_config" in checkpoint:
+        return InferenceConfig(**checkpoint["inference_config"])
+
+    inference_config = InferenceConfig()
+    # "architecture_name" was added to the checkpoint after the v2 release, so if this
+    # isn't present we assume it's v2.
+    if "architecture_name" not in checkpoint:
+        inference_config.PREPROCESS_TRANSFORMS = "v2_default"
+    return inference_config
```
bejaeger: Not quite sure where to comment this but shouldn't we be versioning the entire inference config?
The way it is set up right now would still allow for things to be inconsistent if we, for example, change a configuration other than `PREPROCESS_TRANSFORMS`?
Also, I would use an enum to specify the version rather than a string? Would be a bit more future proof.

src/tabpfn/base.py
```diff
@@ -52,15 +51,22 @@
...
-    def __init__(self, model: Architecture, config: ArchitectureConfig):
+    def __init__(
+        self,
+        model: Architecture,
+        arch_config: ArchitectureConfig,
+        inference_config: InferenceConfig,
+    ):
         self.model = model
-        self.config = config
+        self.architecture_config = arch_config
```
bejaeger: I like `architecture_config`. Right now, we're using both, `arch_config` and `architecture_config`. Should we harmonize? I'd have a slight preference for `architecture_config`.

src/tabpfn/inference_config.py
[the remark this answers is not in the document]
oscarkey: yes I think so. I'll fix and add a test.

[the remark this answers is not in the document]
oscarkey: interestingly I added the test and it seems to work as is...
see `test_inference_config.py`

src/tabpfn/base.py
oscarkey: ah yes I was wondering about that, but yes I think that should work with some more painful plumbing :'(

src/tabpfn/base.py
oscarkey: I went for architecture_config. I thought it would be so long that it messed up all the formating, but seems to be fine.

src/tabpfn/model_loading.py
oscarkey: I think this would have worked, assuming we never changed the defaults in InferenceConfig. I was only trying to provide backwards compatability for v2/v2.5, where we only changed the PREPROCESS_TRANSFORMS, and the config would be loaded from the checkpoint for all future models. But, yes sorry this was unnecessarily confusing and likely to go wrong. I've revamped it a bit, see what you think!

tests/reference_predictions/inconsistency_test_predictions.json
```diff
@@ -1,14 +1,14 @@
 [
   [
-    0.6066582798957825,
-    0.3933417499065399
+    0.6066581606864929,
+    0.3933418393135071
```
bejaeger: is this expected?

src/tabpfn/model_loading.py
```diff
@@ -600,14 +624,43 @@ def load_model(
...
+    is_v2 = "architecture_name" not in checkpoint
+    # TODO: Add check for v2.5
+
+    if isinstance(criterion, FullSupportBarDistribution):
+        if is_v2:
+            return InferenceConfig.get_v2_for_regression()
+        return InferenceConfig.get_default_for_regression()
+    if is_v2:
+        return InferenceConfig.get_v2_for_classification()
+    return InferenceConfig.get_default_for_classification()
```
bejaeger: Should we put that logic inside a function in `InferenceConfig` that decides based on: "is_regression", "model_version" (could use ModelVersion enum)?

bejaeger (approved):
Thanks for the changes! Made one more suggestion that you may or may not want to include in here, and then flagged the test prediction difference as a potential sign of a bug?

tests/reference_predictions/inconsistency_test_predictions.json
oscarkey: this is some weird thing that the consistency tests do to check that they're working: https://github.com/PriorLabs/TabPFN/blob/3432eb2fb2157a5c8407019b27ce29d54168d560/tests/test_consistency.py#L735
It changes every time you run the tests. We should change it to clean up the file, but for now I just reverted the change.

src/tabpfn/model_loading.py
oscarkey: ah nice idea, done!

src/tabpfn/base.py
oscarkey: I was trying to match the estimator property, but I think this makes no sense. So changed!

tests/reference_predictions/inconsistency_test_predictions.json
bejaeger: oh okay... thanks!

src/tabpfn/model_loading.py
bejaeger: Great! thanks

PR #578, opened 2025-10-30
Add unit tests for inference config.

oscarkey:
I forgot to check this in before.


tests/test_inference_config.py
```diff
@@ -0,0 +1,68 @@
...
+        task_type="multiclass", model_version=ModelVersion.V2
+    )
+    overrides = {
+        "PREPROCESS_TRANSFORMS": [
+            {
+                "name": "adaptive",
+                "append_original": "auto",
+                "categorical_name": "ordinal_very_common_categories_shuffled",
+                "global_transformer_name": "svd",
+                "subsample_features": -1,
```
bejaeger: heads-up that this flag will be changed. Maybe leave it out in this test?

oscarkey: ah good idea

PR #579, opened 2025-10-31
Changes to support TabPFN 2.5, once it is released.

oscarkey:
Fixes PRI-125

### DIFF
--- a/src/tabpfn/architectures/base/transformer.py
+++ b/src/tabpfn/architectures/base/transformer.py
@@ -22,6 +22,7 @@ from tabpfn.architectures.base.encoders import (
     SequentialEncoder,
 )
 from tabpfn.architectures.base.layer import PerFeatureEncoderLayer
+from tabpfn.architectures.base.thinking_tokens import AddThinkingTokens
 from tabpfn.architectures.interface import Architecture
 
 if TYPE_CHECKING:
@@ -38,7 +39,6 @@ class LayerStack(nn.Module):
         self,
         *,
         layers: Iterable[nn.Module],
-        recompute_each_layer: bool,
         min_num_layers_layer_dropout: int | None,
     ) -> None:
         super().__init__()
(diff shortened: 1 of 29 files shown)


oscarkey:
@noahho ready to merge?

src/tabpfn/architectures/base/transformer.py
```diff
@@ -543,6 +562,7 @@ def forward(  # noqa: PLR0912, C901
             ),
             single_eval_pos=single_eval_pos,
             cache_trainset_representation=self.cache_trainset_representation,
+            recompute_layer=recompute_layer,
         )  # b s f+1 e -> b s f+1 e
 
         # If we are using a decoder
```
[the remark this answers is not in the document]
noahho: wdyt? @oscarkey

src/tabpfn/model_loading.py
```diff
@@ -97,13 +99,43 @@ def get_regressor_v2(cls) -> ModelSource:  # noqa: D102
             filenames=filenames,
         )
 
+    @classmethod
+    def get_classifier_v2_5(cls) -> ModelSource:  # noqa: D102
+        filenames = [
+            "v2.5_classification_ii02f5gb_best_as_of_2867200.ckpt",
```
noahho, line 105: We want to change the names of these checkpoints to simpler cleaner names. Maybe look into this with @priorphil @LeoGrin

noahho: How about v2.5_classification_03_11_25_synth_1.ckpt (model-version, tasktype, date added, synth/real, if there were multiple added)

oscarkey: as this PR is syncing changes private->public, can we keep as is and update with the final names in a follow-up PR when they're ready?

src/tabpfn/architectures/base/transformer.py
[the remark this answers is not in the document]
oscarkey: This comment is true. We could fix this in a follow-up PR, but as this code is unused and not covered by any tests I suggest we just ignore the problem?

PR #584, opened 2025-11-03
Introduce option to calibrate temperature and tune decision thresholds

bejaeger:
## Motivation and Context

Add option to calibrate temperature and tune predictions to specified eval_metric.

Introduces 2 new arguments to `TabPFNClassifier`:
- `eval_metric`: The eval metric to use
- `tuning_config`: configuration of how the validation data is computed for temperature calibration and threshold tuning.
  - calibrate_temperature=True/False
  - tune_decision_thresholds=True/False
  - ...

For an example usage checkout [examples/tabpfn_with_tuning.py](./examples/tabpfn_with_tuning.py).

The interface is kept similar to autogluon's [TabularPredictor](https://auto.gluon.ai/stable/api/autogluon.tabular.TabularPredictor.html).

## Follow-ups:
- Improve heuristics for computing validation holdout set.
- Allow for an arbitrary metric to calibrate/tune

## Public API Changes

-   [ ] No Public API changes

The user now has the option to specify an `eval_metric` and a `tuning_config` in the initialization of TabPFNClassifier to calibrate temperature and tune thresholds to a specified metric.

## Checklist


src/tabpfn/inference_tuning.py
```diff
@@ -0,0 +1,408 @@
...
+        y_true: The true labels of shape [n_samples].
+        logits_to_probabilities_fn: The function to convert logits to probabilities.
+            The function should take an array of the shape of raw_logits and
+            a softmax temperature as argument.
+        current_default_temperature: The current default temperature.
+
+    Returns:
+        The temperature that minimizes the log loss.
+    """
+    temperatures = np.linspace(0.6, 1.4, 41)
```
noahho: This is very cheap so I'd add points here, making it e.g. 121

src/tabpfn/inference_tuning.py
```diff
@@ -0,0 +1,408 @@
...
+        holdout_frac: The percentage of the data to hold out for tuning.
+        n_splits: Number of stratified random splits to generate.
+        random_state: The random state to use for the split(s).
+
+    Returns:
+        Returns a list of splits as tuples of
+        (X_train_NtF, X_holdout_NhF, y_train_Nt, y_holdout_Nh).
+        Shape suffixes: Nt=num train samples, F=num features, Nh=num holdout samples.
+    """
+    splitter = StratifiedShuffleSplit(
```
noahho: I would change this to StratifiedKFold instead, that will be more robust. Currently we have overlapping splits. We can also do e.g. 5-fold but just evaluate two folds to get a 20% holdout but we make sure we get the data as diverse as possible

noahho (commented):
New learned attrs (softmax_temperature_, tuned_classification_thresholds_, eval_metric_) should be saved/loaded. I don’t see changes in save_fitted_tabpfn_model / load_fitted_tabpfn_model or state dicts to persist these.

src/tabpfn/classifier.py
```diff
@@ -399,6 +423,21 @@ class in Fine-Tuning. The fit_from_preprocessed() function sets this
...
+                calibrating the model's probabilities and tuning the decision
+                thresholds during the `fit()/predict()` calls. The tuning can be
+                enabled by configuring the `tuning_config` argument, see below.
+                For currently supported metrics, see
+                [tabpfn.classifier.ClassifierEvalMetrics][].
+
+            tuning_config:
+                The settings to use to tune the model's predictions for the specified
+                `eval_metric`. See
+                [tabpfn.inference_tuning.ClassiferTuningConfig][] for details
```
noahho:
```suggestion
                [tabpfn.inference_tuning.ClassifierTuningConfig][] for details
```

src/tabpfn/inference_tuning.py
```diff
@@ -0,0 +1,408 @@
...
+
+    Args:
+        metric_name: The name of the metric to optimize.
+        y_true: The true labels of shape [n_samples].
+        y_pred_probas: The predicted probabilities of shape [n_samples].
+
+    Returns:
+        The optimal threshold.
+    """
+    thresholds = np.linspace(0.01, 0.99, 99)
```
noahho: add more thresholds, e.g. 990

noahho (commented):
Not 100% certain about these but I think helpful:
New learned attrs (softmax_temperature_, tuned_classification_thresholds_, eval_metric_) should be saved/loaded. I don’t see changes in save_fitted_tabpfn_model / load_fitted_tabpfn_model or state dicts to persist these.

To avoid overfitting thresholds on small holdouts, add shrinkage toward 0.5 for the decision threshold and 1.0 for the temperature scale.

I'm also wondering, would it make sense to tune temperature per class?

noahho:
Add tests:
- ensure ROC AUC path doesn’t attempt threshold tuning.
- ensure F1 raises warning

src/tabpfn/inference_tuning.py
bejaeger: I increased the thresholds slightly but didn't go to testing 1k configurations. I think that's overkill, given what we've seen in the calibration curves?

src/tabpfn/inference_tuning.py
bejaeger: I increased it to 82. For regression with the 5k bar distribution this will add to the computation and since we're in numpy land this might become significant. Also, the temperature scaling curves we saw looked very smooth so I think it's safe to have a bit of a broader scan.

src/tabpfn/inference_tuning.py
bejaeger: Yes, good one! Changed it!

bejaeger:
Thanks for the review. Got caught up so this took a bit longer.
Some answers:
- Added tests for checking if warnings are thrown
- I only marginally increased the number of thresholds/temperatures we tested
- I decided not to include shrinkage in this PR because I think it adds too much complexity and because it's unclear what default values to choose. We throw a warning if tuning is specified but the dataset is small and I think this is enough for now. We can add shrinkage with good default parameters when we properly benchmark the tuning against different dataset sizes. What do you think?
- The fitted attrs are saved and loaded automatically. I added a test to check if the loaded values are as expected.

RE per-class temp tuning: yes, I think this can help and could be a good follow-up

noahho:
Perfect! Lgtm now :-) maybe just add an example to the examples directory that shows tuning on an imbalanced task?

noahho (approved):

PR #588, opened 2025-11-04
More robust performance test

bejaeger:
## Issue

Previously, the test was prone to statistical fluctuations because we were fitting a very small dataset.
This creates a dataset specific to the test and uses a few more samples to make the test more robust.


brendan-priorlabs (approved):
Thanks, @bejaeger!

PR #589, opened 2025-11-04
Update 'balance probabilities test' and 'get_embeddings' test

bejaeger:
## Issue

This fixes 3 tests for upcoming V2.5:

- The get_embeddings didn't consider thinking tokens. Change this so that "train_embeddings" only returns the embeddings for the data points
- Balance probabilities test was flawed. We don't really expect the probabilities to be uniform for any model. This introduces a new test for the functionality of balance probabilities, as well as a softer criterion for the classifier, to test whether the proba distribution is more uniform with balance-probabilities included.
- test_classifier_in_pipeline: This was using the balanced probabilities check as the previous test which didn't make too much sense because we only want to test whether the classifier works in a pipeline in this test.

### DIFF
--- a/src/tabpfn/architectures/base/transformer.py
+++ b/src/tabpfn/architectures/base/transformer.py
@@ -601,9 +601,11 @@ class PerFeatureTransformer(Architecture):
             )
 
             # out: s b e
-            thinking_rows_offset = 0
-            if self.add_thinking_tokens is not None:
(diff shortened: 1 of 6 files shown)


LeoGrin (approved):
LGTM! Agree that the previous balance probabilty test was too strong. (the new one might be a bit weak but don't have a great solution for that)

PR #592, opened 2025-11-04
Fix seed for random embeddings on CPU & MPS

bejaeger:
We need to fix the seed for CPU & MPS inference to one that works for datasets with a small number of features.
This means we also had to update the consistency tests.


src/tabpfn/architectures/base/transformer.py
```diff
@@ -635,8 +635,20 @@ def add_embeddings(  # noqa: C901, PLR0912
...
         else:
+            seed = self.random_embedding_seed
+
+            # For datasets with a small number of features (1, 2, or 3 features),
+            # the seed becomes important. Since during training, we generate
+            # random numbers on the GPU, the same seed on CPU or MPS can be suboptimal.
+            # For CPU and MPS, we therefore fix the seed to values that work
+            # well also for datasets with few features.
+            if x.device == torch.device("cpu"):
+                seed = 819
```
LeoGrin, line 646: oh yeah that's a good one

LeoGrin (approved):
woohoo

PR #597, opened 2025-11-05
Increase the default and v2.5 pretraining limits.

oscarkey:
To 50k rows x 2000 features.

Also make sure to specify all the values for the v2 and v2.5 default configs, to avoid future issues if the defaults in `InferenceConfig` are changed.

Fixes RES-814


bejaeger (approved):
LGTM!

PR #598, opened 2025-11-05
update pretraining limits docs

bejaeger:
Update docs with new pretraining limits


oscarkey:
ahhh nice! Should we add a note about the subsampling above 500 features? e.g. in case someone uses one estimator for some reason

bejaeger:
Yes, good one! Added it

oscarkey (approved):
sweeeeet

PR #599, opened 2025-11-05
Update the device doc string to include multi-gpu inference.

oscarkey:
We could enable multi-gpu inference when device="auto", but I suggest we leave it off for now until a couple of fixes to improve inference speed on small datasets are in? This way the user has to explicitly turn it on, and has more of a chance of seeing if it's faster or not.


bejaeger (approved):
Sounds good!

PR #600, opened 2025-11-05
Forward HF errors as they explain gating / missing tokens nicely.

priorphil:


priorphil:
Note: this essentially just removes a try / catch block, it's just the diff looking ugly.

PR #611, opened 2025-11-06
Anonymously track `model_version` and `model_path`

safaricd:
## Issue

- Anonymously track `model_version` and `model_path` when users have telemetry enabled.

## Public API Changes

-   [x] No Public API changes

---

## Checklist

---

### DIFF
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -594,6 +594,7 @@ def _log_model_config(
 
     path: Path = model_paths[0]
     # Check to avoid that we pass in arbitrary paths containing e.g. PII
+    # Ensure we whitelist model names so that no PII can be released.
     if path.name in model_source.filenames:
         set_model_config(path.name, version.value)
     else:


oscarkey:
I'll review this once the changes to https://github.com/PriorLabs/tabpfn_common_utils/pull/45 are in :)

src/tabpfn/model_loading.py
```diff
@@ -489,6 +490,17 @@ def load_model_criterion_config(
         )
     )
 
+    # Even though model_path can be a list of paths, we currently
+    # only use a single path
+    if len(resolved_model_paths) == 1:
```
oscarkey: What if we log the constant "multiple_paths" or something instead? As we might start using them soon.

src/tabpfn/model_loading.py
```diff
@@ -489,6 +490,17 @@ def load_model_criterion_config(
...
+        # Only set model config if the model path is in the list of available model
+        # filenames to avoid passing arbitrary paths containing e.g. PII
+        _filenames = _get_model_filenames(which, model_version.value)
+
+        _path = resolved_model_paths[0]
+        if _path.name in _filenames:
+            set_model_config(_path.name, model_version.value)
```
oscarkey: What about putting this whole bit in a function `_log_model_paths()` to reduce the complexity of `load_model_criterion_config()`?

src/tabpfn/model_loading.py
```diff
@@ -569,6 +581,28 @@ def _resolve_model_version(model_path: ModelPath | None) -> ModelVersion:
...
+    Returns:
+        A list of the available model file names.
+    """
+    # Convert string literals to enums
+    model_type = ModelType(which)
+    model_version = ModelVersion(version)
+
+    # Get the ModelSource which contains filenames
+    model_source = _get_model_source(model_version, model_type)
+    return [filename for filename in model_source.filenames]
```
oscarkey: can just inline `_get_model_source(version, ModelType(which))` instead? and delete this function and avoid converting version to a string and back to an enum

src/tabpfn/model_loading.py
safaricd: Good suggestion, sure! Added ✅

src/tabpfn/model_loading.py
safaricd: We'll tackle this once we start using them 😊

src/tabpfn/model_loading.py
safaricd: Simplified and pushed ✅

src/tabpfn/model_loading.py
```diff
@@ -569,6 +573,30 @@ def _resolve_model_version(model_path: ModelPath | None) -> ModelVersion:
...
+        version: The model version (currently only 'v2' or 'v2.5').
+    """
+    if len(model_paths) != 1:
+        return
+
+    model_type = ModelType(which)
+    model_source = _get_model_source(version, model_type)
+
+    path: Path = model_paths[0]
+    if path.name in model_source.filenames:
```
oscarkey, line 598: just one more thing: maybe comment here that we do this check to avoid logging pii?

brendan-priorlabs, line 598:
```suggestion
    # Ensure we whitelist model names so that no PII can be released.
    if path.name in model_source.filenames:
```

src/tabpfn/model_loading.py
```diff
@@ -569,6 +573,30 @@ def _resolve_model_version(model_path: ModelPath | None) -> ModelVersion:
...
+        set_model_config(path.name, version.value)
```
brendan-priorlabs, line 599: Nice, this looks safe! Maybe an extra bucket for everything else?

```suggestion
        set_model_config(path.name, version.value)
    else:
        set_model_config("OTHER", version.value)
```

src/tabpfn/model_loading.py
safaricd: Sure, happy to add that!

PR #614, opened 2025-11-07
Don't try to hf.login() in the example notebook.

oscarkey:
Because it's confusing to ask for the token without explaining? Instead the users will just hit the standard error about this when they try and fit, which is probably okay? I add a note to the text explaining this.

part of RES-847


oscarkey:
@noahho what do you think?

PR #618, opened 2025-11-11
Remove unlink of config.json.

priorphil:


brendan-priorlabs:
I think I'm still missing the why here. Where you able to track down who wrote the comment originally?

priorphil:
Close in favor of #649

PR #622, opened 2025-11-11
Add option to accept list of indices for subsample samples argument in inference config

bejaeger:
- [x] Add to Changelog

### DIFF
--- a/src/tabpfn/preprocessing.py
+++ b/src/tabpfn/preprocessing.py
@@ -579,7 +579,7 @@ class EnsembleConfig:
         cls,
         *,
         num_estimators: int,
-        subsample_samples: int | float | None,
+        subsample_samples: int | float | list[list[int]] | None,
         max_index: int,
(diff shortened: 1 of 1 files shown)


LeoGrin (approved):
LGTM thanks!

Nit: it would be cool to have a more end-to-end test with some mocking to make sure the indices are passed correctly to each estimator, but not sure if worth the time at this stage.

Nit: at some point we discussed adding a setter, did you decide against it or is it for another PR?

bejaeger:
Thanks for the review!

RE e2e tests: I think this one should be safe (as safe as the current code is) cause we don't change the handling after the subsample_indices are defined. We just parse them to the EnsembleConfig. 

RE setter: Right! Forgot to comment on it. Adding a setter would have only created some boilerplate code and didn't actually make using it outside of this class much better. So I decided to leave the estimator classes more untouched.

PR #623, opened 2025-11-12
Move cuda sync to the consumer thread during multi-gpu inference.

oscarkey:
- Wait for the cuda computation to complete on the main thread rather than the worker thread. This allows the worker thread to start queuing the next estimator rather than waiting.
- Don't create an additional cuda stream for each estimator, as each device has it's own default stream so can operate independently anyway. We could reintroduce additional streams if we wanted to overlap the copy of the output between GPUs with the computation, but this would make the inference code complex and probably has a tiny effect, so I decide not to do it.

Does it make it faster? Only very slightly for larger datasets, but it's nice to do things "properly":
```
before change:
                                                  duration_per_call_seconds      
                                                                       mean   std
train test cols n_gpus wrapper.memory_saving_mode                                
50    200  10   2      False                                           0.44  0.04
200   200  10   2      False                                           0.43  0.04
20000 1000 100  2      True                                            7.32  0.07

after change:
                                                  duration_per_call_seconds      
                                                                       mean   std
train test cols n_gpus wrapper.memory_saving_mode                                
50    200  10   2      False                                           0.43  0.04
200   200  10   2      False                                           0.43  0.04
20000 1000 100  2      True                                            7.14  0.07
```
This is over 10 repeats. This makes sense because: for small datasets, the GPU keeps up with the CPU so the sync is instant. For large datasets, we do spend some time waiting.

To validate it, I ran the benchmarking suite with and without the fix, using 2 or 4 gpus:
- 2 gpus: [multiclass results](https://scale.neptune.ai/o/PriorLabs/org/Benchmarking/runs/compare?viewId=a0307dbb-90ed-4335-95f3-9629aa5b348c&dash=charts&compare=uGH9bg7OjaIlBQyyVMR_2U-Yk3zLXJ4mlff1mW2ljgHk&showOnlySelectedRuns=true), [regression results](https://scale.neptune.ai/o/PriorLabs/org/Benchmarking/runs/compare?viewId=a0307dbb-90ed-4335-95f3-9629aa5b348c&dash=charts&compare=uSjWydJrm91SlQK5GwwFbXQkwVTsu7yliA2tqalI5KJU&showOnlySelectedRuns=true)
- 4 gpus: [multiclass results](https://scale.neptune.ai/o/PriorLabs/org/Benchmarking/runs/compare?viewId=a0307dbb-90ed-4335-95f3-9629aa5b348c&dash=charts&compare=ulO7BIYpRkeszCkIYJrRUE_1TglVZ6zyfv2CbOn3ZCws&showOnlySelectedRuns=true), [regression results](https://scale.neptune.ai/o/PriorLabs/org/Benchmarking/runs/compare?viewId=a0307dbb-90ed-4335-95f3-9629aa5b348c&dash=charts&compare=uhGYBqMQDW8yz47qAzl9NCmyH4Z_78aiKEx6ixULN7iA&showOnlySelectedRuns=true)

Part of RES-797

### DIFF
--- a/src/tabpfn/parallel_execute.py
+++ b/src/tabpfn/parallel_execute.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import queue
 from collections.abc import Generator, Iterable, Sequence
 from multiprocessing.pool import ThreadPool
-from typing import Generic, Protocol, TypeVar
+from typing import Callable, Generic, Protocol, TypeVar
 
 import torch
 
@@ -17,9 +17,6 @@ class ParallelFunction(Protocol, Generic[R_co]):
 
     def __call__(self, *, device: torch.device, is_parallel: bool) -> R_co:
         """Execute the function.
-
-        If using CUDA, `parallel_execute()` will set the current stream, and this
-        function should not change it.
 
         Args:
             device: PyTorch device that all computation should be performed on.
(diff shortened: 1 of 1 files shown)


src/tabpfn/parallel_execute.py
```diff
@@ -84,31 +81,38 @@ def _execute_with_multithreading(
...
-                # we synchronize before returning so that the output is ready for the
-                # consumer. It would be more efficient for the consumer to wait, so this
-                # thread can start with the next function, but this approach is simpler.
-                stream.synchronize()
-                return output
+
+                # The output will be consumed on a different cuda stream, which needs to
+                # wait for the computation on this stream to be complete. Thus we insert
+                # "ready" event after the model evaluation, and return a function to the
+                # consumer that waits on this event.
```
brendan-priorlabs, line 103: I think this is just me working with a half(read: quarter)-baked mental model of CUDA here, but if we're using the default stream on each device why is it safe to synchronize in the consumer allowing other work to proceed on the same device in the meantime?

brendan-priorlabs (commented):
Very clean on the code level. Just have a question about CUDA...

src/tabpfn/parallel_execute.py
oscarkey: Could you explain a little more what you mean? It's definitely possible I missed something as I'm not that familiar with this either.

What I'm trying to do is get the consumer thread to block until an estimator's kernels are complete, and the output is written to memory. I think it's okay for other kernels to execute before the output is read, as the output memory will still be allocated and so won't be overwritten.

We might end up with an execution sequence that looks like this:
`...M1 M1 M1 M1 E1 M2 M2 M2 C1 M2 E2 C2...`
where M1 are kernels that evaluate the model during estimator 1, E1 the "done" event for estimator 1, C1 the kernel consuming the output of estimator 1, and M2, E2, C2 similar kernels for estimator 2.
M1, E1, M2, E2 are queued by the worker thread, while C1 and C2 are queued by the consumer thread. This PR ensures that C1 is not queued until M1 is finished.

A more efficient option would be to use `E1.wait()`, which effectively queues a kernel on a particular stream that waits for E1 to complete (rather than blocking on the cpu, as in this PR). But that's more error prone as we have to know which stream C1 will be queued on. In practice, C1 is often copying the output of the estimator from GPU x to GPU 0 via DMA, and I'm not sure if this is queued on GPU 0 or GPU x...

brendan-priorlabs: Thanks for breaking down for me. Super helpful!

> I think it's okay for other kernels to execute before the output is read, as the output memory will still be allocated and so won't be overwritten.

This was the key part I was fuzzy on, but now that you state it explicitly it seems natural.

PR #625, opened 2025-11-12
Refactor to move restoring models into InferenceEngine.load_state().

oscarkey:
This preparatory work for caching the models on device during inference.

At the same time, throw an exception if the model being saved uses a KV cache, as this doesn't work (the KV cache was never saved, so I assume you would get junk predictions after loading). It would be nicer and fairly easy to make this work, but avoid it for now.

Notes:
- Don't deepcopy the models when adding them to the inference engine. We don't do this during normal inference, so I don't think we need to do it in this case.
- No longer move the models to the target device. We move them at the start of inference anyway, and this is consistent with the behaviour after normal initialisation.          

Part of RES-797

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -121,7 +121,7 @@ class InferenceEngine(ABC):
         This does not support the KV cache, and will raise an error if this is an
         InferenceEngineCacheKV.
         """
-        _raise_if_kv_cache_enabled(self)
+        _raise_if_kv_cache_enabled_on_save_or_load(self)
 
         state_copy = deepcopy(self)
         state_copy.models = None  # type: ignore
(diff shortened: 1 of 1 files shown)


src/tabpfn/inference.py
```diff
@@ -118,19 +118,36 @@ def use_torch_inference_mode(self, *, use_inference: bool) -> None:
...
+        """
+        engine: InferenceEngine = joblib.load(Path(path))
+
+        _raise_if_kv_cache_enabled(engine)
+
+        engine.models = list(models)
+        return engine
+
+
+def _raise_if_kv_cache_enabled(engine: InferenceEngine) -> None:
```
bejaeger:
```suggestion
def _raise_if_kv_cache_enabled_on_save_or_load(engine: InferenceEngine) -> None:
```

bejaeger (approved):
LGTM!

src/tabpfn/inference.py
oscarkey: nice idea, done!

Issue #523, opened 2025-09-22
Improve inference speed by keeping model on GPU/MPS between fit/predict calls


DanieleMorotti:
I tested the example code for the classifier:
```python
from sklearn.datasets import load_breast_cancer
from sklearn.metrics import accuracy_score, roc_auc_score
from sklearn.model_selection import train_test_split

from tabpfn import TabPFNClassifier


# Load data
X, y = load_breast_cancer(return_X_y=True)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5, random_state=42)

# Initialize a classifier
clf = TabPFNClassifier(device="mps")
clf.fit(X_train, y_train)
print("Device: ", next(clf.executor_.model.parameters()).device)
# It prints "cpu"

# Predict probabilities
prediction_probabilities = clf.predict_proba(X_test)
print("ROC AUC:", roc_auc_score(y_test, prediction_probabilities[:, 1]))
# Predict labels
predictions = clf.predict(X_test)
print("Accuracy", accuracy_score(y_test, predictions))

clf.executor_.model = clf.executor_.model.to("mps")
print("New device: ", next(clf.executor_.model.parameters()).device)
# Now it prints "mps:0"
```

When I print the device of the model parameters, they appear to be on the `cpu`, even though `mps` is available.
Is this the expected behavior, or is it unintended? The same issue occurs when I set the device to `auto`.

## Debug info
PyTorch version: 2.8.0
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 15.6.1 (arm64)
GCC version: Could not collect
Clang version: 17.0.0 (clang-1700.3.19.1)
CMake version: version 3.31.5
Libc version: N/A

Python version: 3.12.1 (main, Jun  3 2024, 17:33:54) [Clang 15.0.0 (clang-1500.3.9.4)] (64-bit runtime)
Python platform: macOS-15.6.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M2 Pro

Dependency Versions:
--------------------
tabpfn: 2.2.1
torch: 2.8.0
numpy: 2.3.3
scipy: 1.16.2
[...]

PR #628, opened 2025-11-12
Cache the models on each device between estimators.

oscarkey:
Rather than copying the model at the start of each estimator. This does not yet keep the models on the device between inference calls (see https://github.com/PriorLabs/TabPFN/issues/523).

Notes:
- Remove sorting the estimators by model index, as now the models are kept on the device this is no longer required? I think this is okay, but unsure? @bejaeger 
- There is a risk that this change increases GPU memory usage, leading to OOMs. However, the model is very small so it's not a big risk. Continue to copy the KV cache from/to the target device for each estimator.
- It's difficult to write explicit tests for this as we only have single-device executors, but it's partially covered by our existing tests

The speedup can be about 20%, for small datasets:
```
 -- Before change --
                       duration_per_call_seconds       peak_memory_GB      
                                            mean   std           mean   std
train test cols n_gpus                                                     
100   100  50   2                           0.63  0.01           0.16  0.04
           100  2                           0.74  0.14           0.32  0.02
500   100  50   2                           0.65  0.00           0.36  0.00
           100  2                           0.81  0.12           0.42  0.00
1000  100  50   2                           0.69  0.01           0.42  0.00
           100  2                           1.03  0.01           0.53  0.00
10000 100  50   2                           2.85  0.43           1.45  0.00
           100  2                           4.70  0.23           2.53  0.00

 -- After change --
                        duration_per_call_seconds       peak_memory_GB      
                                            mean   std           mean   std
train test cols n_gpus                                                     
100   100  50   2                           0.50  0.02           0.16  0.04
           100  2                           0.54  0.00           0.32  0.02
500   100  50   2                           0.53  0.00           0.36  0.00
           100  2                           0.65  0.07           0.42  0.00
1000  100  50   2                           0.57  0.01           0.42  0.00
           100  2                           0.89  0.01           0.53  0.00
10000 100  50   2                           2.55  0.09           1.45  0.00
           100  2                           4.69  0.13           2.53  0.00
```
(results on 2xA100 80GB, mean + std over 4 repeats of each config)

Part of RES-797

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Don't select MPS devices below PyTorch 2.5 and raise an error if selected, due to poor performance [#619](https://github.com/PriorLabs/TabPFN/pull/619)
-- In multi-GPU inference, cache the model(s) on each device between estimators, to improve speed
+- In multi-GPU inference, cache the model(s) on each device between estimators, to improve speed [#628](https://github.com/PriorLabs/TabPFN/pull/628)
 
 ## [6.0.6] - 2025-11-10
 


src/tabpfn/inference.py
```diff
@@ -808,3 +788,70 @@ def _move_and_squeeze_output(
...
+
+            if not_on_device:
+                if multiple_devices:
+                    self._on_device_cache[device] = deepcopy(self._model)
+                else:
+                    self._on_device_cache.clear()
+                    self._on_device_cache[device] = self._model
+
+        if not_on_device:
+            self._on_device_cache[device].to(device)
```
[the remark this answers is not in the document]
oscarkey: I think this is correct, because we know that only one thread will access each device at a time. But perhaps this should be enforced, or listed in the docs?

src/tabpfn/inference.py
```diff
@@ -808,3 +790,73 @@ def _move_and_squeeze_output(
...
+        """
+        with self._on_device_cache_lock:
+            not_on_device = device not in self._on_device_cache
+
+            if not_on_device:
+                if multiple_devices:
+                    self._on_device_cache[device] = deepcopy(self._model)
+                else:
+                    self._on_device_cache.clear()
+                    self._on_device_cache[device] = self._model
```
bejaeger, line 829:
```suggestion
                    self._on_device_cache[device] = deepcopy(self._model) if multiple_devices else self._model
```

Then we can remove the if/else which I think is harder to read.

src/tabpfn/inference.py
```diff
@@ -808,3 +790,73 @@ def _move_and_squeeze_output(
...
+            # cover both cases by emptying the cache and moving _model.
+            self._on_device_cache.clear()
+            self._model.cpu()
+
+    def set_dtype(self, dtype: torch.dtype) -> None:
+        """Set the dtype of the model's parameters."""
+        with self._on_device_cache_lock:
+            self._model.type(dtype)
+            for model in self._on_device_cache.values():
+                model.type(dtype)
```
bejaeger, line 851: `model.type(dtype)` doesn't change the `model` but returns a copy doesn't it?

src/tabpfn/inference.py
```diff
@@ -808,3 +790,73 @@ def _move_and_squeeze_output(
...
+
+        if not_on_device:
+            self._on_device_cache[device].to(device)
+
+        return self._on_device_cache[device]
+
+    def to_cpu(self) -> None:
+        """Remove the models from the target devices, keeping one copy on the CPU."""
+        with self._on_device_cache_lock:
+            # If .get() was called with is_parallel=True, then ._model will remain on
```
bejaeger: `is_parallel` -> `multiple_devices` ?

src/tabpfn/inference.py
```diff
@@ -808,3 +790,73 @@ def _move_and_squeeze_output(
...
+            # the CPU and only the cached models will be moved to the target devices. If
+            # is_parallel=False, then ._model will be moved to the target device. Thus,
+            # cover both cases by emptying the cache and moving _model.
+            self._on_device_cache.clear()
```
bejaeger, line 843: Should we force deleting the models here or trust the python garbage collector?

src/tabpfn/inference.py
oscarkey: ah good catch, fixed.

src/tabpfn/inference.py
oscarkey: Tensor.type() creates a new tensor, Module.type() modifies in-place: https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.type 😰

src/tabpfn/inference.py
oscarkey: we could delete but I'm not sure how to? I wondered about the `del` keyword, but it turns out it just unbinds the name so is the same as `.clear()`. The good news is that it seems the objects will probably be immediately garbage collected, unless there are cyclic references: https://pytorch.org/blog/understanding-gpu-memory-2/

src/tabpfn/inference.py
oscarkey: what about `self._on_device_cache.clear()`?

src/tabpfn/inference.py
bejaeger: Thanks for double-checking haha. I would have thought they are harmonized.

src/tabpfn/inference.py
bejaeger: Oops! sorry for noise!

src/tabpfn/inference.py
bejaeger: Sounds good to me!

src/tabpfn/inference.py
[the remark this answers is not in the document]
oscarkey: added comment

PR #629, opened 2025-11-14
Remove `kditransform` dependency

bejaeger:
Having the `kditransform`  in our dependencies caused dependency conflicts.
Since we don't use them in our defaults, we can remove this dependency here. It will still be installed in our extension package.

Note that this also fixes the inconsistency test for the new macOS runner and fixes code style

### DIFF
--- a/src/tabpfn/preprocessors/kdi_transformer.py
+++ b/src/tabpfn/preprocessors/kdi_transformer.py
@@ -16,12 +16,6 @@ try:
 
     # This import fails on some systems, due to problems with numba
 except ImportError:
-    warnings.warn(
-        "Cannot use KDITransformer because kditransform is not installed. "
-        "Using PowerTransformer as fallback.",
-        UserWarning,
-        stacklevel=2,
-    )
     KDITransformer = PowerTransformer  # fallback to avoid error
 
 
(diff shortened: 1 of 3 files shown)


src/tabpfn/preprocessors/kdi_transformer.py
```diff
@@ -44,6 +45,16 @@ class KDITransformerWithNaN(KDITransformer):
     mean values and then fills the NaN values with NaNs after the transformation.
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        if KDITransformer is PowerTransformer:
```
LeoGrin: nit: maybe out of scope for this PR but could be make it SafePowerTransformer?

LeoGrin (approved):

src/tabpfn/preprocessors/kdi_transformer.py
bejaeger: Changed it!

bejaeger: okay, reverted it again because of a circular import. don't think it's too important here since Ideally nobody uses this when PowerTransform is actually used.

LeoGrin: yep agree!

PR #630, opened 2025-11-18
Add GitHub workflow that tests tabpfn-extensions and -time-series.

oscarkey:
Checkout main of these repos, install the current ref of TabPFN, and run the repos' tests.

Downsides of this approach:
- Doesn't test all the dependency version combinations
- Duplicates workflow configuration between repos

Upside: simple and easy, probably catches most things?

Fixes PRI-97

### DIFF
--- a/.github/workflows/first_party_compatibility.yml
+++ b/.github/workflows/first_party_compatibility.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: pr-${{ github.ref }}
+  group: compatibility-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:


oscarkey:
@brendan-priorlabs any idea why the tests might be failing to download the gated model? With an auth error

.github/workflows/first_party_compatibility.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Set up uv
+        uses: astral-sh/setup-uv@5a7eac68fb9809dea845d802897dc5c723910fa3 # v7.1.3
+
+      - name: Install dependencies and run tests
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          FAST_TEST_MODE: 1 # Configures -extensions to run a smaller set of tests.
+        run: |
+          cd ${{ matrix.repo-name }}
+          uv run --with ../TabPFN --all-extras pytest -m "not uses_tabpfn_client"
```
brendan-priorlabs, line 50: Were you able to test that the `--with` is working as intended in this workflow? Bit worried about the behavior if (as in this case) there are multiple versions of the library installed. I think importing something from the repo and then using `.__file__` would be sufficient

brendan-priorlabs (approved):
LGTM! Left an idea for some extra testing, but the change itself looks good :)

.github/workflows/first_party_compatibility.yml
oscarkey: It seems like it works:
```
$ cd tabpfn-extensions
$ source .venv/bin/activate
$ uv sync
Resolved 187 packages in 5ms
Audited 97 packages in 5ms
$ uv run python -c "import tabpfn; print(tabpfn.__file__)"   
/Users/oscar/Documents/repos/tabpfn-extensions/.venv/lib/python3.10/site-packages/tabpfn/__init__.py
$ uv run --with ../TabPFN python -c "import tabpfn; print(tabpfn.__file__)"
      Built tabpfn @ file:///Users/oscar/Documents/repos/TabPFN
Installed 52 packages in 335ms
/Users/oscar/.cache/uv/archive-v0/hYgBaVyu0a2E3yCsTnF7d/lib/python3.10/site-packages/tabpfn/__init__.py
```

PR #632, opened 2025-11-19
Put model checkpoints in cache dir during tests.

oscarkey:
At the moment these are put in the cwd, which is annoying when working
on your local machine.

Towards RES-878

PR #633, opened 2025-11-19
Speed up test_classifier_interface and include v2.5

oscarkey:
- Reduce the number of combinations tested, by splitting up the big `test_fit` test.
- Reduce the test dataset size
- Include v2.5 checkpoints.

On my laptop, was over twice as fast, despite including v2.5

Once this is approved, I will apply the same to test_regressor_interface.

Part of RES-878


oscarkey:

- #632
    - #633 ◀

PR #634, opened 2025-11-19
Don't run most of the MPS tests during PRs.

oscarkey:
The MPS tests seem to be twice as slow as the CPU tests. Replace them
with a simple smoke test, and run most of them only on merge to main.

Add a "non_ci" marker to pytest, and mark the MPS tests with this.
Update the CI workflow to apply "not non_ci" during PRs.

Also, make sure tests that test on TABPFN_EXCLUDE_DEVICES override this
variable rather than assuming it's emtpy. Otherwise they break if the CI
sets it.

Part of RES-878

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: pytest tests/
 
-      - name: Run Tests (CI tests only)
+      - name: Run Tests (PR tests only)
         if: ${{ matrix.dependency-set != 'lowest-direct' && github.event_name == 'pull_request' }}
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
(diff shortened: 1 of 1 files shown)


oscarkey:

- #635
    - #634 ◀
        - #637

.github/workflows/ci.yml
```diff
@@ -84,22 +84,35 @@ jobs:
...
+      - name: Run Tests (all tests)
+        if: ${{ matrix.dependency-set != 'lowest-direct' && github.event_name != 'pull_request' }}
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
-        if: ${{ matrix.dependency-set != 'lowest-direct' }}
         run: pytest tests/
 
-      - name: Run Tests (without MPS)
-        if: ${{ matrix.dependency-set == 'lowest-direct' }}
+      - name: Run Tests (CI tests only)
```
brendan-priorlabs: This naming is kinda confusing since "all tests" above also runs on our Github CI this whole file is called ci.yaml. Maybe the naming should contain the main vs. PR distinction. So `non_pr_test` or something like that

oscarkey: ah good point. I changed it to "not_in_prs"?

oscarkey:
Ah sorry the PR description was from a previous interation, see what you think now.

brendan-priorlabs (approved):
Ah, great! SGTM!

PR #635, opened 2025-11-19
Speed up test_regressor_interface and include v2.5 checkpoints.

oscarkey:
Reduce the number of tests, and reduce the dataset size.

Also only do the bigger sweep for the default checkpoint, both here and
for the classifier.

Seems to be about twice as fast, on my laptop.

Part of RES-878


oscarkey:

- #633
    - #635 ◀
        - #634

LeoGrin (approved):
Thanks!

PR #639, opened 2025-11-21
Test both v2 and v2.5 in consistency tests, and refactor the tests.

oscarkey:
Include support for different reference predictions on different platforms/devices. For now, just include MacOS with apple silicon, as this is what's easy to generate on our laptops. I tried out a variety of different PyTorch and Python versions on my laptop, and they all seemed to result in the same reference predictions.

Also fix an in the CI workflow: it selected tests that matched "not non_ci", whereas the marker is actually "not_in_prs", so it should be "not not_in_prs". This resulted in more tests than necessary running during PRs.

One issue is that the consistency tests can be silently skipped if none of the CI platforms match. It seems unlikely this will happen now we just match against macos arm64. What do you think?

Fixes RES-841

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
         if: ${{ matrix.dependency-set != 'lowest-direct' && github.event_name == 'pull_request' }}
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
-        run: pytest -m "not non_ci" tests/
+        run: pytest -m "not not_in_prs" tests/
 
       # We don't support MPS below PyTorch 2.5 (see tabpfn.utils.infer_devices()), thus
       # disable MPS for the lowest-direct dependency set.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         if: ${{ matrix.dependency-set != 'lowest-direct' && github.event_name == 'pull_request' }}
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
-        run: pytest -m "not non_ci" tests/
+        run: pytest -m "not not_in_prs" tests/
```
priorphil, line 97: The "not not" triggered me a bit :) can leave as is for now, but what do you think of renaming this to is_mps since it's used for disabling mps only?

priorphil (approved):
Love code deletions, thanks! :)

.github/workflows/ci.yml
oscarkey: yes you're right that is better haha, I'll make a followup.

PR #646, opened 2025-11-25
Improved analytics for tracking usage of different fit modes

safaricd:
## Issue

Previously, we had no knowledge of the different `fit_mode` params that were used, meaning it was difficult to decide on the exact focus area for improving any of the fit modes. 

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

Unit tests and manual testing.

## Checklist

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   # Once Python 3.10 is the minimum version, this can be removed.
   "eval-type-backport>=0.2.2",
   "joblib>=1.2.0",
-  "tabpfn-common-utils[telemetry-interactive]>=0.2.8",
+  "tabpfn-common-utils[telemetry-interactive]>=0.2.11",
 ]
 requires-python = ">=3.9"
 authors = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   # Once Python 3.10 is the minimum version, this can be removed.
   "eval-type-backport>=0.2.2",
   "joblib>=1.2.0",
-  "tabpfn-common-utils[telemetry-interactive]>=0.2.11",
+  "tabpfn-common-utils[telemetry-interactive]>=0.2.13",
 ]
 requires-python = ">=3.9"
 authors = [
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -31,7 +31,7 @@ import torch
 from sklearn import config_context
 from sklearn.base import BaseEstimator, ClassifierMixin, check_is_fitted
 from sklearn.preprocessing import LabelEncoder
-from tabpfn_common_utils.telemetry import track_model_call, set_init_params
+from tabpfn_common_utils.telemetry import track_model_call
 
 from tabpfn.base import (
     ClassifierModelSpecs,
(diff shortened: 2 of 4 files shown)


src/tabpfn/classifier.py
```diff
@@ -479,6 +479,9 @@ class in Fine-Tuning. The fit_from_preprocessed() function sets this
         self.tuning_config = tuning_config
         initialize_telemetry()
 
+        # Only anonymously record `fit_mode` usage
+        set_init_params({"fit_mode": self.fit_mode})
```
oscarkey: Should we do the same thing as for model_path and validate that it's a known fit mode? To avoid accidentally collecting PII.
We could define `FitMode = Literal["low_memory","fit_preprocessors","fit_with_cache","batched"]` in inference.py, import it here and in the regressor interface, and then use `typing.get_args()` to check the provided one is valid?

noahho (approved):

noahho:
FYI: Merging the most recent code will fix the PR issue you have.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
   # Once Python 3.10 is the minimum version, this can be removed.
   "eval-type-backport>=0.2.2",
   "joblib>=1.2.0",
-  "tabpfn-common-utils[telemetry-interactive]>=0.2.8",
+  "tabpfn-common-utils[telemetry-interactive]>=0.2.11",
```
noahho: You updated pyproject.toml to 0.2.11, but the comment in model_loading.py says:

"...this interface was introduced in tabpfn_common_utils 0.2.13 and not all users have upgraded..."

should it be >=0.2.13?

src/tabpfn/model_loading.py
```diff
@@ -653,6 +654,55 @@ def _log_model_config(
         set_model_config("OTHER", version.value)
 
 
+def log_model_init_params(
+    estimator: TabPFNClassifier | TabPFNRegressor, params: dict[str, Any]
+) -> None:
+    """Anonymously log the some model initialization parameters for anonymized
```
noahho:
```suggestion
    """Anonymously model initialization parameters for anonymized
```

pyproject.toml
safaricd: We can set it to a minimum of 0.2.13 but whenever someone installs TabPFN the latest version will be pulled because we don't fix it. So, both approaches work & are valid 😊

src/tabpfn/model_loading.py
noahho: Would it actually make sense to move this code into the telemetry utils? Ideally we have the telemetry code not in this package but just linked here

noahho: It seems like this code could be moved?

pyproject.toml
noahho: Hmm but if the user already has 0.2.11 installed that would be outdated? With this logic any minimum settings wouldnt matter

noahho: You should set the minimum version that is compatible with your code

safaricd: > Hmm but if the user already has 0.2.11 installed that would be outdated? With this logic any minimum settings wouldnt matter

If the user already has a previous TabPFN version, running `pip install --upgrade tabpfn` will also upgrade `tabpfn_common_utils` to the latest available version -> `0.2.13`. `uv` is even more aggressive on that front. But anyway, for the sake of this PR, happy to push the change.

PR #647, opened 2025-11-26
Add save_peak_memory_factor and force_recompute_layer arch interface.

oscarkey:
Move save_peak_memory_factor from function that sets state to parameter of forward(), as we always set it for each forward pass anyway as it depends on the size of the input.

Also:
- Disable Ruff documentation checking for `@override` and `@overload` methods, to avoid duplicating documentation from the parent method. Downside is this removes enforcement of documentation of differences from parent, but I think this is okay?
- Update ruff hook name in pre-commit. The old name was deprecated.

Fixes RES-909

### DIFF
--- a/tests/test_classifier_interface.py
+++ b/tests/test_classifier_interface.py
@@ -38,6 +38,7 @@ from .utils import (
     get_pytest_devices,
     is_cpu_float16_supported,
     mark_mps_configs_as_slow,
+    patch_layernorm_no_affine,
 )
 
 exclude_devices = {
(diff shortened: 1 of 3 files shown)


src/tabpfn/architectures/base/transformer.py
```diff
@@ -334,32 +309,15 @@ def forward(  # noqa: PLR0912, C901
...
-                - `None`, if there are no training rows, as when making predictions
-                  using the KV cache.
-            only_return_standard_out: Whether to only return the standard output.
-            categorical_inds: The indices of categorical features.
-            style: The style vector.
-            data_dags: The data DAGs for each example in the batch.
-            force_recompute_layer: Whether to force to recompute layers (i.e.
-            perform activation checkpointing for each layer).
+            force_recompute_layer: If True, enable activation checkpointing for each
+                Transformer block. Otherwise, checkpoint as set in the config.
```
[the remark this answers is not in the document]
oscarkey: I deliberately include force_recompute_layer here to give the specifics of which layers are checkpointed for this architecture.

src/tabpfn/architectures/base/memory.py
```diff
@@ -95,12 +92,11 @@ def method_(
 MemorySavingMode = Union[bool, Literal["auto"], float, int]
 
 
-def set_save_peak_memory(model: Architecture, *, enabled: bool) -> None:
-    """Set the peak memory factor to the default value, if enabled."""
+def get_save_peak_memory_factor(*, enabled: bool) -> int | None:
+    """Get the default save peak memory factor, or None if disabled."""
```
priorphil: This seems like a weird function given that we pass enabled and it's just a constant - can we import the constant directly?

src/tabpfn/architectures/base/memory.py
```diff
@@ -5,15 +5,12 @@
...
 
 import torch
 
 from tabpfn.settings import settings
 
-if TYPE_CHECKING:
-    from tabpfn.architectures.interface import Architecture
-
 os.environ["PYTORCH_CUDA_ALLOC_CONF"] = settings.pytorch.pytorch_cuda_alloc_conf
 SAVE_PEAK_MEM_FACTOR = 8
```
priorphil: Should we call this DEFAULT_SAVE_PEAK_MEM_FACTOR since we can also set it manually?

priorphil (approved):

src/tabpfn/architectures/base/memory.py
oscarkey: yes good point, done!

src/tabpfn/architectures/base/memory.py
oscarkey: done!

PR #649, opened 2025-11-26
Fix config.json behavior when calling download_model directly

brendan-priorlabs:
Previously we handled checking for the cached model artifact only at a higher level, but we do have some direct callers to `download_model` that expect it to behave similarly

### DIFF
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -212,6 +212,8 @@ def _try_huggingface_downloads(
             warnings.filterwarnings("ignore")
 
         try:
+            should_download_config = not base_path.exists()
+
             # Download model checkpoint
             local_path = hf_hub_download(
                 repo_id=source.repo_id,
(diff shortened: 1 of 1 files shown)


src/tabpfn/model_loading.py
```diff
@@ -212,6 +212,8 @@ def _try_huggingface_downloads(
             warnings.filterwarnings("ignore")
 
         try:
+            should_download_config = not base_path.exists()
```
priorphil, line 215: Just to double check, the base_path is actually the model file path right? Got confused for a second as `download_model` says this is a directory https://github.com/PriorLabs/TabPFN/blob/d19af13d13479427f05f0dcab90513628252f8a2/src/tabpfn/model_loading.py#L334 but I guess it's called https://github.com/PriorLabs/TabPFN/blob/d19af13d13479427f05f0dcab90513628252f8a2/src/tabpfn/model_loading.py#L386 which means it includes the model filename.

priorphil (approved):
Thanks!

src/tabpfn/model_loading.py
brendan-priorlabs: Good catch on the stale comment. I verified locally that it's behaving as expected and updated the comment to:
```        to: The file path to download the model to.```

brendan-priorlabs:
Thanks for the review, @priorphil!

Issue #631, opened 2025-11-18
Fit modes give different results


oscarkey:
### Describe the bug

Different settings of the `fit_mode` option give different outputs. The fit mode shouldn't affect the output, up to float precision issues.

### Steps/Code to Reproduce

```python
import random
from sklearn.datasets import load_diabetes
from sklearn.model_selection import train_test_split
import numpy as np
import torch

from tabpfn import TabPFNRegressor

X, y = load_diabetes(return_X_y=True)
X_train, X_test, y_train, y_test = train_test_split(
    X, y, test_size=0.33, random_state=42
)


def _set_seeds() -> None:
    torch.manual_seed(0)
    np.random.seed(0)
    random.seed(0)


_set_seeds()
reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache = reg.predict(X_test)

reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache_repeat = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_preprocessors", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_cache_preproc = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_with_cache", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_kv_cache = reg.predict(X_test)


def _max_diff(a: np.ndarray, b: np.ndarray) -> float:
    return np.max(np.abs(a - b) / np.abs(a))


print("max relative diffs")
print("no_cache vs no_cache_repeat:", _max_diff(preds_no_cache, preds_no_cache_repeat))
print("no_cache vs cache_preproc:", _max_diff(preds_no_cache, preds_cache_preproc))
print("no_cache vs kv_cache:", _max_diff(preds_no_cache, preds_kv_cache))
```

I set `TABPFN_EXCLUDE_DEVICES=mps` to get float64 support on MacOS.

### Expected Results

```
no_cache == no_cache_repeat: 0.0
no_cache == cache_preproc: ~0.0
no_cache == kv_cache: ~0.0
```

### Actual Results

```
no_cache == no_cache_repeat: True
no_cache == cache_preproc: 0.044
no_cache == kv_cache: 1.423
```

### Versions

```shell
PyTorch version: 2.9.0
[...]

PR #650, opened 2025-11-28
Fix issue fit modes different results

magnus-priorlabs:
## Issue
Fixed the issue mentioned here https://github.com/PriorLabs/TabPFN/issues/631
 where different fit-modes give different results.
Two main factors led to the issue:

- The preprocessings were initialized at different steps in the code (depending on fit-mode) using seeds that were sampled on the spot. Consequently, the preprocessing function "fit_preprocessors" got different seeds depending on the fit-mode. Now the same seeds, invariant of fit-mode, initialize preprocessing.

- The transformer forward pass used thinking tokens twice for fit_with_cache mode—once during inference and once during fitting. This is now solved.

There is a small (< ~1e-4) difference in fit-mode predictions when using fit_with_cache, due to some precision errors in the attention (I think in torch.nn.functional.scaled_dot_product_attention, depending on the batch size) in the "attention_between_features" call. I take this as a tolerable performance difference.

Fixes PRI-154

## Motivation and Context

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
locally. Also re-enabled pytests. And using a script (https://gist.github.com/magnus-priorlabs/ca9ff20a38cd09a8a3d487618208cf6b) on local laptop and cluster GPU. 
---

## Checklist

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Run tests
+name: CI
 on:
   pull_request:
   push:
--- a/.github/workflows/first_party_compatibility.yml
+++ b/.github/workflows/first_party_compatibility.yml
@@ -1,0 +1,50 @@
+# Checks that our first-party packages that depend on TabPFN work correctly.
+name: First Party Compatibility
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: compatibility-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test_first_party_packages:
+    name: Test ${{ matrix.repo-name }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        repo-name:
+          - tabpfn-extensions
+          - tabpfn-time-series
+
+    steps:
+      - name: Checkout TabPFN at current ref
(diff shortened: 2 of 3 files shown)


tests/test_classifier_interface.py
```diff
@@ -492,32 +492,39 @@ def test_balance_probabilities_alters_proba_output() -> None:
...
-    tabpfn = TabPFNClassifier(fit_mode="low_memory", **kwargs)
+    tabpfn = TabPFNClassifier(fit_mode="fit_with_cache", **kwargs)
     tabpfn.fit(X, y)
-    np.testing.assert_array_almost_equal(probs, tabpfn.predict_proba(X))
-    np.testing.assert_array_equal(preds, tabpfn.predict(X))
+    np.testing.assert_array_almost_equal(probs, reference_probs, decimal=4)
+    np.testing.assert_array_almost_equal(
+        preds,
+        reference_preds,
+    )
```
oscarkey: Sorry for jumping in, but is this missing a .predict() call? And same in the test below.

magnus-priorlabs: oh yea. Good catch! I'll fix this.

magnus-priorlabs:
Tests keep failing due to a HF authentification error. I'll close this PR and open an identical new one to see if it works then.

priorphil:
So, turns out PRs from forked repos don't have access to secrets: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

PR #662, opened 2025-12-08
Switch to different token for TabPFN-2.5.

oscarkey:
This will allow public contributions to access the model weights.

The token has read access to the 2.5 hf repo, and nothing else.


oscarkey:
It's attached to my huggingface account for now.

PR #668, opened 2025-12-10
Use github actions caching for the model files when in PRs.

oscarkey:
Move the cache into the working directory so that it's consistent between OSes and accessible to hashFiles() (which is used to invalidate the cache).

The main aim is to allow PRs from forks to run the tests, as these don't have access to the HF_TOKEN and so can't download the model. This is isn't the ideal solution this problem, as the cache will be deleted [if no PRs are merged for 7 days](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy), but it's a simple one. From [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache), I'm fairly confident that branches from forks will be able to access the cache populated in main, but I guess we will find out for sure...

Other advantages are avoiding spurious failures due to HF (which we do see sometimes), although only in PRs, and speeding up the PR CI by maybe 20 seconds.

The downside is that downloading the models will not be tested in PRs, but I don't restore the cache on main to test downloading there.

Fixes RES-912
Fixes RES-911

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,8 @@ jobs:
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
-      - name: Restore model cache
+      - &restore-model-cache
+        name: Restore model cache
         # We use the cache in PRs so that PRs from forks, which can't access the
         # HF_TOKEN, have access to the model files, and also to speed them up.
         # On merge to main we don't use the cache, to test downloading.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -166,10 +208,25 @@ jobs:
             dependency-set: highest
 
     runs-on: ubuntu-22.04-4core-gpu
+    env:
+      TABPFN_MODEL_CACHE_DIR: ${{ github.workspace }}/model_cache
 
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
+      - name: Restore model cache
```
brendan-priorlabs: I think it's possible to reduce some of this duplication with YAML anchors. See https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases. There are some ways to share config across files tool, but this would be a quick and easy start

brendan-priorlabs (approved):
Nice!

.github/workflows/ci.yml
oscarkey: oooh very nice, done!

PR #670, opened 2025-12-10
Try to fix model cache.

oscarkey:
Relying on the tests to populate the cache didn't always seem to work on MacOS (sometimes the cache was empty). There was also a design flaw: we don't always use all the models when running in PRs, as some tests are skipped for speed. Thus, an entirely or partially empty cache sometimes got saved, which then broke PRs from forks.

Instead, explicitly use the "download_all_models()" function. Adapt it to skip models it already downloaded. Skip these steps entirely on main to continue to test to real download functionality.

This also doesn't fix the first-party compatibility test for tabpfn-extensions, as this seems to ignore the model cache dir, but I think that's fine for now (this check doesn't block submission).

Fixes RES-912

### DIFF
--- a/.github/workflows/first_party_compatibility.yml
+++ b/.github/workflows/first_party_compatibility.yml
@@ -53,8 +53,8 @@ jobs:
           restore-keys: |
             model-cache-
           enableCrossOsArchive: true
-        # We don't both to save the model cache in this workflow as the main CI workflow
-        # doing it is enough.
+        # We don't save the model cache in this workflow as the main CI workflow already
+        # saves it.
 
       - name: Download models from Hugging Face
         run: |


src/tabpfn/model_loading.py
```diff
@@ -385,8 +385,14 @@ def download_all_models(to: Path) -> None:
         (ModelVersion.V2_5, ModelSource.get_regressor_v2_5(), "regressor"),
     ]:
         for ckpt_name in model_source.filenames:
+            path = to / ckpt_name
```
brendan-priorlabs, line 388: `download_model` already does this (via HuggingFace), right? I'd rather not duplicate this logic

.github/workflows/first_party_compatibility.yml
```diff
@@ -35,37 +36,35 @@ jobs:
...
-        if: github.event_name == 'pull_request'
         id: restore-model-cache
         uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ${{ github.workspace }}/model_cache
           key: model-cache-${{ hashFiles('model_cache/**') }}
           restore-keys: |
             model-cache-
+          enableCrossOsArchive: true
+        # We don't both to save the model cache in this workflow as the main CI workflow
```
brendan-priorlabs: typo?

oscarkey: fixed!

src/tabpfn/model_loading.py
oscarkey: The `hf_hub_download()` seems to hit HF anyway. I'm not sure if this is a bug or to check for updates to the model file. But this fails if the HF_TOKEN isn't set. We do have a check if the file exists, but this is before `download_model()`: https://github.com/PriorLabs/TabPFN/blob/22c435ad0f291f094830166d13c75fe4a227b9a0/src/tabpfn/model_loading.py#L555

brendan-priorlabs: Gotcha. Might be worth centralizing that logic in `download_model`? Looks like it would need a new optional `download_if_not_exists` param defaulting to true, but otherwise it might be pretty simple?

oscarkey: imo it's neater to have the two checks. Otherwise we have a switching function like `download_model_if_not_exists(disable_download=)`, but I can't come up with a name that's not confusing. What do you think?

brendan-priorlabs: I don't feel too strongly. Was reacting to the duplication, but I do agree that it's minor for now. Also, I was thinking more like `download_model(..., download_if_not_exists=True)`, but the semantics for that are funky too. Feel free to merge whenever!

PR #671, opened 2025-12-11
Use the model cache even on main.

oscarkey:
I'd forgotten that cache entries are associated with a branch, and PRs only read from their branch and their base, not other PRs. Thus, we need to update the cache on main so PRs without the HF_TOKEN have a cache to read.

This means less testing of the downloading, but I think that's okay. It's fairly obvious if it's broken, and we don't change it much.


oscarkey:
That's a good idea! I'll make a PR.

PR #683, opened 2025-12-17
Very small cleanup.

oscarkey:
Part of PRI-75

PR #684, opened 2025-12-17
Split the inference engines into single and multi-device base classes.

oscarkey:
Stop using the PerDeviceModelCache for the engines that only support one device, and just use the PyTorch model directly.

This change makes things easier when adding the .to() function, as:
- Different behaviour is required for the .to() method in each case
- For fine-tuning (the batch inference engine), we need to make sure that the model we evaluate is the same as the model passed to the inference engine constructor, rather than a copy of it. This is so the correct parameters are attached to the optimiser. Removing the PerDeviceModelCache makes this more explicit/less error prone.

Part of PRI-75


oscarkey:

- #683
    - #684 ◀
        - #685

PR #685, opened 2025-12-17
Add a .to() method to the classifier and regressor interfaces.

oscarkey:
Apologies for the slightly large PR.

- Update the inference engines to maintain the device state internally (rather than being passed during fit/predict), and to move the models around as needed.
- Update `_PerDeviceModelCache` to move the models using `.to()`, rather than on-demand in `.get()`. This allows removing the locking, as `.to()` is always called from the main thread.
- Remove the `is_parallel` flag from the parallel execution code. This was used to decide whether to copy the model (when using multiple devices) or to use a single copy (when using a single device). This is no longer required, as the copying takes place during `.to()` as necessary.
- Add `.to()` to the estimators which just calls `estimator_to_device()` in base.
- Update `load_fitted_tabpfn_model()` to use the new `.to()` functions, rather than monkeying around with the inside of the estimator.
- Add tests

Part of PRI-75

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Add a `.to()` method to `TabPFNClassifier` and `TabPFNRegressor`, allowing the device to be changed after `.fit()` has been called [#685](https://github.com/PriorLabs/TabPFN/pull/685)
+- Add a `.to()` method to `TabPFNClassifier` and `TabPFNRegressor`, allowing the device to be changed after `.fit()` has been called. This change also stores the model on the GPU between `.fit()` and `.predict()` calls, use `.to("cpu")` to release this GPU memory. [#685](https://github.com/PriorLabs/TabPFN/pull/685)
 
 ### Changed
 


oscarkey:

- #683
    - #684
        - #685 ◀
            - #686
                - #687

tests/test_estimators.py
```diff
@@ -0,0 +1,203 @@
+"""Tests that cover both the classification and regression interfaces."""
```
LeoGrin, line 1: file name and docstring could be more informative, it seems all the tests are about device transfers?

LeoGrin (approved):
Looks great, thanks a lot!

tests/test_estimators.py
oscarkey: imo it's convenient for the test structure to match the main module structure, so we know where to look for tests. In this case I named it estimators, because it covers both the classifier and regressor. If I renamed it to `test_classifier_and_regressor.py`, how would that sound?

LeoGrin: hum yeah that's a good point! Yeah might be a bit better but actually `test_estimators.py` is also good actually, I thought this would be only about device transfers but now I understand it's intended to be used for much more!

oscarkey: okay I'll stick then :)

PR #692, opened 2025-12-24
Custom error for input validation

simo-prior:
### Why

TabPFN is used both as an open-source library and behind our API. Today, many invalid user inputs (bad shapes, NaNs, invalid parameter values, etc.) surface as raw `ValueError` (and sometimes other exceptions), which makes it hard for the API predictor to reliably distinguish **user-caused failures** (should be HTTP 400) from **internal/system failures** (should remain 500/503).

We want a consistent, _sklearn-friendly_ exception surface:

-   OSS users still see standard exception types (`ValueError`/`TypeError`) and existing sklearn conventions keep working.
    
-   The API layer can catch a single semantic base type for “user fault” and return a 400.
    
https://linear.app/priorlabs/issue/ENG-374/add-custom-user-data-error-in-tabpfn-package-handle-in-predictor

### Summary of changes

-   Introduce a small TabPFN exception hierarchy in `src/tabpfn/errors.py`:
    
    -   `TabPFNError`: base for all TabPFN-specific exceptions
        
    -   `TabPFNUserError`: base for errors caused by invalid user data/params (safe to map to HTTP 400)
        
    -   `TabPFNValidationError(ValueError, TabPFNUserError)`: invalid user-provided data/config values
        
    -   _(optional / follow-up)_ `TabPFNParameterError(TypeError, TabPFNUserError)` for invalid argument types
        
-   Raise `TabPFNValidationError` in a few high-signal validation sites where the failure clearly indicates bad input
        
-   In sklearn compatibility validation (`_sklearn_compat.validate_data`), re-throw **only** user-input exceptions (`ValueError`/`TypeError`) as `TabPFNValidationError`, preserving error messages via exception chaining. We intentionally do **not** wrap broad exceptions to avoid misclassifying internal bugs as user errors.
    

### Key behavior

-   Library users can continue to catch `ValueError` as before, because `TabPFNValidationError` subclasses `ValueError`.
    
-   API/predictor can catch `TabPFNUserError` (or `TabPFNValidationError`) and map these failures to HTTP 400 consistently.
    
-   Unexpected internal exceptions are not “hidden” behind validation errors.
    

### Compatibility / risk notes

-   We avoid changing sklearn semantics by keeping validation errors as `ValueError`\-compatible.
    
-   We avoid wrapping `Exception` broadly to prevent turning internal failures (e.g., `AttributeError`, `RuntimeError`) into user-facing 400s.
    

### Done when

-   Invalid user inputs that previously raised raw `ValueError` now raise `TabPFNValidationError` (still catchable as `ValueError`).
    
-   Predictor/server can reliably map `TabPFNUserError` / `TabPFNValidationError` to HTTP 400 without swallowing internal errors.

### DIFF
--- a/src/tabpfn/errors.py
+++ b/src/tabpfn/errors.py
@@ -9,9 +9,9 @@ class TabPFNError(Exception):
     """Base class for all TabPFN-specific exceptions."""
 
 
-class TabPFNValidationError(TabPFNError):
+class TabPFNUserError(TabPFNError):
     """Base class for errors caused by invalid user input (safe to map to HTTP 400)."""
 
 
-class TabPFNValidationError(ValueError, TabPFNValidationError):
+class TabPFNValidationError(ValueError, TabPFNUserError):
     """User provided invalid data (shape, NaNs, categories, etc.)."""


src/tabpfn/regressor.py
```diff
@@ -929,10 +930,12 @@ def predict(
             quantiles = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
         else:
             assert all((0 <= q <= 1) and (isinstance(q, float)) for q in quantiles), (
```
priorphil: Could you convert this to a proper Exception instead of assert?

src/tabpfn/misc/_sklearn_compat.py
```diff
@@ -415,8 +417,11 @@ def validate_data(
         https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.validate_data.html#sklearn.utils.validation.validate_data
         """
         if skip_check_array:
-            _check_n_features(_estimator, X, reset=reset)
-            _check_feature_names(_estimator, X, reset=reset)
+            try:
+               _check_n_features(_estimator, X, reset=reset)
+               _check_feature_names(_estimator, X, reset=reset)
+            except Exception as e:
```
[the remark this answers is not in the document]
safaricd: Agree with the suggestion, would rather catch slecific validation related types

src/tabpfn/errors.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+
+class TabPFNValidationError(Exception):
```
priorphil: Could you add some context to the docstring why we need this custom error and don't use a built-in one?

src/tabpfn/misc/_sklearn_compat.py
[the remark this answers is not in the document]
priorphil: What's the rationale for this change? When defining custom exceptions, I'd typically not use them for re-raising, since here we now signal any bug in an underlying library is our fault, making it harder to debug IMHO.

src/tabpfn/errors.py
noahho: Also, should this inherit from ValueError or UserError?

noahho: If you switch to TabPFNValidationError(ValueError), you keep sklearn behavior much safer (code that catches ValueError still works).

src/tabpfn/misc/_sklearn_compat.py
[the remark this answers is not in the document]
noahho: @priorphil I just added more context to the PR and changed the PR description. I agree with the suggestion and would go even further to raise a specific type of exception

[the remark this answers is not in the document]
noahho: ValueError -> TabPFNValueError which inherited from ValueError so that above code still receives same types of base errors

src/tabpfn/errors.py
noahho:
```
class TabPFNError(Exception):
    """Base class for all TabPFN-specific exceptions."""

class TabPFNUserError(TabPFNError):
    """Base class for errors caused by invalid user input (safe to map to HTTP 400)."""

class TabPFNValidationError(ValueError, TabPFNUserError):
    """User provided invalid data (shape, NaNs, categories, etc.)."""

class TabPFNParameterError(TypeError, TabPFNUserError):
    """User provided invalid parameters/types (e.g. wrong kwarg types)."""

class TabPFNConfigError(ValueError, TabPFNUserError):
    """Invalid model/config settings provided by the user."""

class TabPFNInternalError(RuntimeError, TabPFNError):
    """Unexpected internal failure (bug, invariant violation)."""
```

so that we can raise so that above code still receives same types of base errors

src/tabpfn/architectures/base/transformer.py
```diff
@@ -503,7 +504,7 @@ def forward(  # noqa: PLR0912, C901
         embedded_input = torch.cat((embedded_x, embedded_y.unsqueeze(2)), dim=2)
 
         if torch.isnan(embedded_input).any():
-            raise ValueError(
+            raise TabPFNUserError(
```
safaricd: I personally don't like the naming, `TabPFNUserError` - it's generic and has a blame emphasis. Perhaps `TabPFNInputError` or something similar? Also, with `TabPFNValidationError`, is it really needed? Because here, we raise when an input is invalid, not passing input data **validation**.

noahho: TabPFNValidationError would make perfect sense here IMO

noahho (approved):
LGTM

PR #696, opened 2025-12-29
Fix KDI transformer init signature for sklearn compatibility

noahho:
## Summary
- add explicit parameters to `KDITransformerWithNaN.__init__` to follow scikit-learn estimator requirements
- route initialization parameters appropriately when falling back to `PowerTransformer`

## Testing
- not run (not requested)

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_6952e22d70bc833399a0e4585b13d56d)

### DIFF
--- a/src/tabpfn/preprocessors/kdi_transformer.py
+++ b/src/tabpfn/preprocessors/kdi_transformer.py
@@ -50,7 +50,7 @@ class KDITransformerWithNaN(KDITransformer):
     def __init__(
         self,
         alpha: float = 1.0,
-        output_distribution: str = "normal",
+        output_distribution: str = "uniform",
         *,
         standardize: bool = True,
         copy: bool = True,
@@ -67,7 +67,6 @@ class KDITransformerWithNaN(KDITransformer):
             super().__init__(
                 alpha=alpha,
                 output_distribution=output_distribution,
-                standardize=standardize,
                 copy=copy,
             )
 
--- a/src/tabpfn/preprocessors/kdi_transformer.py
+++ b/src/tabpfn/preprocessors/kdi_transformer.py
@@ -52,15 +52,12 @@ class KDITransformerWithNaN(KDITransformer):
         standardize: bool = True,
         copy: bool = True,
     ) -> None:
-        self.alpha = alpha
-        self.output_distribution = output_distribution
-        self.standardize = standardize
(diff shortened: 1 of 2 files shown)


src/tabpfn/preprocessors/kdi_transformer.py
```diff
@@ -41,12 +41,35 @@
...
+        # base class.
+        if KDITransformer is PowerTransformer:
+            self.alpha = alpha
+            self.output_distribution = output_distribution
+            super().__init__(standardize=standardize, copy=copy)
+        else:
+            super().__init__(
+                alpha=alpha,
+                output_distribution=output_distribution,
+                standardize=standardize,
```
bejaeger: looks like `KDITransformer` doesn't have a `standardize` argument. Unfortunately, we don't test this path, could you add a test for the actual `KDITransformer`?

src/tabpfn/preprocessors/kdi_transformer.py
```diff
@@ -41,12 +41,35 @@
...
+        # backends, only pass the parameters that are supported by the active
+        # base class.
+        if KDITransformer is PowerTransformer:
+            self.alpha = alpha
+            self.output_distribution = output_distribution
+            super().__init__(standardize=standardize, copy=copy)
+        else:
+            super().__init__(
+                alpha=alpha,
+                output_distribution=output_distribution,
```
bejaeger, line 70: The default `output_distribution` in `KDITransformer` is `uniform`, here we use `normal` and overwrite it per default; should we align?

bejaeger (changes requested):
Thanks for fixing this! There are still some issues see comments.

src/tabpfn/preprocessors/kdi_transformer.py
noahho: it's now tested, however on CI will fallback to this implementation with PowerTransformer since KDI isnt installed by default (and has some heavy-ish deps)
standardize also fixed, thanks for noticing

src/tabpfn/preprocessors/kdi_transformer.py
noahho: Adapted

bejaeger (approved):
LGTM!

PR #701, opened 2025-12-30
Add finetuning wrapper for classifier

bejaeger:
## Motivation and Context

Adds a clean interface for finetuning the classifier on a single dataset. All finetuning related code is moved to a new directory `finetuning/`.

This also removes the public API function `get_preprocessed_datasets` from classifier and regressor.

## Intended followups after this PR:

- [ ] Add option to fix feature order & ordinal order
- [ ] Add finetuning wrapper for regressor
- [ ] Refactor some of the tests.


src/tabpfn/finetuning/train_util.py
```diff
@@ -0,0 +1,319 @@
...
+    from tabpfn import TabPFNClassifier, TabPFNRegressor
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(message)s",
+)
+
+logging.getLogger().setLevel(logging.INFO)
+
+
```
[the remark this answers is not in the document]
bejaeger: Will fix the logging in a follow-up PR

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -0,0 +1,702 @@
...
+    get_cosine_schedule_with_warmup,
+    save_checkpoint,
+)
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(message)s",
+)
+
+logging.getLogger().setLevel(logging.INFO)
```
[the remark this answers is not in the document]
bejaeger: logger will be handled in follow-up PR

noahho (commented):
Overall great changes! This makes things much clearer

Some concerns:
- when building the regressor code is it possible / practical to merge some code and reuse between both. I'm a bit scared it's going to be hard to maintain these two classes if they define a lot of code separately.
- Can we warn / clearly flag that regressor will be refactored as well?

bejaeger:
Thanks for the early feedback.
My idea was to get a review of the general implementation first for only the classifier (to avoid making this PR extremely big and to avoid unnecessary clean-up work). The regressor can follow right after this is in.

I agree, that it would be nice to have a common base class. I think it should be quite feasible and I'll keep it in mind when adding the regressor.

bejaeger:
Hi @noahho , this is ready for another review. I noted the intended follow-ups in the PR description.

examples/finetune_classifier.py
```diff
@@ -1,212 +1,139 @@
...
+VALIDATION_SPLIT_RATIO = 0.1
+NUM_FINETUNE_CTX_PLUS_QUERY_SAMPLES = 10_000
+FINETUNE_CTX_QUERY_SPLIT_RATIO = 0.2
+NUM_INFERENCE_SUBSAMPLE_SAMPLES = 50_000
+USE_ACTIVATION_CHECKPOINTING = True
+
+# Ensemble configuration
+NUM_ESTIMATORS_FINETUNE = 2
+NUM_ESTIMATORS_VALIDATION = 2
+NUM_ESTIMATORS_FINAL_INFERENCE = 8
```
noahho, line 53: This example script could be simpler if we use the defaults for a lot of these. If these are good parameters they should be default? I think rn the example script is quite long & complex - even if much better than before!

examples/finetune_regressor.py
```diff
@@ -8,6 +8,7 @@
 support for the Apple Silicon (MPS) backend is still under development.
 """
 
+import warnings
```
noahho: Can you add a note here that regressor will be refactored? I think rn people that see these two examples files will be confused why regressor looks so different. I think rather than in PR we want to show this to people who read the code. Maybe even remove the regressor finetuning code if it is misleading rn

noahho: I'd probably remove the file for now

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -0,0 +1,696 @@
...
+        SDPBackend.MATH,
+    ]
+    _sdpa_kernel_context = partial(sdpa_kernel, _SDPA_BACKENDS)
+except ImportError:
+    _sdpa_kernel_context = partial(  # type: ignore[misc]
+        torch.backends.cuda.sdp_kernel,
+        enable_flash=True,
+        enable_math=True,
+        enable_mem_efficient=True,
+    )
```
noahho: Would it make sense to move these to a compat.py file or similar? it will be shared also in regression and make the entry point in this file look messy, e.g. people looking at the signature of the classsifier

noahho (approved):
After my comments LGTM

examples/finetune_classifier.py
bejaeger: Hmm..yes I partially agree. I think it depends a bit on how strongly we want to communicate that the finetuning most likely needs tweaking for different datasets.

I removed a couple of these configurations and added some comments that hopefully makes it more understandable.

examples/finetune_regressor.py
bejaeger: Removed!

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: Good point! moved it to a separate file.

PR #705, opened 2025-12-31
Add regressor finetuning wrapper

bejaeger:
## Changes

Adds regressor fine-tuning. Adds common base class for finetuning classifier and regressor

## Follow-ups

- Add RPS loss as an option


src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,630 @@
...
+                    total_steps,
+                    warmup_steps,
+                )
+
+            progress_bar = tqdm(
+                finetuning_dataloader,
+                desc=f"Finetuning Epoch {epoch + 1}/{self.epochs}",
+            )
+            for batch in progress_bar:
+                optimizer.zero_grad()
```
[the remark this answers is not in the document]
bejaeger: we actually don't want that check anymore and rather fix the data preparation. the code would fail in the forward pass if we don't skip here which is what we want.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,634 @@
...
+
+                self._setup_batch(batch)
+
+                self.finetuned_estimator_.fit_from_preprocessed(
+                    batch.X_context,
+                    batch.y_context,
+                    batch.cat_indices,
+                    batch.configs,
+                )
```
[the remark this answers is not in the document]
bejaeger: this is what we want see comment above

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -171,7 +73,7 @@ def __init__(  # noqa: PLR0913
         weight_decay: float = 0.001,
         validation_split_ratio: float = 0.1,
         n_finetune_ctx_plus_query_samples: int = 20_000,
-        finetune_ctx_query_split_ratio: float = 0.02,
+        finetune_ctx_query_split_ratio: float = 0.2,
```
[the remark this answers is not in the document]
bejaeger: this is a better default.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+        n_finetune_ctx_plus_query_samples: The total number of samples per
+            meta-dataset during fine-tuning (context plus query) before applying
+            the `finetune_ctx_query_split_ratio`. Defaults to 20_000.
+        finetune_ctx_query_split_ratio: The proportion of each fine-tuning
+            meta-dataset to use as query samples for calculating the loss. The
+            remainder is used as context. Defaults to 0.2.
+        n_inference_subsample_samples: The total number of subsampled training
+            samples per estimator during validation and final inference.
+            Defaults to 50_000.
+        meta_batch_size: The number of meta-datasets to process in a single
```
anuragg1209: Should we remove this `meta_batch_size` from the parameters right now, since it is always one and users cannot control it anyway?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+        random_state: int = 0,
+        early_stopping: bool = True,
+        early_stopping_patience: int = 8,
+        min_delta: float = 1e-4,
+        grad_clip_value: float | None = 1.0,
+        use_lr_scheduler: bool = True,
+        lr_warmup_only: bool = False,
+        n_estimators_finetune: int = 2,
+        n_estimators_validation: int = 2,
+        n_estimators_final_inference: int = 8,
```
anuragg1209, line 152: Also, I was wondering about the `use_fixed_preprocessing_seeds` option. In our experiments that really helped in getting much better performance. Do we plan to add it in a subsequent PR?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+        self,
+        X: np.ndarray,
+        y: np.ndarray,
+        output_dir: Path | None = None,
+    ) -> FinetunedTabPFNBase:
+        """Internal implementation of fit that runs the finetuning loop."""
+        # Store the original training size for checkpoint naming
+        train_size = X.shape[0]
+
+        X_train, X_val, y_train, y_val = self._get_train_val_split(X, y)
```
anuragg1209: Should we set an upper limit on the val samples to avoid making per-epoch eval slower in case of big data? Like:

```
if (
            self.max_validation_samples is not None
            and len(y_val) > self.max_validation_samples
        ):
            rng = np.random.default_rng(self.random_state)
            val_indices = rng.choice(
                len(y_val), size=self.max_validation_samples, replace=False
            )
            if isinstance(X_val, pd.DataFrame):
                X_val = X_val.iloc[val_indices]
                y_val = y_val.iloc[val_indices]
            else:
                X_val = X_val[val_indices]
                y_val = y_val[val_indices]
```

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -7,159 +7,61 @@
...
-            used. Defaults to 8.
-        use_activation_checkpointing: Whether to use activation checkpointing to
-            reduce memory usage. Defaults to False.
-        save_checkpoint_interval: Number of epochs between checkpoint saves. If
-            None, no intermediate checkpoints are saved. The best model checkpoint
-            is always saved regardless of this setting. Defaults to 10.
-        **kwargs: Additional keyword arguments to pass to the underlying
-            `TabPFNClassifier`, such as `n_estimators`.
+        FinetunedTabPFNClassifier specific arguments:
+
```
noahho, line 113: To keep this sklearn compatible we unfortunately have to define all the parameters. There is no way around this unfortunately

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -181,456 +83,230 @@ def __init__(  # noqa: PLR0913
...
+        )
+        eval_classifier.fit(X_train, y_train)
+
+        try:
+            probabilities = eval_classifier.predict_proba(X_val)  # type: ignore
+            roc_auc = (
+                roc_auc_score(
+                    y_val,
+                    probabilities,
+                    multi_class="ovr",
```
noahho:
```suggestion
  if probabilities.shape[1] > 2:
      roc_auc = roc_auc_score(y_val, probabilities, multi_class="ovr")
  else:
      roc_auc = roc_auc_score(y_val, probabilities[:, 1])
```
This is easier to read; also checking probabilities.shape[1] is faster (compared to unique op) and safer (when a val batch has only one class present (which stratification should avoid but well))

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -181,456 +83,230 @@ def __init__(  # noqa: PLR0913
...
+            lr_warmup_only=lr_warmup_only,
+            n_estimators_finetune=n_estimators_finetune,
+            n_estimators_validation=n_estimators_validation,
+            n_estimators_final_inference=n_estimators_final_inference,
+            use_activation_checkpointing=use_activation_checkpointing,
+            save_checkpoint_interval=save_checkpoint_interval,
+        )
         self.classifier_kwargs = extra_classifier_kwargs or {}
+        # Store for sklearn get_params compatibility
+        self.extra_classifier_kwargs = extra_classifier_kwargs
```
noahho: Can we just rename / remove classifier_kwargs?

examples/finetune_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+# Ensemble configuration
+# number of estimators to use during finetuning
+NUM_ESTIMATORS_FINETUNE = 8
+# number of estimators to use during train time validation
+NUM_ESTIMATORS_VALIDATION = 8
+# number of estimators to use during final inference
+NUM_ESTIMATORS_FINAL_INFERENCE = 8
+
+# Reproducibility
+RANDOM_STATE = 0
```
noahho, line 54: Can we just use the defaults here as well? Why do we need to set this long list of magic numbers, makes the example much harder. Same for classifier?

examples/finetune_regressor.py
```diff
@@ -0,0 +1,134 @@
...
+# the ratio of the total dataset to be used for validation during training
+VALIDATION_SPLIT_RATIO = 0.1
+# total context split into train/test
+NUM_FINETUNE_CTX_PLUS_QUERY_SAMPLES = 10_000
+FINETUNE_CTX_QUERY_SPLIT_RATIO = 0.5
+NUM_INFERENCE_SUBSAMPLE_SAMPLES = 20_000
+# to reduce memory usage during training we can use activation checkpointing,
+# may not be necessary for small datasets
+USE_ACTIVATION_CHECKPOINTING = True
+WEIGHT_DECAY = 0.1
```
noahho: High weight decay? Usually ~0.01?

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -181,456 +83,230 @@ def __init__(  # noqa: PLR0913
...
-                    torch.cat([t.flatten() for t in y_query_batch]).unique().tolist()
-                )
-                if not qry.issubset(ctx):
-                    logger.warning(
-                        "Skipping batch: query labels %s are not a subset of "
-                        "context labels %s",
-                        qry,
-                        ctx,
-                    )
-                    continue
```
noahho, line 472: This logic was removed, also in base class. This could be problematic?!

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+        )
+        final_inference_eval_config = self._build_eval_config(
+            base_estimator_config,
+            self.n_estimators_final_inference,
+        )
+
+        if self.device.startswith("cuda") and torch.cuda.is_available():
+            eval_devices = [f"cuda:{i}" for i in range(torch.cuda.device_count())]
+        else:
+            eval_devices = ["cpu"]  # Used in tests
```
noahho: should we use our utils function infer_devices? https://github.com/PriorLabs/TabPFN/blob/c562d9b6787f3904703d9adba4aa281eb29c3909/src/tabpfn/utils.py#L186

src/tabpfn/finetuning/data_util.py
```diff
@@ -329,7 +357,7 @@ def __getitem__(self, index: int) -> tuple[Any, ...]:  # noqa: C901, PLR0912
         # it is not set as an attribute of the Regressor class
         # This however makes also sense when considering that
         # this attribute changes on every dataset
-        if regression_task:
+        if is_regression_task:
             train_mean = np.mean(y_train_raw)
             train_std = np.std(y_train_raw)
             y_test_standardized = (y_test_raw - train_mean) / train_std
```
noahho, line 376: Check for constant targets? Add eps or skip constant targets to avoid Inf?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+        """Set up the final inference model after fine-tuning completes."""
+        ...
+
+    @abstractmethod
+    def predict(self, X: np.ndarray) -> np.ndarray:
+        """Predict target values for X."""
+        ...
+
+    def fit(
+        self, X: np.ndarray, y: np.ndarray, output_dir: Path | None = None
```
noahho: Should we allow for pd.dataframe here as well and use the standard sklearn validation calls to check inputs? These are more robust and battle proven. https://github.com/PriorLabs/TabPFN/blob/c562d9b6787f3904703d9adba4aa281eb29c3909/src/tabpfn/utils.py#L485 Here we call it from the base classes - there is some additional TabPFN logic as well

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,635 @@
...
+    get_and_init_optimizer,
+    get_checkpoint_path_and_epoch_from_output_dir,
+    get_cosine_schedule_with_warmup,
+    save_checkpoint,
+)
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(message)s",
+)
```
noahho: Remove, this shouldn't be called from a library (maybe in example file or nowhere)

noahho (changes requested):
Great refactor all in all! Some suggestions made

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: Yes, this was a weird relique

examples/finetune_regressor.py
bejaeger: I think this was just me briefly testing different values for that specific dataset, which is quite easy to overfit. 0.01 works reasonable as well it seems.

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: This was done intentional after a brief discussion in Freiburg that we rather want the code to break if this is the case than skipping (to fix the actual issue in the data prepration). I think for the public version it's probably fine to have a skip here so I added it back.

src/tabpfn/finetuning/data_util.py
bejaeger: Good point! done!

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Yes, nice suggestion. Added the `validate_Xy_fit()` check

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Oops!

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Removed it.

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Yes we should add it. The problem is that the current preprocessing code makes it quite hard to add it in a clean way. The idea was to add this after the preprocessing refactor.

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Yep, good point. Instead of creating yet another input variable I added a hard cut on the max validation samples and throw a warning if it is being used. I think this is sufficient. The user can set the val size via the ratio argument.

src/tabpfn/finetuning/finetuned_base.py
anuragg1209: Sounds good! Marking this as resolved for now.

examples/finetune_regressor.py
bejaeger: removed some parameters!

examples/finetune_regressor.py
```diff
@@ -0,0 +1,112 @@
...
+from sklearn.model_selection import train_test_split
+
+from tabpfn import TabPFNRegressor
+from tabpfn.finetuning.finetuned_regressor import FinetunedTabPFNRegressor
+
+warnings.filterwarnings(
+    "ignore",
+    category=FutureWarning,
+    module=r"google\.api_core\._python_version_support",
+)
```
noahho, line 22: I think we need to move the logging here now: 
`logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")`

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,673 @@
...
+        # Store the original training size for checkpoint naming
+        train_size = X.shape[0]
+
+        inference_config = self._estimator_kwargs.get("inference_config", {})
+        base_estimator_config: dict[str, Any] = {
+            **self._estimator_kwargs,
+            "ignore_pretraining_limits": True,
+            "device": self.device,
+            "random_state": self.random_state,
+            "inference_config": inference_config,
```
noahho, line 395: Potential config aliasing: base_estimator_config takes inference_config from self._estimator_kwargs by reference (.get(..., {})). Later we modify inference_config for eval (SUBSAMPLE_SAMPLES) in _build_eval_config. Even though _build_estimator_config deep-copies base_config, it’s easy for future refactors (or non-deepcopyable values) to accidentally create mutation/aliasing back into extra_*_kwargs.
Suggestion: defensively deep-copy inference_config immediately when building base_estimator_config, e.g. inference_config = copy.deepcopy(self._estimator_kwargs.get("inference_config", {}) or {}).

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -0,0 +1,673 @@
...
+        X, y, _, _ = validate_Xy_fit(
+            X,
+            y,
+            estimator=self.finetuned_estimator_,
+            ensure_y_numeric=self._model_type == "regressor",
+            max_num_samples=-1,  # ignored if ignore_pretraining_limits is True
+            max_num_features=-1,  # ignored if ignore_pretraining_limits is True
+            ignore_pretraining_limits=True,
+        )
+
```
noahho, line 447: Reproducibility: We seed train_test_split via random_state and seed the DataLoader generator, but we don’t seed NumPy / Torch globally in the training loop. This can make runs non-deterministic.
Suggestion: set seeds explicitly in _fit() (and optionally per-epoch) e.g.:

once: np.random.seed(self.random_state) and torch.manual_seed(self.random_state) (+ torch.cuda.manual_seed_all if CUDA)

per-epoch: seed with self.random_state + epoch to match your epoch-dependent data splitting.
This will make fine-tuning runs more repeatable across executions.

noahho (approved):
LGTM after taking a look at my comments and deciding if helpful

examples/finetune_regressor.py
bejaeger: Moved it to the example scripts! Thanks

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Done!

src/tabpfn/finetuning/finetuned_base.py
bejaeger: I checked and all stochastic methods should use the seed or a generator that has been seeded. Also did a brief test and things look identical between runs. I'd therefore avoid relying on global seeds here.

bejaeger:
Thanks a lot for the thorough reviews and useful feedback! I addressed all comments.

PR #708, opened 2026-01-05
Release 7.0.0 

simo-prior:


oscarkey (approved):
Thanks! Seems like people forgot to update the changelog, so I'll send a separate PR for that.

oscarkey:
so we decided to go for 6.3.0 in the end? 🙃

PR #689, opened 2025-12-19
Use of RPS for better distributional regression

jonaslandsgesell:
## Motivation and Context

In this PR, we (Jonas Landsgesell and Pascal Knoll) implement the Ranked Probability Score (RPS) for fine-tuning TabPFN.

Ranked Probability Score (RPS) for distributional regression is a distance-sensitive scoring rule—unlike Cross Entropy (CE) loss, which is a local scoring rule. Therefore, RPS has advantages in the regression setting.

We would love to see how a TabPFN model pretrained with the RPS loss performs compared to a TabPFN model pretrained on the Cross Entropy loss. Preliminary evaluation on the regression finetuning example shows benefits of RPS over CE, when run with multiple seeds.

Below you can see our RPS vs. CE evaluation on a custom NN architecture (not TabPFN) on holdout test sets for different datasets:

Better MAE with RPS-trained NN than CE-trained NN:
image

Better R2 with RPS-trained NN than CE-trained NN:
image

Better RPS with RPS-trained NN than CE-trained NN:
image

General Literature:
* [Making and Evaluating Point Forecasts Tilmann Gneiting](https://arxiv.org/pdf/0912.0902)

* https://scoringrules.readthedocs.io/en/latest/theory.html

* [Rage against the mean](https://www.sciencedirect.com/science/article/abs/pii/S2452306221000824)

* [Gneiting, Strictly Proper Scoring Rules, Prediction, and Estimation](https://sites.stat.washington.edu/raftery/Research/PDF/Gneiting2007jasa.pdf )

## Public API Changes

-   [ x] No Public API changes
-   [ ] Yes, Public API changes (Details below)

---

## How Has This Been Tested?
Yes on the example finetuning script for regression (classification is unchanged)
---

## Checklist

-   [x ] The changes have been tested locally.
-   [ ] Documentation has been updated (if the public API or usage changes).
-   [ ] A entry has been added to `CHANGELOG.md` (if relevant for users).
-   [ ] The code follows the project's style guidelines.
-   [ ] I have considered the impact of these changes on the public API.

---

🎅🎄

PR #711, opened 2026-01-07
Add crps losses to regressor

bejaeger:
This adds Ranked Probability Score (RPS) and Ranked Logarithmic Score (RLS) to the regressor finetuning code. 

Note that this PR builds upon the original work by Jonas Landsgesell and Pascal Knoll in #689. It adapts their implementation to be compatible with our latest finetuning refactor and adds RLS.

Adding the RPS loss in the finetune example results in significantly better performance.

To find good defaults for the regression loss weights, a broader benchmarking would need to be done.


bejaeger:

- #711 ◀

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -30,41 +30,181 @@
...
+    target_cdf_BQL = (bin_indices_L.view(1, 1, -1) >= target_bins_BQ.unsqueeze(-1)).to(
+        probs_BQL.dtype
+    )
+
+    if loss_type == "rps":
+        cdf_diff_BQL = pred_cdf_BQL - target_cdf_BQL
+        cdf_term_losses_BQL = cdf_diff_BQL.square()
+    else:
+        eps = 1e-8
+        log_arg_BQL = (pred_cdf_BQL + target_cdf_BQL - 1.0).abs().clamp(min=eps)
```
noahho: RLS stability/clarity: the current -log(abs(CDF + y - 1)) form can be numerically unstable near the boundaries. Consider rewriting as the standard cumulative log score: y * -log(CDF) + (1-y) * -log(1-CDF) using log1p for stability:

eps = torch.finfo(pred_cdf_BQL.dtype).eps
cdf = pred_cdf_BQL.clamp(eps, 1 - eps)
cdf_term_losses_BQL =  target_cdf_BQL* (-torch.log(cdf)) + (1 - target_cdf_BQL) * (-torch.log1p(-cdf))

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -29,7 +29,7 @@
     from tabpfn.finetuning.data_util import ClassifierBatch
 
 
-def compute_classification_loss(
+def _compute_classification_loss(
     *,
     predictions_BLQ: torch.Tensor,
```
noahho: Minor naming nit: in the regressor you switched to logits_BQL and the suffix order is BQL; here it’s still predictions_BLQ. Can we standardize suffix ordering + naming across classifier/regressor (e.g., logits_BQL everywhere)?

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -30,41 +30,181 @@
...
+        total_loss = total_loss + rps_loss_weight * rps_loss
+
+    if rls_loss_weight > 0.0:
+        rls_loss = _ranked_probability_score_loss_from_bar_logits(
+            logits_BQL=logits_BQL,
+            targets_BQ=targets_BQ,
+            bardist_loss_fn=bardist_loss_fn,
+            loss_type="rls",
+        )
+        total_loss = total_loss + rls_loss_weight * rls_loss
```
noahho: Perf nit: _compute_regression_loss does multiple .mean() reductions and initializes total_loss via torch.tensor(0.0, device=...). Not a big deal, but we could avoid a new tensor alloc + extra reductions by accumulating per-element loss and doing a single .mean() at the end (or use logits_BQL.new_zeros(()) for the scalar).

noahho: Feel free to ignore

src/tabpfn/finetuning/finetuned_regressor.py
noahho: Feel free to ignore if you don't think it's right, I'm not sure

bejaeger: Thanks, using `log1p` seems to be more numerically stable. Changed it.

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: Good catch. I renamed to `logits` but kept the indices because we need to have the logits as second dimension for the CE loss. Wanted to avoid having 2 reshapes.

src/tabpfn/finetuning/finetuned_regressor.py
bejaeger: I kept it the way it was cause I found it to be a bit more readable and performance doesn't seem important here.

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -30,41 +30,185 @@
...
+
+    return total_loss
+
+
+def _ranked_probability_score_loss_from_bar_logits(
+    *,
+    logits_BQL: torch.Tensor,
+    targets_BQ: torch.Tensor,
+    bardist_loss_fn: Any,
+    loss_type: Literal["rps", "rls"] = "rps",
```
jonaslandsgesell:
```suggestion
    loss_type: Literal["crps", "crls"] = "crps",
```
I've updated the loss_type to CRPS and CRLS. In my previous draft, I overlooked a naming convention: when the weighting $w_k$ (the distance between bins) is explicitly included, we refer to it as the Continuous Ranked Probability Score. This is distinct from the standard RPS, which is often used for ordinal regression where the labels are ordered but the metric distance between them is undefined (e.g., treating the gap between labels 3 and 9999 the same as between 1 and 2). For our current implementation, keeping the weights is what we intend.

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -131,10 +275,21 @@ class FinetunedTabPFNRegressor(FinetunedTabPFNBase, RegressorMixin):
 
         extra_regressor_kwargs: Additional keyword arguments to pass to the
             underlying `TabPFNRegressor`, such as `n_estimators`.
+        ce_loss_weight: Weight for the bar distribution negative log-likelihood term
+            (cross-entropy-like). Defaults to 0.0.
+        rps_loss_weight: Weight for the ranked probability score (RPS) term computed
```
jonaslandsgesell:
```suggestion
        crps_loss_weight: Weight for the continous ranked probability score (CRPS) term computed
```

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -263,18 +428,23 @@ def _forward_with_loss(self, batch: RegressorBatch) -> torch.Tensor:  # type: ig
...
         )
 
-        return compute_regression_loss(
-            predictions_BLQ=predictions_BLQ,
+        return _compute_regression_loss(
+            logits_BQL=logits_BQL,
             targets_BQ=targets_BQ,
             bardist_loss_fn=bardist_loss_fn,
+            ce_loss_weight=self.ce_loss_weight,
+            rps_loss_weight=self.rps_loss_weight,
```
jonaslandsgesell:
```suggestion
            crps_loss_weight=self.crps_loss_weight,
```

src/tabpfn/finetuning/finetuned_regressor.py
bejaeger: Thanks a lot for the clarification. I went ahead and used your suggested convention and changed all rps/rls -> crps/crls.

jonaslandsgesell:
Great work Ben, thanks a lot for the integration in the current version. Really looking forward👍

PR #713, opened 2026-01-07
[Prep-refactor-1] Move preprocessing logic into different files

bejaeger:
This just separates preprocessing code into files, trying to decouple code.
There is no functionality changed or added and no renaming done.

### DIFF
--- a/src/tabpfn/preprocessing/configs.py
+++ b/src/tabpfn/preprocessing/configs.py
@@ -1,4 +1,4 @@
-"""Preprocessor and ensemble configurations."""
+"""Preprocessor and ensemble config objects."""
 
 from __future__ import annotations
 
--- a/src/tabpfn/preprocessing/ensemble.py
+++ b/src/tabpfn/preprocessing/ensemble.py
@@ -1,4 +1,4 @@
-"""Methods to define ensemble configurations."""
+"""Module for generating ensemble configurations."""
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


bejaeger:

- #713 ◀

src/tabpfn/finetuning/data_util.py
```diff
@@ -81,6 +80,54 @@ class RegressorBatch:
     y_query_raw: torch.Tensor
 
 
+@dataclass
+class BaseDatasetConfig:
```
alanprior, line 84: @bejaeger I'm unsure what exactly this is, and whether should it belong in here. It seems like this is going to be the schema as it comes from the user, like the most basic thing he might pass. Does it belong under finetuning?

bejaeger: It's only used for preparing the batches in the finetuning code. The naming is probably too general, given the fact that it's more of a helper datamodel. I'd leave it as is for now because we will likely change the definition later anyways to use the `DatasetView` abstraction. (And to conform with the PR description that no renaming was done)

bejaeger:
Thanks for the quick reviews!

src/tabpfn/finetuning/data_util.py
alanprior: Right, I guess this will be changed down the road. So fine by me for now.

Issue #718, opened 2026-01-08
process_text_na_dataframe corrupts data due to column reordering by OrdinalEncoder


shaigilat:
### Describe the bug

The function process_text_na_dataframe in tabpfn/utils.py calculates the indices of string columns based on the input DataFrame X. However, it applies the NA mask using these indices to X_encoded, which is the output of a ColumnTransformer (the ord_encoder).

Since ColumnTransformer reorders columns (typically grouping categorical features first), the indices from the input X do not match the indices in X_encoded. This results in:

Silent Data Corruption: The NA mask is applied to the wrong columns (often overwriting valid numerical data with NaN).

### Steps/Code to Reproduce

```
import pandas as pd
from tabpfn import TabPFNClassifier
url = "https://raw.githubusercontent.com/datasciencedojo/datasets/master/titanic.csv"
print(f"Loading dataset from: {url}")
df = pd.read_csv(url)

# Select a mix of features to match your requirements:
# Select a mix of features (Numeric, Categorical, some with NaNs)
features = ['Pclass', 'Sex', 'Age', 'Embarked', 'Ticket', 'Name', 'Cabin']
target = 'Survived'

X = df[features]
y = df[target]
clf = TabPFNClassifier(device="auto", n_estimators=2)
clf.fit(X, y)

print(X.head())
_, X_processed, y_processed = clf._initialize_dataset_preprocessing(X, y, rng=42)
print(X_processed[:5])
```

### Expected Results

The model should correctly handle missing values in string columns and encode the dataset without altering valid numerical data in other columns. The resulting training matrix passed to the model should contain all original numerical feature values.

<img width="799" height="425" alt="Image" src="https://github.com/user-attachments/assets/5dc3d30f-be3c-402d-bbc7-5462aff98699" />

### Actual Results

The code produces silent data corruption. It does not crash, but it overwrites valid numerical data with NaN values.

This happens because the NA mask is calculated using indices from the input DataFrame but applied to the output encoded array. Since ColumnTransformer reorders columns (moving categorical features to the front), the mask is applied to the wrong columns. Consequently, numerical features that have shifted into those indices are erroneously overwritten with NaN.

Evidence: In the Titanic dataset example (see images), the Age or Fare numerical columns (which appear at the end of the encoded array) are overwritten with NaNs because they now occupy the indices that originally belonged to string columns with missing values (like Cabin).

[...]

PR #719, opened 2026-01-08
Fix: Align NA mask with shuffled columns in process_text_na_dataframe

shaigilat:
### Issue
[process_text_na_dataframe corrupts data due to column reordering by OrdinalEncoder #718](https://github.com/PriorLabs/TabPFN/issues/718)

### Motivation and Context
This PR fixes a critical bug in process_text_na_dataframe where the NA_PLACEHOLDER mask was applied to the wrong columns in the encoded output array.

**The Problem**: When ord_encoder (a ColumnTransformer) is used, it reorders the output array: Categorical columns are moved to the front, followed by Numerical columns. The previous implementation calculated indices based on the Input dataframe X but applied the mask to the Output array X_encoded using those same indices.

**The Result**: This caused the code to overwrite the wrong columns with NaNs—specifically, it would often corrupt numerical data that had shifted position, leading to silent data corruption or downstream crashes.

**The Fix:**

1. Updated process_text_na_dataframe to correctly map the input indices to the encoded output indices.

2. Updated the function docstring to explicitly warn that the output columns are reordered (Categorical first, Numeric last).

3. Un-skipped and fixed the unit tests in tests/test_utils.py to correctly account for this column shuffling during verification.

### Public API Changes
[x] No Public API changes

[ ] Yes, Public API changes (Details below)

(Note: The function signature remains the same, but the docstring has been updated to reflect the existing column reordering behavior.)

### How Has This Been Tested?
I have verified this change with both unit tests and end-to-end integration tests:

1. Unit Tests (tests/test_utils.py):

    - Un-skipped test_process_text_na_dataframe.

    - Updated the test logic to correctly map the "Original" Ground Truth columns to the "Shuffled" Output columns before comparing them.

    - Verified that all cases (mixed types, all numeric, all categorical) now pass.

2. End-to-End Tests (tests/test_classifier_interface.py):

    - Verified that test_classifier_with_text_and_na now passes (it previously failed/crashed).

    - Ran the full test suite locally.

### Test Environment:

- OS: macOS / Linux

- Python: 3.12

- Pytest: 9.0.2

### Checklist
[x] The changes have been tested locally.

[x] Documentation has been updated (if the public API or usage changes).

[ ] A entry has been added to CHANGELOG.md (if relevant for users).

[x] The code follows the project's style guidelines.

[x] I have considered the impact of these changes on the public API.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -189,9 +189,12 @@ def test__infer_devices__mps_specified_but_torch_too_old__raises(
 
 
 # --- Test Data for the "test_process_text_na_dataframe" test ---
+# Notice that test_process_text_na_dataframe reorder the columns
 test_cases = [
     {
         # Mixed dtypes & None / pd.Na
+        # Original Order: [Ratio(N), Risk(C), Height(C), Amount(N), Type(C)]
+        # Output Order:   [Risk(C), Height(C), Type(C), Ratio(N), Amount(N)]
         "df": pd.DataFrame(
             {
                 "ratio": [0.4, 0.5, 0.6],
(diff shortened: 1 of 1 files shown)


bejaeger:
Hi @shaigilat , thanks a lot for this contribution.

We are aware of this issue and have a temporary fix that you can use. Please checkout these [`lines`](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/preprocessing/steps/preprocessing_helpers.py#L286-L298) where the issue is described: 

You can adapt the [`get_ordinal_encoder()`](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/preprocessing/steps/preprocessing_helpers.py#L298) function so it returns a [`OrderPreservingColumnTransformer`](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/preprocessing/steps/preprocessing_helpers.py#L184C7-L184C17) instead of a `ColumnTransformer`. The latter does not keep the column order which creates the issue you identified.

We will soon implement a more stable fix and integrate it into the main branch.

I'm closing this issue for now. Feel free to re-open if there are further questions.

PR #720, opened 2026-01-08
[Prep-refactor-2] Refactor processing of differential logic and group code

bejaeger:
This is another broad refactoring without any functionality change or introducing any new objects.
I think it's useful to get an overview and before starting to define better abstractions in the next step.

## Changes

- Splits paths between initialization code with differentiable input and standard inputs in `classifier.py`.
- Moves code to appropriate files -> Note that the naming here might still change and is not set in stone.
- Groups code and shares data tagging and cleaning logic between classifier and regressor.

### DIFF
--- a/changelog/720.added.md
+++ b/changelog/720.added.md
@@ -1,1 +1,4 @@
-Refactored preprocessing-related code
+- Moved preprocessing-related code to dedicated modules inside `src/tabpfn/preprocessing/`
+- Renamed public functions: 
+    - `validate_X_predict` → `ensure_compatible_predict_input_sklearn`
+    - `validate_Xy_fit` → `ensure_compatible_fit_inputs_sklearn`
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -80,8 +80,8 @@ from tabpfn.utils import (
 )
 from tabpfn.validation import (
     ensure_compatible_differentiable_inputs,
-    ensure_compatible_input_sklearn_X_predict,
-    ensure_compatible_inputs,
+    ensure_compatible_fit_inputs,
+    ensure_compatible_predict_input_sklearn,
     validate_num_classes,
 )
 
@@ -601,7 +601,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
         # TODO: Fix the types later.
         # In the following code, we have multiple conversions between DataFrames and
         # NumPy arrays. In a follow-up PR, we will fix this.
-        X, y, feature_names, n_features = ensure_compatible_inputs(
+        X, y, feature_names, n_features = ensure_compatible_fit_inputs(
             X,
             y,
             estimator=self,
@@ -994,7 +994,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
         check_is_fitted(self)
 
         if not self.differentiable_input:
-            X = ensure_compatible_input_sklearn_X_predict(X, self)
+            X = ensure_compatible_predict_input_sklearn(X, self)
             # TODO: The below steps should be handled by a "data sanitizer object"
             X = fix_dtypes(X, cat_indices=self.inferred_categorical_indices_)
             X = process_text_na_dataframe(
--- a/src/tabpfn/regressor.py
+++ b/src/tabpfn/regressor.py
@@ -74,8 +74,8 @@ from tabpfn.utils import (
     translate_probs_across_borders,
 )
 from tabpfn.validation import (
-    ensure_compatible_input_sklearn_X_predict,
-    ensure_compatible_inputs,
+    ensure_compatible_fit_inputs,
+    ensure_compatible_predict_input_sklearn,
 )
 
 if TYPE_CHECKING:
@@ -587,7 +587,7 @@ class TabPFNRegressor(RegressorMixin, BaseEstimator):
         # TODO: Fix the types later.
         # In the following code, we have multiple conversions between DataFrames and
         # NumPy arrays. In a follow-up PR, we will fix this.
-        X, y, feature_names, n_features = ensure_compatible_inputs(
+        X, y, feature_names, n_features = ensure_compatible_fit_inputs(
             X,
             y,
             estimator=self,
@@ -894,7 +894,7 @@ class TabPFNRegressor(RegressorMixin, BaseEstimator):
         check_is_fitted(self)
 
         # TODO: Move these at some point to InferenceEngine
-        X = ensure_compatible_input_sklearn_X_predict(X, self)
+        X = ensure_compatible_predict_input_sklearn(X, self)
 
         check_is_fitted(self)
 
--- a/src/tabpfn/validation.py
+++ b/src/tabpfn/validation.py
@@ -28,7 +28,7 @@ if TYPE_CHECKING:
     from tabpfn.constants import XType, YType
 
 
-def ensure_compatible_inputs(
+def ensure_compatible_fit_inputs(
     X: XType,
     y: YType,
     *,
@@ -40,7 +40,7 @@ def ensure_compatible_inputs(
     devices: tuple[torch.device, ...],
 ) -> tuple[np.ndarray, np.ndarray, list[str] | None, int]:
     """Validate and convert inputs to compatible format."""
-    X, y, feature_names_in, n_features_in = ensure_compatible_inputs_sklearn(
+    X, y, feature_names_in, n_features_in = ensure_compatible_fit_inputs_sklearn(
         X,
         y,
         estimator=estimator,
@@ -61,7 +61,7 @@ def ensure_compatible_inputs(
     return X, y, feature_names_in, n_features_in
 
 
-def ensure_compatible_input_sklearn_X_predict(
+def ensure_compatible_predict_input_sklearn(
     X: XType,
     estimator: TabPFNRegressor | TabPFNClassifier,
 ) -> np.ndarray:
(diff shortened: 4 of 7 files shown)


bejaeger:

- #713
    - #720 ◀

src/tabpfn/regressor.py
```diff
@@ -781,6 +788,7 @@ def fit(self, X: XType, y: YType) -> Self:
             return self
 
         mean, std = np.mean(y), np.std(y)
+        # TODO: y_train_std_ and y_train_mean_ don't seem to be used anywhere else.
```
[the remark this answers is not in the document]
bejaeger: the Q is if we need to define them as member variable.

src/tabpfn/validation.py
alanprior, line 1: Usually won't pick up on this, but since we are rethinking about the structure, I would question if this doesn't belong to a folder? I would leave outside only more general files, although I'm unsure what's the best practice tbh.

src/tabpfn/validation.py
```diff
@@ -0,0 +1,297 @@
...
+    if num_samples > max_num_samples:
+        raise TabPFNValidationError(
+            f"Number of samples `{num_samples:,}` in the input data is greater than "
+            f"the maximum number of samples `{max_num_samples:,}` officially supported"
+            f" by TabPFN. Set `ignore_pretraining_limits=True` to override this "
+            f"error!",
+        )
+
+
+def _check_cpu_warning_for_num_samples(
```
alanprior: This feels a bit out of place, as all this file is mainly about input validation. Would think about having some "hardware.py" or "device.py" file perhaps.

src/tabpfn/preprocessing/type_detection.py
alanprior, line 1: Small comment, but I wonder if tag is clear enough without context. Should we go for something more explicit, such as "dtype_detection.py" or perhaps "typing" or "detection"?

src/tabpfn/classifier.py
```diff
@@ -542,91 +543,106 @@ def _initialize_model_variables(self) -> tuple[int, np.random.Generator]:
         """
         return initialize_model_variables_helper(self, "classifier")
 
-    def _initialize_dataset_preprocessing(
+    def _initialize_for_differentiable_input(
```
alanprior, line 546: Could we add a one-liner explanation here what's the difference between differentiable vs standard input?

I assume that this is related to prompt-tuning which I'm not familiar of. Might be worthwhile to add some documentation or context for the uneducated user.

alanprior:
So all in all it feels like the changes are a good step towards what we want to do. It is indeed very necessary to start splitting into more logical files rather than what we had. A few general comments:

1. It's obviously a bit hard for the reviewer to make sure that nothing really breaks. I would hope that we have enough test coverage and AI agents wandering around, but obviously it will be impossible for me to catch a bad import etc.

2. I think that some functionalities are unclear or non-intuitive even for people which by now have some experience with TabPFN :) For instance, I'm not familiar with differentiable input / promptuning (?), and I believe that there are other areas. I wonder whether adding slightly more documentation explaining such cases would be relevant. In the same spirit, I encourage slightly more verbose names when possible. For instance, "tag.py" could be more explicit; this is only an example of a broader approach I believe we should consider in general, but especially now that we redesign the code: the word 'tag' will stay from here on. Do we feel it's the best name for it?

3. Structure: I'm unsure about whether some files should belong outside/inside folders, and it's a good chance to think about what would be the correct skeleton. For instance, it's great we are breaking down `utils.py`, perhaps `utils` could be a folder with many unrelated files containing more specific names, e.g. hardware and whatever else we'll have down the road?


Anyway - all of this is very subjective, and shouldn't stop us from moving fast. As long as (1) holds, feel free to go ahead.

alanprior (approved):

src/tabpfn/validation.py
```diff
@@ -0,0 +1,297 @@
...
+        max_num_samples=max_num_samples,
+        max_num_features=max_num_features,
+        ignore_pretraining_limits=ignore_pretraining_limits,
+    )
+    _check_cpu_warning_for_num_samples(
+        devices=devices,
+        num_samples=X.shape[0],
+        allow_cpu_override=ignore_pretraining_limits,
+    )
+    return X, y, feature_names_in, n_features_in
```
alanprior, line 61: @bejaeger I believe that "validate" function should receive copies of the objects, and not return them. The word "validation" reflects they are never going to change them, and if they do, then it's no longer "validation". This applies to the whole file.

src/tabpfn/validation.py
```diff
@@ -0,0 +1,297 @@
...
+import warnings
+from collections.abc import Sequence
+from typing import TYPE_CHECKING, Any
+
+import torch
+from sklearn.base import is_classifier
+from sklearn.utils.multiclass import check_classification_targets
+
+from tabpfn.errors import TabPFNValidationError
+from tabpfn.misc._sklearn_compat import check_array, validate_data
```
alanprior, line 19: Wouldn't this 'tabpfn.misc._sklearn_compat.validate_data' here belong to this file?

src/tabpfn/preprocessing/type_detection.py
bejaeger: Hmm..yeah, I think I agree. I tried to stick to one word but it's hard here. 
I would avoid "dtype", which I would leave for the actual storage type, and distinguish it to "semantic_type"; as well as "typing" which is usually reserved for types related to the language e.g. for type hints.
I changed the name to `type_detection.py`

src/tabpfn/validation.py
bejaeger: I would see the `_sklearn_compat` more as a "library middleware/wrapper". It's just needed to support older sklearn versions as well IIRC. So I'd leave it in the `compatibility` file.

src/tabpfn/validation.py
bejaeger: True! Just as context, sklearn uses the word `validate` a bit loosely, which is probably why we stuck to that naming. 
But obviously that doesn't mean we can't change our internal functions, so I went for `ensure_compatible_inputs...` wherever inputs are also converted (which is actually only the case for when we use the sklearn functions)

src/tabpfn/classifier.py
bejaeger: The class has a doc string for the differentiable input property. I would argue this is enough?

src/tabpfn/validation.py
bejaeger: I'd argue that this 'validates' the input just as much as the other samples. It checks for the size of the dataset dependingn on the provided device type. I changed the name to `_validate_num_samples_for_cpu` to be more harmonized with the other function names in that file.

src/tabpfn/validation.py
bejaeger: Yes I agree, there are a lot of things in `src/tabpfn` already. Not sure where this would fit best though so I'd keep it for now.

bejaeger:
Thanks a lot for the review @alanprior !

I did some renaming and commented on the things I kept as is.

RE your general remarks:
1. when we change the implementations we should benchmark the model on real data. Before that, I think we can live with the tests and reviews and proceed carefully with the refactor.
2. Changed some of the names and hope this is fine. However, I wouldn't go overboard with explaining specialized finetuning methods in this base class.
3. I agree for the most part. We can think about that going forward.

src/tabpfn/classifier.py
```diff
@@ -542,91 +543,107 @@ def _initialize_model_variables(self) -> tuple[int, np.random.Generator]:
...
+        # TODO(investigate): We do count the class before the
+        # label encoding. This could introduce subtle bugs if the
+        # class indices change after label encoding.
+        _, counts = np.unique(y, return_counts=True)
+        self.class_counts_ = counts
+
+        self.label_encoder_ = LabelEncoder()
+        y = self.label_encoder_.fit_transform(y)
+        self.classes_ = self.label_encoder_.classes_
+        self.n_classes_ = len(self.classes_)
```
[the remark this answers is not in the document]
bejaeger: do this later.

changelog/720.added.md
```diff
@@ -0,0 +1 @@
+Refactored preprocessing-related code
```
LeoGrin: These changelog fragments should be written for the package users. If the refactor changes some behavior for the user or fixes some bug, it should be written there. Otherwise, doesn't need a fragment (you can add the "no changelog label to the PR").

LeoGrin: Here I think we should write which public functions were renamed? (at least if they don't seem to obscure). For instance `ensure_compatible_input_sklearn_X_predict `

src/tabpfn/classifier.py
alanprior: Fair point, missed it

src/tabpfn/preprocessing/type_detection.py
alanprior: Yeah I like it

src/tabpfn/validation.py
alanprior: Oh cool, I didn't go into enough depth and thought it was validating something else. Renaming it works better.

src/tabpfn/validation.py
alanprior: Ok, interesting. And I understand the context.

My main point would be to beware of doing changes to `X, y` here, which could qualify as cleaning/detection/preprocessing, and might go against what we are aiming for. If it happens right now, we might need to question ourselves if it belongs here.

alanprior:
> Thanks a lot for the review @alanprior !
> 
> I did some renaming and commented on the things I kept as is.
> 
> RE your general remarks:
> 
> 1. when we change the implementations we should benchmark the model on real data. Before that, I think we can live with the tests and reviews and proceed carefully with the refactor.
> 2. Changed some of the names and hope this is fine. However, I wouldn't go overboard with explaining specialized finetuning methods in this base class.
> 3. I agree for the most part. We can think about that going forward.

Sounds good!

changelog/720.added.md
bejaeger: Yes, good point. I added the most important things. Also did another minor renaming.

PR #724, opened 2026-01-09
Raise exception when trying to finetune without gpu

bejaeger:


bejaeger:

- #724 ◀

bejaeger:
We could try it in the future. For small datasets it might be reasonable.

PR #728, opened 2026-01-12
Feature Modality Detector

alanprior:
This PR introduces a "Feature Modality" detector, currently disconnected from the rest of the code. This is part of a refactor to better decouple the "preprocessing" which is right now scattered across different stages of the code, and our functions should behave as an entry point.

## Issue

## Motivation and Context

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

---

## Checklist

---

### DIFF
--- a/tests/test_feature_modality_detection.py
+++ b/tests/test_feature_modality_detection.py
@@ -54,6 +54,16 @@ def _for_test_detect_with_defaults(
     )
 
 
+def _for_test_detect_modality(
+    series_data: list[Any], test_name: str, expected: FeatureModality
+) -> None:
+    s = pd.Series(series_data)
+    result = _for_test_detect_with_defaults(s)
+    if result != expected:
+        error = f"Expected {expected} but got {result} for {test_name}: {series_data}"
+        raise AssertionError(error)
+
+
 @pytest.mark.parametrize(
     ("series_data", "test_name"),
     [
@@ -71,11 +81,7 @@ def _for_test_detect_with_defaults(
     ],
 )
 def test__detect_for_constant(series_data: list[Any], test_name: str) -> None:
-    s = pd.Series(series_data)
-    result = _for_test_detect_with_defaults(s)
-    if result != FeatureModality.CONSTANT:
-        error = f"Expected CONSTANT but got {result} for {test_name}: {series_data}"
-        raise AssertionError(error)
+    return _for_test_detect_modality(series_data, test_name, FeatureModality.CONSTANT)
 
 
 @pytest.mark.parametrize(
@@ -89,11 +95,9 @@ def test__detect_for_constant(series_data: list[Any], test_name: str) -> None:
     ],
 )
 def test__detect_for_categorical(series_data: list[Any], test_name: str) -> None:
-    s = pd.Series(series_data)
-    result = _for_test_detect_with_defaults(s)
-    if result != FeatureModality.CATEGORICAL:
-        error = f"Expected CATEGORICAL but got {result} for {test_name}: {series_data}"
-        raise AssertionError(error)
+    return _for_test_detect_modality(
+        series_data, test_name, FeatureModality.CATEGORICAL
+    )
 
 
 def test__numerical_series():


src/tabpfn/constants.py
```diff
@@ -15,12 +15,16 @@
...
+from torch import Tensor
 
 TaskType: TypeAlias = Literal["multiclass", "regression"]
 TaskTypeValues: tuple[TaskType, ...] = ("multiclass", "regression")
 
 # TODO
-XType: TypeAlias = Any
+XType: TypeAlias = Union[
+    np.ndarray, DataFrame, Tensor
+]  # TODO: it makes sense to be more specific here than "Any", I guess?
```
alanprior: @bejaeger I'm unsure whether this is OK. I see this type is used, but I don't understand how/why it's helpful.

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+
+
+def _array_to_series(col: np.ndarray) -> Series:
+    # TODO (1): the last part here looks like something that should be part
+    # of the validation.
+    # TODO (2): what is the point of casting it to a series?
+    # wouldn't numpy have built in functions for this?
+    try:
+        return Series(col)
+    except TypeError as e:
```
alanprior: @bejaeger 
(1) could I assume here that this simply works? I would think that failures would occur during validation?
(2) I'm unsure why we want here a numpy array when there the code moves to pandas. I believe it will be best to assume we work with `pandas` all the way rather than numpy, especially as we'd like in the future to move to use columns names etc. WDYT?

oscarkey:
hey! I swapped the reviewers around to what I hope what was intended (the auto-assigner is very eager). But ofc add me back if you'd like my review :)

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +79,145 @@ def infer_categorical_features(
...
+    x_text = X[:, type2idx[FeatureType.TEXTUAL]]
+    return DatasetView(x_num, x_cat, x_text)
+
+
+def _get_feature_type_indices(X: np.ndarray, *, min_samples_for_inference: int, max_unique_for_category: int, min_unique_for_numerical: int, reported_categorical_indices: Sequence[int] | None = None) -> FeatureTypeIndices:
+    type2idx = defaultdict(list)
+    large_enough_x_to_infer = X.shape[0] > min_samples_for_inference
+    for idx, col in enumerate(X.T):
+        reported_categorical = idx in (reported_categorical_indices or ())
+        feat_type = _detect_feature_type(X[:, col], reported_categorical=reported_categorical, max_unique_for_category=max_unique_for_category, min_unique_for_numerical=min_unique_for_numerical, large_enough_x_to_infer_categorical=large_enough_x_to_infer)
```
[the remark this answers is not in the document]
alanarazi7: Fixed, outdated

src/tabpfn/architectures/base/encoders.py
```diff
@@ -680,6 +680,7 @@ def _transform(
         return x, nans_indicator
 
 
+# TODO: This functionality would be soon handled by the FeatureTypeDetector, so we would like to remove it in the close future.
```
bejaeger: Let's remove this comment.

src/tabpfn/constants.py
```diff
@@ -15,12 +15,16 @@
 import joblib
 import numpy as np
 from packaging import version
+from pandas import DataFrame
```
bejaeger: We typically use `import pandas as pd` and then use `pd.DataFrame`

src/tabpfn/constants.py
```diff
@@ -15,12 +15,16 @@
...
+from torch import Tensor
```
bejaeger: let's use `torch.Tensor` and import `torch` like in other files

src/tabpfn/constants.py
bejaeger: Since this is used in a lot of places, let's keep `Any` for now.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
```diff
@@ -13,6 +13,8 @@
 )
 
 
+# TODO: (1) this should be part of the FeatureTypeDetector. (2) Unclear if this differs
+# from RemoveEmptyFeaturesEncoderStep
```
bejaeger: Same here, let's remove this type of comment.

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -2,15 +2,23 @@
...
 from __future__ import annotations
 
+from collections import defaultdict
 from collections.abc import Sequence
+from dataclasses import dataclass
+from enum import Enum
 from typing import TYPE_CHECKING
 
-import pandas as pd
+from pandas import Series
```
bejaeger: same here, please `import pandas as pd` and then `pd.Series`

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+    x_cat: XType
+    x_text: XType
+
+
+class FeatureType(str, Enum):
+    """The type of a feature."""
+
+    NUMERICAL = "numerical"
+    CATEGORICAL = "categorical"
+    TEXTUAL = "textual"
```
bejaeger: should we just say `TEXT`?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+
+
+# This should inheric from FeaturePreprocessingTransformerStep-like object
+class FeatureTypeDetector:
+    """Detector for feature types."""
+
+    feature_type_indices_: FeatureTypeIndices
+
+    # TODO: fit, transform etc. the fit should be calling `detect_feature_types`
+    # and storing it. Transform should be a no-op.
```
bejaeger: Can you implement these?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+
+
+FeatureTypeIndices = dict[FeatureType, list[int]]
+
+
+# This should inheric from FeaturePreprocessingTransformerStep-like object
+class FeatureTypeDetector:
+    """Detector for feature types."""
+
+    feature_type_indices_: FeatureTypeIndices
```
bejaeger: This could be part of the `DatasetView` and then we can have one `X` as `pd.DataFrame` as soon as we load the data. Would still be my preference at this stage. Don't see the benefit of splitting early.

Also, not sure if this needs to follow the same API as the other preprocessing steps. Since this ingests the input data (before the cleaning), it could be function returning a `DatasetView` with tagged features. Then we do the cleaning and the rest of the preprocessing after. What do you think?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+
+    # TODO: fit, transform etc. the fit should be calling `detect_feature_types`
+    # and storing it. Transform should be a no-op.
+
+
+# TODO: this function should be the 'fit' function of a FeatureTypeDetector class
+# that inherits from "FeaturePreprocessingTransformerStep" or sort.
```
bejaeger: See above

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+def detect_feature_types(
+    X: np.ndarray,
```
bejaeger: `XType` ?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+        X,
+        min_samples_for_inference=min_samples_for_inference,
+        max_unique_for_category=max_unique_for_category,
+        min_unique_for_numerical=min_unique_for_numerical,
+        reported_categorical_indices=reported_categorical_indices,
+    )
+    x_num = X[:, type2idx[FeatureType.NUMERICAL]]
+    x_cat = X[:, type2idx[FeatureType.CATEGORICAL]]
+    x_text = X[:, type2idx[FeatureType.TEXTUAL]]
+    # Dropping constant features here
```
bejaeger: Should we do the `cleaning` as next step and have this funciton ONLY do `detection`?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+            The maximum number of unique values for a
+            feature to be considered categorical.
+        min_unique_for_numerical:
+            The minimum number of unique values for a
+            feature to be considered numerical.
+
+    Returns:
+        A DatasetView object with the features types.
+    """
+    type2idx = _get_feature_type_indices(
```
bejaeger: nit: I would avoid numbers in variables so would go for `feature_type_to_col_indices` or `feature_type_indices`?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+        X,
+        min_samples_for_inference=min_samples_for_inference,
+        max_unique_for_category=max_unique_for_category,
+        min_unique_for_numerical=min_unique_for_numerical,
+        reported_categorical_indices=reported_categorical_indices,
+    )
+    x_num = X[:, type2idx[FeatureType.NUMERICAL]]
+    x_cat = X[:, type2idx[FeatureType.CATEGORICAL]]
+    x_text = X[:, type2idx[FeatureType.TEXTUAL]]
```
bejaeger: let's not do the split here and return `DatasetView(X, feature_type_indices=feature_type_indices)`

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+    max_unique_for_category: int,
+    min_unique_for_numerical: int,
+    reported_categorical_indices: Sequence[int] | None = None,
+) -> FeatureTypeIndices:
+    type2idx = defaultdict(list)
+    large_enough_x_to_infer = X.shape[0] > min_samples_for_inference
+    for idx, col in enumerate(X.T):
+        reported_categorical = idx in (reported_categorical_indices or ())
+        feat_type = _detect_feature_type(
+            X[:, col],
```
bejaeger: Does that code run? Looks like there is a bug in the loop when you iterate over X.T? 
```
num_features = X.shape[1]
for idx in range(num_features):
...
```
?

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+
+def _get_feature_type_indices(
+    X: np.ndarray,
+    *,
+    min_samples_for_inference: int,
+    max_unique_for_category: int,
+    min_unique_for_numerical: int,
+    reported_categorical_indices: Sequence[int] | None = None,
+) -> FeatureTypeIndices:
+    type2idx = defaultdict(list)
```
bejaeger:
```suggestion
    feature_type_indices = defaultdict(list)
```

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+    large_enough_x_to_infer = X.shape[0] > min_samples_for_inference
```
bejaeger: I know you copied this from old logic but can we make this more descriptive. Maybe `has_enough_num_samples_to_infer_categoricals`

src/tabpfn/preprocessing/type_detection.py
bejaeger: Yes totally agree, we should convert everything to pandas as early as possible. Ideally we support torch as well, but if we create the `DatasetView` we can hide the implementation there of whatever we need.

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+        ) from e
+
+
+def _detect_categorical(
+    num_distinct: int,
+    max_unique_for_category: int,
+    min_unique_for_numerical: int,
+    *,
+    reported_categorical: bool,
+    large_enough_x_to_infer_categorical: bool,
```
bejaeger: Same comment here on variable name.

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +78,188 @@ def infer_categorical_features(
...
+        large_enough_x_to_infer_categorical=large_enough_x_to_infer_categorical,
+    ):
+        return FeatureType.CATEGORICAL
+    if _detect_textual(
+        s, num_distinct=num_distinct, max_unique_for_category=max_unique_for_category
+    ):
+        return FeatureType.TEXTUAL
+    if is_numeric_dtype(s.dtype):
+        return FeatureType.NUMERICAL
+    raise TypeError(f"Unknown feature type: {s.dtype}: {s.unique()}")
```
bejaeger: Can you add more info to the printout so when reading it's clear that the number is the unique values?

bejaeger: This could even be a `TabPFNUserError` with a description that a column had an invalid type.

src/tabpfn/constants.py
alanarazi7: Agreed it's better not to touch, but what would be the point of me calling it? Shouldn't I just ask for np.array?

src/tabpfn/preprocessing/type_detection.py
alanarazi7: Perfect, so in my opinion, this function should already get a dataframe. There should be a function before it that makes sure you have a dataframe, and if it's numpy/tensor then it just convert it. Would that make sense?

src/tabpfn/preprocessing/type_detection.py
alanprior: Sure but let's sync on the desired API as I would like to inherit from a class that you'd probably use as well

src/tabpfn/preprocessing/type_detection.py
alanprior: In my opinion, yes. It's best to decouple.

src/tabpfn/preprocessing/type_detection.py
alanprior: The thing is that this function follows the fit(), transform() API.

The fit() should probably only create a FeatureTypeColumns() object, which is simply a dictionary mapping feature types to columns names.

The transform() should then map (X, FeatureTypeColumns) -> DatasetView.

Since this should be cached / saved such as other preprocessors, if you don't follow the same API you need to make sure you somehow store this information somewhere?

src/tabpfn/preprocessing/type_detection.py
alanprior: So I changed this to DataFrame, as that it's what I expect to get, and not an "Any" object. I don't understand what XType means in this context, and which kind of hints it provides. Let's sync offline on that?

src/tabpfn/preprocessing/type_detection.py
alanprior: I answered there, let's keep the conversation. Anyway, removing the comment

src/tabpfn/preprocessing/type_detection.py
alanprior: Yeah I think it was wrong (it was before the tests), anyway changed to DataFrame

alanprior:
> Thanks a lot for this draft @alanprior I commented on a few things, mostly style related and I think there is one issue.
> 
> Could you also add tests so we know exactly what's happening here?

Yup for sure. I'm adding now many tests of series level, and then a few more on the main function (X -> DatasetView).

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +77,166 @@ def infer_categorical_features(
...
+            reported_categorical=reported_categorical,
+            max_unique_for_category=max_unique_for_category,
+            min_unique_for_numerical=min_unique_for_numerical,
+            big_enough_n_to_infer_cat=big_enough_n_to_infer_cat,
+        )
+        feature_type_to_columns[feat_type].append(col)
+    return feature_type_to_columns
+
+
+def _detect_feature_type(
```
alanprior: @LennartPurucker Please take a look and let's sync, we probably can do better than this :) but it's a starting point

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +77,176 @@ def infer_categorical_features(
...
+    X: pd.DataFrame
+    feature_type_to_columns: FeatureTypeColumns
+
+
+class FeatureType(str, Enum):
+    """The type of a feature."""
+
+    NUMERICAL = "numerical"
+    CATEGORICAL = "categorical"
+    TEXT = "text"
```
LennartPurucker: String, not text? to be in line with pandas semantics

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +77,176 @@ def infer_categorical_features(
...
+
+
+FeatureTypeColumns = dict[FeatureType, list[str]]
+
+
+# This should inheric from FeaturePreprocessingTransformerStep-like object
+class FeatureTypeDetector:
+    """Detector for feature types."""
+
+    feature_type_indices_: FeatureTypeColumns
```
LennartPurucker: Can we move away from indices and towards name-based linking? Or just using the pandas dataframe to store this information, and once the code needs the indices, it can pull them from the DF?

src/tabpfn/preprocessing/type_detection.py
LennartPurucker: An important question is, when this code is called. I remember that the old function `infer_categorical_features` was called after check_array, and thus a lot of the valuable original metadata was already lost in the sklearn checks and numpy conversions. 

One way forward could be to call this at the start before check_array and similar, and to treat all input as a dataframe afterward.

src/tabpfn/preprocessing/type_detection.py
```diff
@@ -70,3 +77,176 @@ def infer_categorical_features(
...
+    )
+
+
+def _is_numeric_pandas_series(s: pd.Series) -> bool:
+    if pd.api.types.is_numeric_dtype(s.dtype):
+        return True
+    return bool(all(_is_numeric_value(x) for x in s))
+
+
+def _detect_numeric_as_categorical(
```
LennartPurucker: I am not super sure if it is better or worse to switch numericals. I think it depends on pretraining and the data itself. I feel like we might want to benchmark this. 

I think the more important logic is to go from object dtype as input (so we do not know if string or cat) to the decision of being string or cat. Which you are kind of doing with the cardinality filter above.

src/tabpfn/preprocessing/type_detection.py
alanprior: Yes so indeed you were missing a lot of context, but you got it. It will be one of the first things we do, operating on a dataframe so we can detect based on that. In the future this might also become part of a potential logic, for instance perhaps a "free-text" field which is named "id" could be treated differently, or a "unix_time" numeric field might be detected as a date (not sure it's a good idea, but just a few examples of how columns names could also help detection).

src/tabpfn/preprocessing/type_detection.py
alanprior: Hmm, so I wouldn't use `string` as if it's ["dog", "cat", "parrot"] it would be categorical. The "text" is probably something that resembles more your "sentence" terminology. I feel de-ja-vu from the discussion :) I'm open to change it but I think string is not a good choice, because some strings are categorical.

src/tabpfn/preprocessing/type_detection.py
alanprior: YES!

This part is actually a leftover, I already moved to column names :)

src/tabpfn/preprocessing/type_detection.py
alanprior: Right, so `_detect_numeric_as_categorical` is right now backwards compatible. As this is a public repo refactor I wouldn't challenge this assumption just yet :) Agreed we might change things down the road

src/tabpfn/preprocessing/type_detection.py
bejaeger: I'd decouple storage dtype (string) from semantic type here. What we detect here should be the semantic types.

src/tabpfn/preprocessing/type_detection.py
bejaeger: Agreeing here, let's not change any functionality with this PR. We can move code around after.

src/tabpfn/preprocessing/type_detection.py
alanprior: Right, so I think that I'm doing that. Right now a text is a string column that has more unique values than "max_unique_for_category"

tests/test_feature_modality_detection.py
```diff
@@ -0,0 +1,240 @@
...
+def test__detect_for_series_with_nan_and_floats():
+    s = pd.Series([np.nan, 1.0, np.nan, 1.0])
+    result = _for_test_detect_with_defaults(s)
+    assert result == FeatureType.CATEGORICAL
+
+
+def test__detect_for_series_with_few_null_types():
+    s = pd.Series([np.nan, None, np.nan, None])
+    result = _for_test_detect_with_defaults(s)
+    assert result == FeatureType.CONSTANT
```
bejaeger: For some of these it would be easier if you used `@pytest.mark.parametrize`, e.g. for the ones testing for CONSTANT?

src/tabpfn/preprocessing/type_detection.py
LennartPurucker: > string column that has 

String and object column? Or are all object columns first transformed to string?

bejaeger: That's curretly happening in the `fix_dtypes` function. Needs to come before the feature detection when we plug this in.

src/tabpfn/preprocessing/type_detection.py
bejaeger: Let's do that in a follow-up and when we plug in this feature detection

src/tabpfn/preprocessing/type_detection.py
LennartPurucker: > I'd decouple storage dtype (string) from semantic type here. What we detect here should be the semantic types.

I am not super sure we are doing semantic types here. Note, I do not care too much about the name as string==text. But for the sake of understanding the problem:

Numerical and categorical are both storage dtypes. Our check for text is also semantic-unaware and mostly checks how we want to treat this column in future encodings. 

Semantic types of a string would be more like one of (number + units, entity names, sentences). 
Similarly, you can map numbers and categoricals to more semantic types. 

But for us, we are mostly checking right now how we want to store this column, such that subsequent pipeline steps work with the storage dtype to extract different semantics. 

Rel. categoricals with string names: the categorical pipeline would likely extract different semantics from categoricals with string names than the string pipeline from a "random collection" of strings

bejaeger: Thanks for opening the discussion. I think "semantic" is not a good word here, so I agree with what you said. And also that categoricals and numericals could be seen as storage dtypes in pandas world.

I'm arguing for some hierarchy between the incoming dtype (a string could be numerical, categorical, or actually text) and decision how we want to treat it (maybe we call this modality?).

LennartPurucker: Modality is a good term!

alanprior: If it's helpful, we could rename everything to be "modality detection" instead of "type detection". I think it will make things much more clear. It seems you both agree, so I'll go ahead and change accordingly.

alanprior: (Also, I really like the term for this)

PR #729, opened 2026-01-12
[Prep-refactor-3] Introduce ensemble preprocessor class

bejaeger:
- Introduces `TabPFNEnsemblePreprocesor` class that owns the preprocessing pipeline
- Adds `TabPFNPreprocessedEnsembleMember` that knows everything about an ensemble member
- Some miscellaneous other improvements

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -9,7 +9,7 @@ from collections.abc import Iterator, Sequence
 from copy import deepcopy
 from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING
 from typing_extensions import override
 
 import joblib
@@ -18,6 +18,7 @@ import torch
 
 from tabpfn.architectures.base.memory import (
     DEFAULT_SAVE_PEAK_MEMORY_FACTOR,
+    MemorySavingMode,
     should_save_peak_mem,
 )
 from tabpfn.parallel_execute import parallel_execute
@@ -63,7 +64,7 @@ class InferenceEngine(ABC):
     def __init__(
         self,
         *,
-        save_peak_mem: bool | Literal["auto"] | float | int,
+        save_peak_mem: MemorySavingMode,
         dtype_byte_size: int,
         force_inference_dtype: torch.dtype | None,
     ) -> None:
@@ -76,7 +77,7 @@ class InferenceEngine(ABC):
                 dtype. Otherwise, the default dtype will be used.
         """
         super().__init__()
-        self.save_peak_mem: bool | Literal["auto"] | float | int = save_peak_mem
+        self.save_peak_mem = save_peak_mem
         self.dtype_byte_size = dtype_byte_size
         self.force_inference_dtype = force_inference_dtype
 
@@ -197,7 +198,7 @@ class SingleDeviceInferenceEngine(InferenceEngine):
         self,
         *,
         models: list[Architecture],
-        save_peak_mem: bool | Literal["auto"] | float | int,
+        save_peak_mem: MemorySavingMode,
         dtype_byte_size: int,
         force_inference_dtype: torch.dtype | None,
     ) -> None:
@@ -235,7 +236,7 @@ class MultiDeviceInferenceEngine(InferenceEngine):
         self,
         *,
         model_caches: list[_PerDeviceModelCache],
-        save_peak_mem: bool | Literal["auto"] | float | int,
+        save_peak_mem: MemorySavingMode,
         dtype_byte_size: int,
         force_inference_dtype: torch.dtype | None,
     ) -> None:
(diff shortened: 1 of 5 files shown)


bejaeger:

- #713
    - #720
        - #726
            - #729 ◀

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -15,20 +16,135 @@
...
+        """
+        super().__init__()
+        self.configs = configs
+        self.rng = rng
+        self.n_preprocessing_jobs = n_preprocessing_jobs
+
+        # TODO:
+        # 1. Create pipeline in init for balanced feature subsampling
+        # 2. Run pipeline.num_added_features() for each ensemble member
+        # 3. Create feature slices
```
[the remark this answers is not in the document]
bejaeger: Will be handled later

src/tabpfn/regressor.py
```diff
@@ -442,7 +443,9 @@ class in Fine-Tuning. The fit_from_preprocessed() function sets this
             inference_precision
         )
         self.fit_mode: Literal["low_memory", "fit_preprocessors", "batched"] = fit_mode
-        self.memory_saving_mode = memory_saving_mode
+        self.memory_saving_mode: bool | Literal["auto"] | float | int = (
+            memory_saving_mode
+        )
```
oscarkey: Do we need the type here as it's defined in the function parameters? If we do, can use `MemorySavingMode`?

src/tabpfn/regressor.py
```diff
@@ -796,17 +794,20 @@ def fit(self, X: XType, y: YType) -> Self:
             self.znorm_space_bardist_.borders * self.y_train_std_ + self.y_train_mean_,
         ).float()
 
-        # Create the inference engine
+        ensemble_preprocessor = TabPFNEnsemblePreprocessor(
+            configs=ensemble_configs,
+            rng=rng or np.random.default_rng(),
```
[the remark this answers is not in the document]
oscarkey: seems correct?

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -15,20 +16,129 @@
...
+
+    def get_static_seed(self) -> int:
+        """Get the static seed for the ensemble data processor."""
+        return self.rng.integers(0, int(np.iinfo(np.int32).max))
```
oscarkey, line 94: I'm worried it's confusing if you get a different seed every time.
- We could rename it to `next_static_seed()`?
- Or we could cache it, but maybe that's more confusing. I'm guessing we can't sample it once in __init__() because this changes the order? Gosh our rng handling is a mess haha.

src/tabpfn/inference.py
```diff
@@ -74,7 +76,7 @@ def __init__(
                 dtype. Otherwise, the default dtype will be used.
         """
         super().__init__()
-        self.save_peak_mem = save_peak_mem
+        self.save_peak_mem: bool | Literal["auto"] | float | int = save_peak_mem
```
oscarkey: do we need the type here?

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -15,20 +16,129 @@
...
+        self.rng = rng
+        self.n_preprocessing_jobs = n_preprocessing_jobs
+
+        # TODO:
+        # 1. Create pipeline in init for balanced feature subsampling
+        # 2. Run pipeline.num_added_features() for each ensemble member
+        # 3. Create feature slices
+
+    def get_static_seed(self) -> int:
+        """Get the static seed for the ensemble data processor."""
```
oscarkey: Might be worth mentioning what the "static seed" is

oscarkey (approved):
looks great to me! Thank you so much for cleaning it up.

src/tabpfn/regressor.py
bejaeger: Hmm...yeah the type checker complains if that is not set (I think because of the Literal?).  
I think it's anyway time for a `types.py` file.

src/tabpfn/preprocessing/ensemble.py
bejaeger: Yeah this is very specialized and only used in `InferenceEngineOnDemand`. I changed it to `next_static_seed` and added a bit of a longer comment. And yes haha.. hopefully the rnd handling becomes cleaner after the overhaul.

src/tabpfn/inference.py
bejaeger: As mentioned above, somehow pyright is failing here?
I changed it to `MemorySavingMode`

bejaeger: And maybe we should check these when we create a `types.py`

oscarkey: sgtm!

PR #730, opened 2026-01-13
Finetuning API Improvements

LennartPurucker:
## Motivation and Context

This PR adds new logic to the finetuning code to make it easier to use within existing pipelines.

Changes: 
* add: allows users to select the primary evaluation metric for validation 
* add: allows users to set a time limit for fitting 
* add: allow users to pass their own validation data X_val, y_val
* fix: avoid test split crash for tiny data
* add: evaluate the default checkpoint and use it as a baseline for the performance improvements of finetuning (i.e., fallback to the default if finetuning does not improve performance instead of the "worse" checkpoint after the first epoch)

---

## Public API Changes

-   [ ] No Public API changes
-   [x] Yes, Public API changes

See above: all changes are backward-compatible, except for evaluating the default checkpoint, which I made mandatory.  

---

## How Has This Been Tested?

It ran on TabArena without any crashes. 

---

## Checklist

---

### DIFF
--- a/src/tabpfn/finetuning/finetuned_base.py
+++ b/src/tabpfn/finetuning/finetuned_base.py
@@ -356,21 +356,12 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
             stratify=y if self._model_type == "classifier" else None,
         )
 
+    @abstractmethod
     def _get_valid_finetuning_query_size(
-        self, *, n_finetune_ctx_plus_query_samples: int, y_train: np.ndarray | None
+        self, *, query_size: int, y_train: np.ndarray | None
     ) -> int:
         """Calculate a valid finetuning query size."""
-        # finetuning_query_size should be greater or equal to the number of classes
-        finetuning_query_size = int(
-            n_finetune_ctx_plus_query_samples * self.finetune_ctx_query_split_ratio
-        )
-        if self._model_type == "classifier":
-            assert y_train is not None, (
-                "y_train required to compute finetuning query size for classification."
-            )
-            n_classes = len(np.unique(y_train))
-            finetuning_query_size = max(finetuning_query_size, n_classes)
-        return finetuning_query_size
+        ...
 
     def fit(
         self,
@@ -533,7 +524,9 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
         scheduler: LambdaLR | None = None
 
         finetuning_query_size = self._get_valid_finetuning_query_size(
-            n_finetune_ctx_plus_query_samples=n_finetune_ctx_plus_query_samples,
+            query_size=int(
+                n_finetune_ctx_plus_query_samples * self.finetune_ctx_query_split_ratio
+            ),
             y_train=y_train,
         )
         for epoch in range(epoch_to_start_from, self.epochs):
--- a/src/tabpfn/finetuning/finetuned_classifier.py
+++ b/src/tabpfn/finetuning/finetuned_classifier.py
@@ -331,6 +331,18 @@ class FinetunedTabPFNClassifier(FinetunedTabPFNBase, ClassifierMixin):
         }
 
     @override
+    def _get_valid_finetuning_query_size(
+        self, *, query_size: int, y_train: np.ndarray
+    ) -> int:
+        """Calculate a valid finetuning query size."""
+        # finetuning_query_size should be greater or equal to the number of classes
+        assert y_train is not None, (
+            "y_train required to compute finetuning query size for classification."
+        )
+        n_classes = len(np.unique(y_train))
+        return max(query_size, n_classes)
+
+    @override
     def _log_epoch_evaluation(
         self, epoch: int, eval_result: EvalResult, mean_train_loss: float | None
     ) -> None:
--- a/src/tabpfn/finetuning/finetuned_regressor.py
+++ b/src/tabpfn/finetuning/finetuned_regressor.py
@@ -503,6 +503,16 @@ class FinetunedTabPFNRegressor(FinetunedTabPFNBase, RegressorMixin):
         return {"mse": eval_result.primary}
 
     @override
+    def _get_valid_finetuning_query_size(
+        self,
+        *,
+        query_size: int,
+        **kwargs: Any,
+    ) -> int:
+        """Calculate a valid finetuning query size."""
+        return query_size
+
+    @override
     def _log_epoch_evaluation(
         self, epoch: int, eval_result: EvalResult, mean_train_loss: float | None
     ) -> None:


changelog/730.added.md
```diff
@@ -0,0 +1,3 @@
+- Add new features to finetuning (metric selection, time limit, passing validation data)
```
bejaeger, line 1: Should we elaborate here and specify exactly the arguments? `time_limit`, `eval_metric` during `init` and `X_val`, `y_val` during `fit()`.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -445,8 +465,16 @@ def _fit(  # noqa: C901,PLR0912
 
         self.X_ = X
         self.y_ = y
-
-        X_train, X_val, y_train, y_val = self._get_train_val_split(X, y)
+        if X_val is not None and y_val is not None:
+            X_train, y_train = X, y
+            X_val, y_val, _, _ = ensure_compatible_fit_inputs_sklearn(
```
bejaeger, line 477: Could we make that `if/else` part of the `_get_train_val_split` and parse the val data there? Would read cleaner.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -471,14 +499,30 @@ def _fit(  # noqa: C901,PLR0912
...
+        best_metric: float = eval_result.primary
 
-        best_metric: float = self._get_initial_best_metric()
+        logger.info("--- 🚀 Starting Fine-tuning ---")
         patience_counter = 0
         best_model = None
 
         scheduler: LambdaLR | None = None
 
+        max_data_size = n_finetune_ctx_plus_query_samples
```
bejaeger: is this rename necessary?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -471,14 +499,30 @@ def _fit(  # noqa: C901,PLR0912
...
+        # finetuning_query_size should be greater or equal to the number of classes
+        finetuning_query_size = int(max_data_size * self.finetune_ctx_query_split_ratio)
```
bejaeger: In the attempt to keep the train loop clean, what do you think of adding a method to the classifier and regressor implementation that just does:
`_get_valid_finetuning_query_size()`
Can then also add the comment above as docstring.
You can call it directly in the `partial` below.

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -139,10 +145,12 @@ def __init__(  # noqa: PLR0913
         use_activation_checkpointing: bool = True,
         save_checkpoint_interval: int | None = 10,
         extra_classifier_kwargs: dict[str, Any] | None = None,
+        eval_metric: Literal["roc_auc", "log_loss"] | None = None,
```
bejaeger, line 148: Should we have `roc_auc` as default? right now it looks like the code would throw a value error if this is not specified?

src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -340,21 +359,31 @@ def _setup_inference_model(
...
             output_dir: Directory path for saving checkpoints. If None, no
                 checkpointing is performed and progress will be lost if
                 training is interrupted.
 
         Returns:
             The fitted instance itself.
         """
-        super().fit(X, y, output_dir)
+        if self.eval_metric is None:
+            self.eval_metric = "roc_auc"
```
bejaeger, line 396: Okay, here you set a default. Why not in init()?

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -518,21 +529,31 @@ def _setup_inference_model(
...
             output_dir: Directory path for saving checkpoints. If None, no
                 checkpointing is performed and progress will be lost if
                 training is interrupted.
 
         Returns:
             The fitted instance itself.
         """
-        super().fit(X, y, output_dir)
+        if self.eval_metric is None:
+            self.eval_metric = "mse"
```
bejaeger, line 564: same here, why not set this in init?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -662,6 +706,23 @@ def _fit(  # noqa: C901,PLR0912
...
+                    )
+                    break
+
+                n_epochs_run = epoch + 1 - epoch_to_start_from
+                if elapsed_time + (elapsed_time / n_epochs_run) > self.time_limit:
+                    logger.info(
+                        "🛑 Not enough time remaining for another epoch. Stopping "
+                        "training.",
+                    )
+                    break
```
bejaeger, line 730: nit: all the above could also be a `if _is_time_limit_reached(start_time): break` call which would maybe read a bit nicer. I'll leave it up to you

src/tabpfn/finetuning/finetuned_base.py
LennartPurucker: Mhm, not sure, I think it is cleaner to pass data to a function that does not use it? Currently, `_get_train_val_split` is "cleaner" for its own purpose of splitting data.

changelog/730.added.md
LennartPurucker: Done

src/tabpfn/finetuning/finetuned_base.py
LennartPurucker: No, removed

src/tabpfn/finetuning/finetuned_base.py
LennartPurucker: Added, but moved it before the loop to avoid recomputation given that its input does not change across iterations.

src/tabpfn/finetuning/finetuned_classifier.py
LennartPurucker: I use this desgin patterns for these kinds of defaults for two reasons: (1) they allow us to adapt the default based on other hyperparameters, (2) they allow a pipeline to pass None if it does not know if the base model is a classifier or regressor.

(1) is similar in design to an "auto" literal. 
(2) avoids a lot of if-else statements based on the base model type that are quite unnecessary.

src/tabpfn/finetuning/finetuned_base.py
LennartPurucker: I think we want to communicate to users that the break happened because of us thinking we wont have enough time.

The example that often confuses users is that they set a 4-hour time limit, but we break after 3 hours because one epoch took 3 hours. So they are confused why we do not run the full 4 hours. 

But otherwise agree we could merge this somehow. But I feel the readability of the new merged if-else-log statement is worse than two breaks.

src/tabpfn/finetuning/finetuned_base.py
bejaeger: I'd say yes with a clear doc string. But not a biggy, we can leave it as is.

src/tabpfn/finetuning/finetuned_base.py
bejaeger: more meant a simple refactor and take all of it into a separate function. But I see that it would require more arguments that I missed above. Happy to leave as is.

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: Thanks makes sense!

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -348,14 +356,37 @@ def _get_train_val_split(
             stratify=y if self._model_type == "classifier" else None,
         )
 
+    def _get_valid_finetuning_query_size(
+        self, *, n_finetune_ctx_plus_query_samples: int, y_train: np.ndarray | None
+    ) -> int:
```
bejaeger, line 362: Sorry, maybe I wasn't clear before. Can you make this abstract here and then implement it in the classifier and regressor, respectively? I'd avoid many model type checks in the base class.

LennartPurucker: Jup will do. I understood it the first time already, but I did not see the need. 

Can you give me a short why? Or rather, what is your design philosophy for this? 
I am going to change it, so this is not a discussion but rather me trying to get a gist of your preferences. 

The subclassing is quite a lot of code overhead and duplication (mostly in class definitions and parameters) for some parts of the logic. I think I prefer more functional code than class-based-inheritance on average. I kind of got the feeling we are planning to support a lot of different subclasses in the future, although I do not see more than two any time soon.

LennartPurucker: Added a minimal version by adding the duplicated code into the function call in the base class.

bejaeger: Yep, there shouldn't be any code duplication. The definitions and parameters I'm less worried about cause they're are under IDE control.

This might have been overkill for this small function and I would have been fine with the `if/else` but just to give you my general sense why I think it's useful:

- Code readability: You can read the code in the base class fluently and know exactly what it does "logically" without going through "if else" cognitive load
- Testability
- Reducing number of comments. If we have it subclassed, we don't need to argue why we do things different for some type, but can put it in the doc string which is much cleaner. The doc string can be read when hovering so doesn't clutter the code.
- Extensibility: I could see us using this as a playground for testing other tasks in the future, but this is probs a minor point as you say.

bejaeger: Thanks for changing it!
Let's get it merged

PR #733, opened 2026-01-14
[Prep-refactor 4] Model encoder refactor

bejaeger:
Another PR to make the code more manageable with no functionality changes.

This is necessary to start to disentangle "encoder" and "preprocessor" inside the model and harmonize torch-based preprocessing with numpy-based preprocessing

This mainly performs a refactoring of the 'encoder' inside the model (breaking up the previous `encoders.py` file into logical parts, moving `architectures.base.encoders.py` to `architectures.encoders` module), and some small refactoring in the `preprocessing/` folder. 
Changes to the implementation were only done to make ruff happy.

### DIFF
--- a/src/tabpfn/architectures/encoders/steps/__init__.py
+++ b/src/tabpfn/architectures/encoders/steps/__init__.py
@@ -15,7 +15,6 @@ __all__ = [
     "CategoricalInputEncoderPerFeatureEncoderStep",
     "FrequencyFeatureEncoderStep",
     "InputNormalizationEncoderStep",
-    "InputNormalizationEncoderStep",
     "MulticlassClassificationTargetEncoderStep",
     "NanHandlingEncoderStep",
     "RemoveDuplicateFeaturesEncoderStep",
--- a/src/tabpfn/architectures/encoders/steps/input_normalization_encoder_step.py
+++ b/src/tabpfn/architectures/encoders/steps/input_normalization_encoder_step.py
@@ -86,7 +86,7 @@ class InputNormalizationEncoderStep(SeqEncStep):
 
         if self.normalize_x:
             (
-                x,
+                _,
                 (
                     self.mean_for_normalization,
                     self.std_for_normalization,
--- a/tests/test_encoders/test_encoders.py
+++ b/tests/test_encoders/test_encoders.py
@@ -57,7 +57,7 @@ def test_input_normalization():
     x[:, 2, 6:] = 100.0
     out = encoder({"main": x}, single_eval_pos=5)["main"]
     assert (out[:, 0, :] == out_ref[:, 0, :]).all(), (
-        "Changing one batch should not affeect the others."
+        "Changing one batch should not affect the others."
     )
     assert (out[:, 2, 0:5] == out_ref[:, 2, 0:5]).all(), (
         "Changing unnormalized part of the batch should not affect the others."
--- a/tests/test_encoders/test_projections.py
+++ b/tests/test_encoders/test_projections.py
@@ -123,7 +123,7 @@ def test_combination():
     x[6:, 2, 2] = 100.0
     out = encoder({"main": x}, single_eval_pos=5)["main"]
     assert (out[:, 0, :] == out_ref[:, 0, :]).all(), (
-        "Changing one batch should not affeect the others."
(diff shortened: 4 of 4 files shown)


bejaeger:

- #733 ◀

src/tabpfn/architectures/encoders/pipeline_interfaces.py
```diff
@@ -0,0 +1,199 @@
...
+            *args: A list of SeqEncStep instances to apply in order.
+            output_key:
+                The key to use for the output of the encoder in the state dict.
+                Defaults to "output", i.e. `state["output"]` will be returned.
+            **kwargs: Additional keyword arguments passed to `nn.Sequential`.
+        """
+        super().__init__(*args, **kwargs)
+        self.output_key = output_key
+
+    def forward(self, input: dict, **kwargs: Any) -> torch.Tensor:
```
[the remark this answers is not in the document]
bejaeger: Fix later when doing functional changes

src/tabpfn/preprocessing/pipeline_interfaces.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Args:
+            X: 2d array of shape (n_samples, n_features).
+        """
+        # TODO: Get rid of this, it's always test in `transform`
+        result = self._transform(X, is_test=True)
+        return TransformResult(result, self.categorical_features_after_transform_)
+
+
+class SequentialFeatureTransformer(UserList):
```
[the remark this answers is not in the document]
bejaeger: Fix later when doing functional changes

changelog/733.added.md
```diff
@@ -0,0 +1 @@
+Refactor of model encoder. Move imports from `tabpfn.architectures.base.encoders` to `tabpfn.architectures.encoders`
```
alanprior, line 1: @bejaeger I'm wondering if we should start deprecating the word "encoder". So far encoders were a vague word, but I think we are converging into processors and embedders, and perhaps this means we should start treating the new code accordingly?

src/tabpfn/architectures/encoders/steps/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+from .multiclass_classification_target_encoder_step import (
+    MulticlassClassificationTargetEncoderStep,
+)
+from .nan_handling_encoder_step import NanHandlingEncoderStep
+from .remove_duplicate_features_encoder_step import RemoveDuplicateFeaturesEncoderStep
+from .remove_empty_features_encoder_step import RemoveEmptyFeaturesEncoderStep
+from .variable_num_features_encoder_step import VariableNumFeaturesEncoderStep
+
+__all__ = [
+    "CategoricalInputEncoderPerFeatureEncoderStep",
```
alanprior, line 15: @bejaeger As mentioned above, I don't think that their correct naming should be encoders, but rather pre-processing steps. At least for some of them. As such, I wouldn't even locate them inside `"tabpfn/architectures"`, they belong IMHO to a different folder potentially called `tabpfn/preprocessing`. We might agree or disagree on that and it might not sound critical, but I believe that these PRs should really try to restructure the repo in a way that aligns with how we envision the different building blocks (even if their internal naming or functionality might be a bit wrong for now).

bejaeger: Yes, agreed! As per slack discussion, let's leave the torch-related parts here for now. And potentially move them over later.

changelog/733.added.md
bejaeger: I think "encoding" is still a part of the process that is different from preprocessing and embedding. E.g. encoding whether a value is nan or not nan. It's not processing anything, and also not embedding to a hidden space, but just adding to the simple cell-level representation, so I think encoding would be appropriate here. 
In general I agree though, we'll have to rename the processing steps.

alanprior: It's semantics at this point, but I see where you are coming from. I definitely don't think it's `embedding`. The word `encoding` on its own might not be bad but it's a bit overloaded at the moment. I'm also not sure which other operations we have in the same spirit - feature grouping for instance? Tokenization?

Probably there is a subtle line and I didn't find a better word, so let's continue with encoding to describe those operations!

PR #738, opened 2026-01-19
fix model download race condition

### LINKED ISSUE #370: FileNotFoundError: Race Condition when TabPFN is loaded across for the first time in multiple jobs
### Describe the bug

If TabPFN's checkpoints are initially loaded across multiple parallel jobs, the jobs overwrite each others checkpoint files, causing TabPFN to fail in all jobs except one.

### Steps/Code to Reproduce

```python
from sklearn.model_selection import cross_val_score
from sklearn.datasets import load_breast_cancer

X, y = load_breast_cancer(return_X_y=True)

model = TabPFNClassifier()

# TabPFN has never been loaded before
# now, 5 parallel jobs are looking for the checkpoints
print(cross_val_score(model, X, y, cv=5, n_jobs=5))
```

### Expected Results

Checkpoints are loaded once, and script prints out cross-validation score over 5 different folds.

### Actual Results

For context: script was executed in a (docker-rootless) dev container, where I am ~~g~~root.
It seems that each job individually tries to load the tabpfn checkpoints.
I assume that only the job finishing last succeeds on the fit, due to overwriting the checkpoints for all other jobs:

```sh
(dev) root@39f78d7b4313:/workspace# python tabpfn_bug.py 
/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py:89: UserWarning: Downloading model to /root/.cache/tabpfn/tabpfn-v2-classifier.ckpt.
  model, _, config_ = load_model_criterion_config(
/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py:89: UserWarning: Downloading model to /root/.cache/tabpfn/tabpfn-v2-classifier.ckpt.
  model, _, config_ = load_model_criterion_config(
/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py:89: UserWarning: Downloading model to /root/.cache/tabpfn/tabpfn-v2-classifier.ckpt.
  model, _, config_ = load_model_criterion_config(
/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py:89: UserWarning: Downloading model to /root/.cache/tabpfn/tabpfn-v2-classifier.ckpt.
  model, _, config_ = load_model_criterion_config(
/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py:89: UserWarning: Downloading model to /root/.cache/tabpfn/tabpfn-v2-classifier.ckpt.
  model, _, config_ = load_model_criterion_config(
tabpfn-v2-classifier.ckpt: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0M/29.0M [00:00<00:00, 190MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 37.0/37.0 [00:00<00:00, 350kB/s]
tabpfn-v2-classifier.ckpt: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0M/29.0M [00:00<00:00, 198MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 37.0/37.0 [00:00<00:00, 466kB/s]
tabpfn-v2-classifier.ckpt: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0M/29.0M [00:00<00:00, 191MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 37.0/37.0 [00:00<00:00, 473kB/s]
tabpfn-v2-classifier.ckpt: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0M/29.0M [00:00<00:00, 161MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 37.0/37.0 [00:00<00:00, 411kB/s]
tabpfn-v2-classifier.ckpt: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0M/29.0M [00:00<00:00, 157MB/s]
config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 37.0/37.0 [00:00<00:00, 403kB/s]
/opt/micromamba/envs/dev/lib/python3.12/site-packages/sklearn/model_selection/_validation.py:516: FitFailedWarning: 
4 fits failed out of a total of 5.
The score on these train-test partitions for these parameters will be set to nan.
If these failures are not expected, you can try to debug them by setting error_score='raise'.

Below are more details about the failures:
--------------------------------------------------------------------------------
4 fits failed with the following error:
Traceback (most recent call last):
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/sklearn/model_selection/_validation.py", line 859, in _fit_and_score
    estimator.fit(X_train, y_train, **fit_params)
  File "/opt/micromamba/envs/dev/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/classifier.py", line 392, in fit
    self.model_, self.config_, _ = initialize_tabpfn_model(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/base.py", line 89, in initialize_tabpfn_model
    model, _, config_ = load_model_criterion_config(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/model/loading.py", line 467, in load_model_criterion_config
    loaded_model, criterion, config = load_model(path=model_path, model_seed=model_seed)
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/tabpfn/model/loading.py", line 651, in load_model
    checkpoint = torch.load(path, map_location="cpu", weights_only=None)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/torch/serialization.py", line 1425, in load
    with _open_file_like(f, "rb") as opened_file:
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/torch/serialization.py", line 751, in _open_file_like
    return _open_file(name_or_buffer, mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/dev/lib/python3.12/site-packages/torch/serialization.py", line 732, in __init__
    super().__init__(open(name, mode))
                     ^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/tabpfn/tabpfn-v2-classifier.ckpt'

  warnings.warn(some_fits_failed_message, FitFailedWarning)
[       nan        nan 0.98245614        nan        nan]
```

THIS ONLY WORKS (I.E. FAILS) IF TABPFN'S CHECKPOINTS HAVE NEVER BEEN LOADED BEFORE!

On consecutive runs, the script works as expected:

```sh
(dev) root@39f78d7b4313:/workspace# python tabpfn_bug.py 
[0.97368421 0.97368421 0.98245614 0.98245614 0.99115044]
```

### Versions

```shell
Collecting system and dependency information...
PyTorch version: 2.6.0
CUDA used to build PyTorch: 12.6
ROCM used to build PyTorch: N/A

OS: Debian GNU/Linux 12 (bookworm) (x86_64)
GCC version: (Debian 12.2.0-14+deb12u1) 12.2.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.36

Python version: 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-5.15.0-142-generic-x86_64-with-glibc2.36
Is CUDA available: True
CUDA runtime version: 12.9.86
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA L40
Nvidia driver version: 535.230.02
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        40 bits physical, 48 bits virtual
Byte Order:                           Little Endian
CPU(s):                               64
On-line CPU(s) list:                  0-63
Vendor ID:                            AuthenticAMD
Model name:                           QEMU Virtual CPU version 2.5+
CPU family:                           15
Model:                                107
Thread(s) per core:                   1
Core(s) per socket:                   32
Socket(s):                            2
Stepping:                             1
BogoMIPS:                             6199.99
Flags:                                fpu de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx lm rep_good nopl cpuid extd_apicid tsc_known_freq pni ssse3 cx16 sse4_1 sse4_2 x2apic popcnt aes hypervisor lahf_lm cmp_legacy 3dnowprefetch vmmcall
Hypervisor vendor:                    KVM
Virtualization type:                  full
L1d cache:                            4 MiB (64 instances)
L1i cache:                            4 MiB (64 instances)
L2 cache:                             32 MiB (64 instances)
L3 cache:                             1 GiB (64 instances)
NUMA node(s):                         1
NUMA node0 CPU(s):                    0-63
Vulnerability Gather data sampling:   Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Not affected
Vulnerability Reg file data sampling: Not affected
Vulnerability Retbleed:               Not affected
Vulnerability Spec rstack overflow:   Not affected
Vulnerability Spec store bypass:      Not affected
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Retpolines; STIBP disabled; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Not affected

Dependency Versions:
--------------------
tabpfn: 2.0.9
torch: 2.6.0
numpy: 2.2.6
scipy: 1.16.0
pandas: 2.3.0
scikit-learn: 1.7.0
typing_extensions: 4.14.0
einops: 0.8.1
huggingface-hub: 0.33.1
```

poonai:
## Issue
Fixes #370 

## Motivation and Context
When user tries to initialize TabPFN, `load_model_criterion_config` enters critical region. This issues is fixed by adding lock to the load model function. This ensures no two process can download the model at a time. 

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
- Test case is added to simulate parallel download
- Manually, tested with reproduction script provided in the GitHub issue. 

### Logs of current change:
```
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:05<00:00, 8.52MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 253kB/s]
/home/saint/TabPFN/src/tabpfn/validation.py:56: UserWarning: Running on CPU with more than 200 samples may be slow.
Consider using a GPU or the tabpfn-client API: https://github.com/PriorLabs/tabpfn-client
  _validate_num_samples_for_cpu(
/home/saint/TabPFN/src/tabpfn/validation.py:56: UserWarning: Running on CPU with more than 200 samples may be slow.
```
### Logs of unmodified code:
```
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:05<00:00, 8.58MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 189kB/s]
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:04<00:00, 8.95MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 186kB/s]
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:04<00:00, 10.2MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 264kB/s]
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:04<00:00, 8.90MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 230kB/s]
tabpfn-v2.5-classifier-v2.5_default.ckpt: 100%|██████████████████| 42.9M/42.9M [00:04<00:00, 8.77MB/s]
config.json: 100%|██████████████████████████████████████████████████| 35.0/35.0 [00:00<00:00, 266kB/s]
/home/saint/tabpfn_test/.venv/lib/python3.13/site-packages/tabpfn/validation.py:56: UserWarning: Running on CPU with more than 200 samples may be slow.
```

Without the current change, TabPFN download the same checkpoint multiple times. 

---

## Checklist

---

### DIFF
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -460,7 +460,6 @@ def _get_download_lock() -> FileLock:
     return FileLock(lock_path, timeout=-1)
 
 
-# download lock instance.
 _download_lock: FileLock = _get_download_lock()
 
 


tests/test_model_loading.py
```diff
@@ -367,3 +370,73 @@ def test__prepend_cache_path__multiple_paths__filename_unchanged() -> None:
...
+            _, _, _, _ = model_loading.load_model_criterion_config(
+                model_path=shared_checkpoint_path,
+                check_bar_distribution_criterion=False,
+                cache_trainset_representation=False,
+                which="classifier",
+                version="v2",
+                download_if_not_exists=True,
+            )
+            return True
+        except _:
```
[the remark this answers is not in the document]
poonai: not required

oscarkey:
hey @poonai thanks for the contribution!

I'm worried that, for the average user, this creates more chance of a failure due to the stuck lock, which they might not know how to solve. Maybe we add a 20 second timeout, and after that delete any existing lock? Our model is only 10s of mb big, so this should be more than enough to download it in most cases. cc @brendan-priorlabs what do you think?

brendan-priorlabs:
Thanks for the change @poonai! This does seem worth addressing.

> I'm worried that, for the average user, this creates more chance of a failure due to the stuck lock, which they might not know how to solve. Maybe we add a 20 second timeout, and after that delete any existing lock? Our model is only 10s of mb big, so this should be more than enough to download it in most cases. cc @brendan-priorlabs what do you think?

These are good things to consider. I haven't looked at this carefully yet (in particular I don't know the behavior of the filelock package offhand), but I think we should consider:

1. Making sure this works in both multi-threading and multi-processing cases
2. This bit: https://py-filelock.readthedocs.io/en/latest/index.html#filelocks-and-threads (Honestly on a first pass that documentation almost sounds buggy?)
3. Some kind of automatic cleanup at startup for stale locks (if it's possible for them to get in such a state -- consider testing with `os._exit()` while the lock is acquired). Note that this could be at odds with multi-process support.
4. For the 20 second timeout in particular, I think we may want to avoid that in the interest of being mindful that there's lots of slower connections out there.

poonai:
Hi @oscarkey @brendan-priorlabs , 

Thanks a lot for reviewing the code. Filelock uses kernel level locking system instead of creating a lockfile to notify `not to disturb 😅 `. We don't have to worry about the cleanup. If the process crashes, kernel will release the lock on its own. 

>> Making sure this works in both multi-threading and multi-processing cases

I wrote a sample program to demonstrate the behavior of the Filelock library. 
#### mulit-thread:
```python
import threading
import time
from filelock import FileLock

# Shared lock file
lock_path = "/tmp/test.lock"
lock = FileLock(lock_path)

def worker(name):
    print(f"{name}: trying to acquire lock")
    with lock:  # Uses flock under the hood
        print(f"{name}: acquired lock")
        time.sleep(5)
        print(f"{name}: releasing lock")

# Start two threads
t1 = threading.Thread(target=worker, args=("Thread-1",))
t2 = threading.Thread(target=worker, args=("Thread-2",))

t1.start()
t2.start()
```
output:
```
Thread-1: trying to acquire lock
Thread-2: trying to acquire lock
Thread-1: acquired lock
Thread-1: releasing lock
Thread-2: acquired lock
Thread-2: releasing lock
```
no two thread  acquired lock at a time. 
 
#### multi process:
```python
from filelock import FileLock
import time

lock = FileLock("test.lock")
lock.acquire()
print("lock acquired")
time.sleep(30)
```
I ran this program in multiple terminal. As soon as I kill the process, the second process able to acquire the lock. We don't have to worry about lock file clean up. 
#### terminal 1:
```
saint@saint:~/test$ uv run lock.py
lock acquired
^CTraceback (most recent call last):
  File "/home/saint/test/lock.py", line 7, in <module>
    time.sleep(30)
KeyboardInterrupt
```
#### terminal 2:
```
saint@saint:~/test$ uv run lock.py
lock acquired
lock released
```

>> This bit: https://py-filelock.readthedocs.io/en/latest/index.html#filelocks-and-threads (Honestly on a first pass that documentation almost sounds buggy?

I tested manually with both the configuration. when thread_local is set to `False` both thread can acquire lock. The state which we want to avoid. 
```
Thread-1: trying to acquire lock
Thread-2: trying to acquire lock
Thread-1: acquired lock
Thread-2: acquired lock
Thread-1: releasing lock
Thread-2: releasing lock
```
>> Some kind of automatic cleanup at startup for stale locks (if it's possible for them to get in such a state -- consider testing with os._exit() while the lock is acquired). Note that this could be at odds with multi-process support.

explained above. 

>> For the 20 second timeout in particular, I think we may want to avoid that in the interest of being mindful that there's lots of slower connections out there.

Added 30 sec timeout for the worse case scenario.

brendan-priorlabs:
> Added 30 sec timeout for the worse case scenario.

Like I said before, I don't think we want this. @oscarkey, let me know if you strongly disagree.

oscarkey:
My thought was that the number of people who both run many tabpfn instances in parallel and have slow internet is probably very small, and for them the behaviour doesn't get worse. Whereas a stuck lock file makes the experience much worse for potentially a larger group of people.

But! If the kernel definitely cleans up the lock then no need, so let's remove it. Although it might be nice to print before we try to acquire the lock so we have a debug trail just in case...

poonai:
@oscarkey @brendan-priorlabs  

I've added log info for future debugging as suggested.  Let me know if anything more needs to be added. Happy to do that as well. Hoping this PR gets merged. 

Thanks!!

changelog/738.fixed.md
```diff
@@ -0,0 +1 @@
+Fix model download race condition
```
oscarkey: Maybe "Fix race condition when model is downloaded simultaneously by multiple processes"?

src/tabpfn/model_loading.py
```diff
@@ -453,6 +455,34 @@ def get_cache_dir() -> Path:  # noqa: PLR0911
     return use_instead_path
 
 
+def _get_download_lock() -> FileLock:
+    lock_path = get_cache_dir() / ".download.lock"
+    lock_path.parent.mkdir(parents=True, exist_ok=True)
+    return FileLock(lock_path, timeout=-1)
+
+
+# download lock instance.
```
oscarkey:
```suggestion
```

src/tabpfn/model_loading.py
```diff
@@ -453,6 +455,34 @@ def get_cache_dir() -> Path:  # noqa: PLR0911
...
+_download_lock: FileLock = _get_download_lock()
+
+
+def _with_download_lock(func: Any) -> Any:
+    """Decorator to wrap function with download lock and logging."""
```
oscarkey: Would you expand this slightly to explain why we need the lock?

tests/test_model_loading.py
```diff
@@ -367,3 +370,73 @@ def test__prepend_cache_path__multiple_paths__filename_unchanged() -> None:
     full_paths = model_loading.prepend_cache_path(["my_dir/my_path.test", "another"])
     assert Path(full_paths[0]).name == "my_path.test"
     assert Path(full_paths[1]).name == "another"
+
+
+@patch.dict(ARCHITECTURES, fake_arch=FakeArchitectureModule())
+def test_parallel_model_download(tmp_path: Path) -> None:
```
oscarkey: We usually try to include the function under test and the expected behaviour in the test name
e.g. `test__load_model_criterion_config__parallel_downloads_do_not_crash()`

tests/test_model_loading.py
```diff
@@ -367,3 +370,73 @@ def test__prepend_cache_path__multiple_paths__filename_unchanged() -> None:
...
+            "architecture_name": "fake_arch",
+        }
+
+        # Write the fake checkpoint
+        torch.save(fake_checkpoint, to)
+        return "ok"
+
+    def attempt_load_model() -> bool:
+        """Attempt to load a model, returning success status."""
+        try:
```
oscarkey: Would it be better not to catch this exception?
Then it will be raised by future.result(), and we'll be able to see the error in the test logs, rather than it being hidden.

tests/test_model_loading.py
```diff
@@ -367,3 +370,73 @@ def test__prepend_cache_path__multiple_paths__filename_unchanged() -> None:
...
+
+            for future in as_completed(futures):
+                results.append(future.result())
+
+    # Verify that all threads completed successfully
+    assert len(results) == num_threads
+    assert all(results), "Some threads failed to load model"
+
+    # asserts only one download happened across 5 thread.
+    assert concurrent_downloads == 1, (
```
oscarkey: Maybe I'm missing something here, but won't this just be incremented by each download, concurrent or not?

src/tabpfn/model_loading.py
```diff
@@ -453,6 +455,34 @@ def get_cache_dir() -> Path:  # noqa: PLR0911
...
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        lock_path = _download_lock.lock_file
+        logger.info(f"Acquiring download lock: {lock_path}")
+        _download_lock.acquire()
+        logger.info(f"Acquired download lock: {lock_path}")
```
oscarkey:
```suggestion
        logger.debug(f"Acquired download lock: {lock_path}")
```

src/tabpfn/model_loading.py
```diff
@@ -453,6 +455,34 @@ def get_cache_dir() -> Path:  # noqa: PLR0911
...
+        try:
+            return func(*args, **kwargs)
+        finally:
+            logger.info(f"Releasing download lock: {lock_path}")
```
oscarkey:
```suggestion
            logger.debug(f"Releasing download lock: {lock_path}")
```

brendan-priorlabs:
Thanks for the followups, @poonai! One final thing from my side, the docs I linked earlier say "Behavior where [FileLock](https://py-filelock.readthedocs.io/en/latest/api.html#filelock.FileLock) is thread-local started in version 3.11.0." If you download an earlier version, e.g., 3.10, and keep your code as-is does the behavior change? Probably worth running both multi-thread and multi-process experiments with that dependency. If it behaves the same, I think we're good to go!

tests/test_model_loading.py
poonai: The lock, turns all the concurrent `load_model_criterion_config` call into  a serial call. Therefore the first call downloads the models and the subsequent call does not call `model_download`. Since, the `model_download` function is executed, only when the model is not already downloaded. 

The existing logic is correct. I'll change the variable name from `concurrent_downloads ` to `download_attempts` to avoid ambiguity.

poonai:
@brendan-priorlabs Your concern is right. The test script didn't behave as expected in `3.10`. Absence of thread_local feature causing two thread to acquire lock, at a time. To mitigate the unexpected behavior, I've bumped the minimum version from 1.0 to 3.11

brendan-priorlabs:
Thanks for looking into that and fixing, @poonai! I think we can call this good from my side. @oscarkey, by all means ping me if anything else comes up. Otherwise will leave final approval to you

tests/test_model_loading.py
oscarkey: ah sweet! That makes sense, thank you.

PR #740, opened 2026-01-19
[Prep-refactor 5] Refactor SequentialEncoder

bejaeger:
No functional changes are done in this PR. It refactors the `SequentialEncoder` for further changes. 
In particular:
- Allows to decouple projections and preprocessing steps.
- Makes preprocessing pipeline more explicit by parsing dict instead of tensors.

More work is needed to disentangle the preprocessing "state" from the model after this PR is merged.

### DIFF
--- a/src/tabpfn/architectures/encoders/steps/normalize_feature_groups_encoder_step.py
+++ b/src/tabpfn/architectures/encoders/steps/normalize_feature_groups_encoder_step.py
@@ -78,11 +78,11 @@ class NormalizeFeatureGroupsEncoderStep(TorchPreprocessingStep):
         # Checks for constant features to scale features in group that
         # have constant features. Constant features could have been added
         # from padding to feature group size.
-        sel = (x[1:] == x[0]).sum(0) != (x.shape[0] - 1)
+        self.non_constants_mask_ = ((x[1:] == x[0]).sum(0) != (x.shape[0] - 1)).cpu()
         # move to cpu to avoid dangling GPU memory, tested in
         # test__to__after_fit_and_predict__no_tensors_left_on_old_device
         self.number_of_used_features_ = torch.clip(
-            sel.sum(-1).unsqueeze(-1),
+            self.non_constants_mask_.sum(-1).unsqueeze(-1),
             min=1,
         ).cpu()
 
@@ -127,5 +127,8 @@ class NormalizeFeatureGroupsEncoderStep(TorchPreprocessingStep):
 
         scale = self.num_features_per_group / self.number_of_used_features_.to(x.device)
         x = x * torch.sqrt(scale) if self.normalize_by_sqrt else x * scale
+
+        # Set constant features to 0 to avoid scaling them
+        x[:, ~self.non_constants_mask_] = 0
 
         return {self.out_keys[0]: x}
--- a/src/tabpfn/architectures/encoders/steps/remove_empty_features_encoder_step.py
+++ b/src/tabpfn/architectures/encoders/steps/remove_empty_features_encoder_step.py
@@ -48,7 +48,7 @@ class RemoveEmptyFeaturesEncoderStep(TorchPreprocessingStep):
         """
         del kwargs
         x = state[self.in_keys[0]]
-        self.column_selection_mask_ = (x[1:] == x[0]).sum(0) != (x.shape[0] - 1)
+        self.column_selection_mask = (x[1:] == x[0]).sum(0) != (x.shape[0] - 1)
 
     @override
     def _transform(
@@ -72,7 +72,7 @@ class RemoveEmptyFeaturesEncoderStep(TorchPreprocessingStep):
         orig_last_dim = x.shape[-1]
         # Ensure that the mask is a bool, because the buffer may get converted to a
         # a float if .to() is called on the containing module.
-        x = select_features(x, self.column_selection_mask_.type(torch.bool))
+        x = select_features(x, self.column_selection_mask.type(torch.bool))
 
         potential_padding = -x.shape[-1] % orig_last_dim
         x = torch.nn.functional.pad(x, pad=(0, potential_padding), value=0)
--- a/tests/test_encoders/test_encoders.py
+++ b/tests/test_encoders/test_encoders.py
@@ -101,7 +101,7 @@ def test__remove_empty_features_encoder():
 
 
 def test__variable_num_features_encoder():
-    N, B, F, fixed_out = 10, 3, 4, 5
+    N, B, F, fixed_out = 10, 3, 5, 5
     x = torch.rand([N, B, F])
 
     kwargs = {"num_features_per_group": fixed_out}


bejaeger:

- #740 ◀

src/tabpfn/architectures/encoders/embedders.py
```diff
@@ -0,0 +1,122 @@
...
+        *,
+        num_features_per_group_with_metadata: int,
+        emsize: int,
+        replace_nan_by_zero: bool = False,
+        bias: bool = True,
+    ) -> None:
+        """Initialize the projection.
+
+        Args:
+            num_features_per_group_with_metadata: The number of input features per group
```
LeoGrin: I find `num_features_per_group_with_metadata` a bit ambiguous. It could be the number of features which have metadata available. And it's a bit confusing IMO because it's not super common to flatten metadata in a tensor (we could explain in the docstring that we do that).
One strategy would be to only talk about the nan_indicators which we have now so `num_features_per_group_maybe_plus_nan_ind` (ideally shorter and clearer haha). If we want to stay general, maybe `group_size_with_metadata`? or `num_features_per_group_incl_metadata` (there's probably something better)

src/tabpfn/architectures/encoders/steps/categorical_input_encoder_per_feature_encoder_step.py
```diff
@@ -3,14 +3,19 @@
...
 
 import torch
 from torch import nn
 
-from tabpfn.architectures.encoders import SeqEncStep
+from tabpfn.architectures.encoders import TorchPreprocessingStep
 
 
-class CategoricalInputEncoderPerFeatureEncoderStep(SeqEncStep):
+# TODO: This step currently has learnable parameters, which means it breaks
```
LeoGrin: not sure but we could also delete this?

src/tabpfn/architectures/encoders/steps/nan_handling_encoder_step.py
```diff
@@ -34,29 +35,54 @@ def __init__(
...
         self.keep_nans = keep_nans
         self.register_buffer("feature_means_", torch.tensor([]), persistent=False)
 
-    def _fit(self, x: torch.Tensor, single_eval_pos: int, **kwargs: Any) -> None:  # noqa: ARG002
+    @override
+    def _fit(
+        self,
+        state: dict[str, torch.Tensor],
+        single_eval_pos: int | None = None,
+        **kwargs: Any,
```
LeoGrin, line 48: nit: out of curiosity do you remember which kwarg is passed to this? (to know how hard it would be to just remove the kwarg)

src/tabpfn/architectures/encoders/steps/feature_group_padding_encoder_step.py
```diff
@@ -0,0 +1,110 @@
...
+
+from typing import Any
+from typing_extensions import override
+
+import torch
+
+from tabpfn.architectures.encoders import TorchPreprocessingStep
+
+
+class FeatureGroupPaddingAndReshapeStep(TorchPreprocessingStep):
```
LeoGrin: nit: what about FeaturePaddingAndReshapeStep? I first thought it applied to a single feature group. Not super sure.

src/tabpfn/architectures/encoders/steps/normalize_feature_groups_encoder_step.py
```diff
@@ -23,72 +24,101 @@ def __init__(
         *,
```
LeoGrin, line 23: nit: I know it's not from this but num_feature here is actually the feature group size right? Maybe we can rename?

src/tabpfn/architectures/encoders/steps/variable_num_features_encoder_step.py
```diff
@@ -23,72 +24,101 @@ def __init__(
...
+        # have constant features. Constant features could have been added
+        # from padding to feature group size.
         sel = (x[1:] == x[0]).sum(0) != (x.shape[0] - 1)
+        # move to cpu to avoid dangling GPU memory, tested in
+        # test__to__after_fit_and_predict__no_tensors_left_on_old_device
         self.number_of_used_features_ = torch.clip(
             sel.sum(-1).unsqueeze(-1),
             min=1,
         ).cpu()
+        self.padding_features_ = -x.shape[-1] % self.num_features_per_group
```
LeoGrin: can this be different from zero if we have FeatureGroupPaddingAndReshapeStep before?

src/tabpfn/architectures/encoders/steps/variable_num_features_encoder_step.py
```diff
@@ -3,13 +3,14 @@
...
+from typing_extensions import override
 
 import torch
 
-from tabpfn.architectures.encoders import SeqEncStep
+from tabpfn.architectures.encoders import TorchPreprocessingStep
 
 
-class VariableNumFeaturesEncoderStep(SeqEncStep):
+class VariableNumFeaturesEncoderStep(TorchPreprocessingStep):
```
LeoGrin: (I find this step quite confusing, in particular in how it interacts with FeatureGroupPaddingAndReshapeStep.)

src/tabpfn/architectures/encoders/embedders.py
bejaeger: Agreed! I made the `embedder.py` much more generic and the implementation/naming in the model will be more task-specific. E.g. we can go with `feature_group_encoding_size`. However, I decided to entirely remove the `embedder.py` for the sake of this PR, because it's not used anywhere right now so probably shouldn't be included. Will be included later when actually needed.

src/tabpfn/architectures/encoders/steps/categorical_input_encoder_per_feature_encoder_step.py
bejaeger: I like it, removed!

src/tabpfn/architectures/encoders/steps/nan_handling_encoder_step.py
bejaeger: Yep, there is `single_eval_pos`, `cache_trainset_representation`, and potentially `categorical_inds` being parsed to the pipeline.

src/tabpfn/architectures/encoders/steps/feature_group_padding_encoder_step.py
bejaeger: I removed it for now because it was specific to something that is not included in this PR.

src/tabpfn/architectures/encoders/steps/normalize_feature_groups_encoder_step.py
bejaeger: Yep it is, renamed it

src/tabpfn/architectures/encoders/steps/variable_num_features_encoder_step.py
bejaeger: Nope, but since I removed the other step for now it can.

src/tabpfn/architectures/encoders/steps/variable_num_features_encoder_step.py
bejaeger: Yes agreed, because it does two steps, padding and scaling. The padding logic was actually flawed I think in the previous implementation but was never used because we pad to multiples of feature group before calling the `encoder` everywhere. So I removed the padding and only call it `ScaleFeatureGroupsEncoderStep`

LeoGrin: oh yeah that's great!

src/tabpfn/architectures/base/__init__.py
```diff
@@ -127,45 +125,51 @@ def get_encoder(  # noqa: PLR0913
...
+) -> TorchPreprocessingPipeline:
+    inputs_to_merge = {"main": {"dim": num_features_per_group}}
 
-    encoder_steps: list[SeqEncStep] = []
+    encoder_steps: list[TorchPreprocessingStep] = []
     if remove_empty_features:
         encoder_steps += [RemoveEmptyFeaturesEncoderStep()]
 
     if remove_duplicate_features:
+        # TODO: This is a No-op currently. We cannot remove it
```
alanprior, line 136: @bejaeger Could we wrap (135, 140) with a single row func that doesn't disturb reading?

i.e.
`_legacy_no_op_duplicates(remove_duplicate_features)`?

src/tabpfn/architectures/base/__init__.py
```diff
@@ -127,45 +125,51 @@ def get_encoder(  # noqa: PLR0913
...
-            VariableNumFeaturesEncoderStep(
-                num_features=num_features,
-                normalize_by_used_features=False,
-                in_keys=["nan_indicators"],
-                out_keys=["nan_indicators"],
-            ),
-        ]
+        inputs_to_merge["nan_indicators"] = {"dim": num_features_per_group}
+
+        if normalize_by_used_features:
```
alanprior, line 147: @bejaeger same comment as above. Remove from the main flow (?) as much as possible redundant text. Keep the super important documentation inside that helper function

src/tabpfn/architectures/base/transformer.py
```diff
@@ -155,29 +155,36 @@ def __init__(  # noqa: D417, PLR0913
...
-                    out_keys=("output",),
-                ),
+            encoder = TorchPreprocessingPipeline(
+                steps=[
+                    LinearInputEncoderStep(
+                        num_features=1,
+                        emsize=config.emsize,
+                        replace_nan_by_zero=False,
+                        bias=True,
+                        in_keys=("main",),
```
alanprior, line 165: @bejaeger I know this is legacy, but I found it hard to understand why we need / want to pass these keys at all, and even worse, why they are hardcoded?

In which cases do we need to control these keys and make them configurable?

src/tabpfn/architectures/base/transformer.py
```diff
@@ -155,29 +155,36 @@ def __init__(  # noqa: D417, PLR0913
...
-                    emsize=config.emsize,
-                    replace_nan_by_zero=False,
-                    bias=True,
-                    in_keys=("main",),
-                    out_keys=("output",),
-                ),
+            encoder = TorchPreprocessingPipeline(
+                steps=[
+                    LinearInputEncoderStep(
+                        num_features=1,
```
alanprior, line 161: @bejaeger unclear to me why is num_features hardcoded here? I thought that this is a parameter?

src/tabpfn/architectures/base/transformer.py
```diff
@@ -155,29 +155,36 @@ def __init__(  # noqa: D417, PLR0913
...
-                    replace_nan_by_zero=False,
-                    bias=True,
-                    out_keys=("output",),
-                    in_keys=("main", "nan_indicators"),
-                ),
+            y_encoder = TorchPreprocessingPipeline(
+                steps=[
+                    NanHandlingEncoderStep(),
+                    LinearInputEncoderStep(
+                        num_features=2,
```
alanprior, line 180: @bejaeger very unclear to me why `y_encoder` gets a num_features=2.

src/tabpfn/architectures/encoders/pipeline_interfaces.py
```diff
@@ -4,99 +4,106 @@
...
+        # output_key is set for backwards compatibility
+        # with encoders that still have the projections in this pipeline.
+        output_key: str | None = None,
+    ):
+        super().__init__(*steps)
         self.output_key = output_key
 
-    def forward(self, input: dict, **kwargs: Any) -> torch.Tensor:
-        """Apply the sequence of encoder steps to the input.
+        # If we are in backwards compatibility mode, we expect the last step to
```
alanprior, line 39: @bejaeger could this be wrapped in a function like self._assert_whatever()?

src/tabpfn/architectures/encoders/pipeline_interfaces.py
```diff
@@ -4,99 +4,106 @@
...
         Args:
-            input:
-                The input state dictionary.
-                If the input is not a dict and the first layer expects one input key,
-                the input tensor is mapped to the key expected by the first layer.
-            **kwargs: Additional keyword arguments passed to each encoder step.
+            input: The input tensor or dictionary of tensors in the case of
+                multiple modalities. If a tensor is provided, it is wrapped in a
+                dictionary with the key "main". The tensor must have the shape
+                [R, B, F] with
```
alanprior: @bejaeger is this notation consistent with (Ri, B, C) that we use later? I'm unsure when we use columns and features, and what is the i added (you said something about train/test but I still don't understand it)

src/tabpfn/architectures/encoders/pipeline_interfaces.py
```diff
@@ -4,99 +4,106 @@
...
+            A dictionary of tensors with all keys created in the pipeline or
+            (for backwards compatibility only) the output tensor specified by
+            the `output_key` parameter of the last step.
         """
-        # If the input is not a dict and the first layer expects one input, mapping the
-        #   input to the first input key must be correct
-        if not isinstance(input, dict) and len(self[0].in_keys) == 1:
-            input = {self[0].in_keys[0]: input}  # noqa: A001
-
+        x = {"main": input} if isinstance(input, torch.Tensor) else input
```
alanprior, line 79: @bejaeger this {"main": input"} would keep happening? Why do we need/want to support tensors directly?

src/tabpfn/architectures/encoders/pipeline_interfaces.py
```diff
@@ -174,26 +190,19 @@ def forward(
...
-            tuple,
-        ), (
-            f"out is not a tuple: {out}, type: {type(out)}, class: "
-            f"{self.__class__.__name__}"
-        )
-        assert len(out) == len(self.out_keys)
-        state.update({out_key: out[i] for i, out_key in enumerate(self.out_keys)})
+        self._validate_input_keys(state)
+
+        do_fit = (
```
alanprior, line 199: @bejaeger nit: perhaps make a func so you do 

if do_fit():

Making the main flow more readable

src/tabpfn/architectures/encoders/steps/feature_group_projections_encoder_step.py
```diff
@@ -1,20 +1,22 @@
-"""Projections from cell-level tensors to embedding space."""
+"""DEPRECATED: Projections from cell-level tensors to embedding space.
```
alanprior, line 1: @bejaeger I assume we want this for backwards compatibility of some sort, but, actually, why should we? Seems like a lot of the code has a lot of old stuff. I'm wondering if that's justified

src/tabpfn/architectures/encoders/steps/nan_handling_encoder_step.py
```diff
@@ -34,29 +35,54 @@ def __init__(
             out_keys: The keys to assign the output tensors to.
         """
         assert len(in_keys) == 1, "NanHandlingEncoderStep expects a single input key"
+        if len(out_keys) > 1 and not keep_nans:
```
alanprior: @bejaeger maybe if self._validate_whatever()?

src/tabpfn/architectures/base/transformer.py
bejaeger: it's the target encoder for which it makes sense. The model is constructed to know a single target so we can hardcode it here.

alanprior: Ok, to be honest I'm a bit lost and unsure what is this class and where it is used. 

I thought `y_encoder` is the target encoder (and unclear to me why it uses num_features=2). Anyway, I see that you aren't changing any behavior and probably I could understand this if I tried to look on the whole code

src/tabpfn/architectures/encoders/steps/feature_group_projections_encoder_step.py
bejaeger: Yep there is a good reason. Unfortunately, the exact order of the preprocessings is tied to the model checkpoint state dict. Before we can deprecate this, we need a layer between model loading and checkpoint when loading the state dict. Since I didn't want to blow up this PR, I think we should leave this for now.

src/tabpfn/architectures/base/__init__.py
bejaeger: Done

src/tabpfn/architectures/base/__init__.py
bejaeger: This one is actually not redundant, so leaving it as is.

src/tabpfn/architectures/base/transformer.py
bejaeger: When defining new pipelines. It's one way of designing a pipeline that's flexible and where each step "consumes" a certain input, and adds it back to the "stream" (object that's parsed through the pipeline).

src/tabpfn/architectures/base/transformer.py
bejaeger: Yeah this is the legacy logic that's very opaque and we want to get rid of slowly. The NanHandlingEncoderStep adds a "nan_indicator" that the `LinearInputEncoderStep` consumes. That's why 2 features.

src/tabpfn/architectures/encoders/pipeline_interfaces.py
bejaeger: Thanks for spotting it. the docstring was outdated. Updated it. 
Since I define exactly what R, B, F means below, I think it's fine. When we don't need the "i" because we don't need to distinguish it to other "row sizes" (e.g. only train), we can drop the "i".

src/tabpfn/architectures/encoders/pipeline_interfaces.py
bejaeger: Fair point, I'll take it into account when applying this pipeline to the new model files

src/tabpfn/architectures/encoders/pipeline_interfaces.py
bejaeger: I think for small conditionals this is a bit easier. When reading, it's easy to skip the logic of the conditional. A function wouldn't make the code shorter imo.

bejaeger:
Thanks @alanprior and @LeoGrin !
I implemented a bunch of your suggestions and a few of the comments will be addressed in future PRs.

src/tabpfn/architectures/encoders/steps/feature_group_projections_encoder_step.py
alanprior: Right, I would imagine that and completely agree... but would it make sense to have a following PR changing the model's state dict and killing this part? (might be too hard or break other things, but in principle)

src/tabpfn/architectures/encoders/pipeline_interfaces.py
alanprior: In this case agree it won't matter as much

src/tabpfn/architectures/encoders/pipeline_interfaces.py
alanprior: Great :)

So what Ri means? I understand R is rows of all. Ri is only train? what would be test? why Ri?

src/tabpfn/architectures/base/transformer.py
alanprior: Ohhhhhhhhhh

alanprior: Well, adding a small comment here could be incredible for sanity, but hopefully I will remember it.

alanprior:
> LGTM thanks a lot!! The new names are much clearer :)

+1 on this. Looks much much clearer and better. I'm learning a lot :)

src/tabpfn/architectures/encoders/steps/feature_group_projections_encoder_step.py
alanprior: (Anyway definitely leave aside of this PR!)

src/tabpfn/architectures/base/transformer.py
alanprior: Hmmm yeah this is not something I've never seen before. I guess it makes sense. I would still think of harcoding them under some StrEnum like, but that's too small to care about. Let's leave for now

src/tabpfn/architectures/encoders/pipeline_interfaces.py
bejaeger: should be documented where we use Ri. ("i" for input I think)

src/tabpfn/architectures/encoders/steps/feature_group_projections_encoder_step.py
bejaeger: Yes, that's what we should do

src/tabpfn/architectures/encoders/pipeline_interfaces.py
alanprior: Hmm, a bit of a confusing way to distinguish, but sadly Train and Test have the same first letter :\

PR #741, opened 2026-01-19
Fix bug in demo notebook where backend input selection did not appear.

oscarkey:
Use console.print() + input() rather than Prompt.ask().

I tried this out in Colab, and now the input box always appears.

This is how it looks in colab:
Screenshot 2026-01-19 at 14 09 22

and in local jupyter:
Screenshot 2026-01-19 at 14 09 50

and in vscode (not so good, but it was the same before):
Screenshot 2026-01-19 at 14 10 34

### DIFF
--- a/examples/notebooks/TabPFN_Demo_Local.ipynb
+++ b/examples/notebooks/TabPFN_Demo_Local.ipynb
@@ -665,7 +665,7 @@
     "backend = None\n",
     "while backend is None:\n",
     "  console.print(\n",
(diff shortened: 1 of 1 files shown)


noahho (approved):
LGTM - I tested twice and it worked. I'm not sure if removing the tip to restart cell is a good call though because it might still fail flakily I am told by ChatGPT

oscarkey:
okay I put it back, just in case.

PR #745, opened 2026-01-22
improved regressor memory usage by 60%

### LINKED ISSUE #354: Regressor memory increase with each estimator iteration

The regressor saves the regression output which is a vector of probabilities for each output
https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/regressor.py#L914
This can grow quite large: 10000 samples * 5000 borders * 16byte * 8 estimators = 6.4GB which are all stored on GPU memory

Saving a compressed version of the borders (dropping out low probability borders) would be much more memory efficient. Since the probabilities and borers are recombined anyways with different borders, we could do the same and remove a lot of the low probability borders.

_No response_

### Additional context

_No response_

### Impact

None

poonai:
## Issue
Fixes #354 

## Motivation and Context
The Github issue suggest to remove the low probability borders to reduce the regressor memory usage. Instead, I refactored the existing code to achieve the same result.  The proposed code change, incrementally calculates the sum of estimator outputs instead of calculating the sum by accumulating all the estimator output. 

#### existing code:
 
```python
outputs = []
for output in iter_estimator_output:
    outputs.append(output)

avg_mean = sum(ouputs) / n_estimator
``` 
#### proposed code:

```python
sum = 0
for output in iter_estimator_output:
    sum += output

avg_mean = sum / n_estimator
```    
---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
I've tested the memory usage by tweaking the fine tuning example. 
```python
def main() -> None:
    data = sklearn.datasets.fetch_california_housing(as_frame=True)
    X_all = data.data
    y_all = data.target

        X_all, y_all, test_size=0.1, random_state=RANDOM_STATE
    )

    print(
        f"Loaded {len(X_train):,} samples for training and "
        f"{len(X_test):,} samples for testing."
    )

    # 2. Initial model evaluation on test set
    base_reg = TabPFNRegressor(
        device=["cuda:0"],
        n_estimators=NUM_ESTIMATORS_FINAL_INFERENCE,
        ignore_pretraining_limits=True,
        inference_config={"SUBSAMPLE_SAMPLES": 50_000},
    )
    base_reg.fit(X_train, y_train)

    for _ in range(5):
        base_pred = base_reg.predict(X_test)    
    print("torch.cuda.memory_allocated: %fGB"%(torch.cuda.memory_allocated(0)/1024/1024/1024))
    print("torch.cuda.memory_reserved: %fGB"%(torch.cuda.memory_reserved(0)/1024/1024/1024))
    print("torch.cuda.max_memory_reserved: %fGB"%(torch.cuda.max_memory_reserved(0)/1024/1024/1024))
    



if __name__ == "__main__":
    if not torch.cuda.is_available():
        raise RuntimeError(
            "CUDA is not available. Please run the script on a CUDA-enabled GPU."
        )
    main()

```

#### previous allocation:
```
torch.cuda.memory_allocated: 0.046939GB
torch.cuda.memory_reserved: 1.638672GB
torch.cuda.max_memory_reserved: 1.638672GB
```

#### current allocation:
```
torch.cuda.memory_allocated: 0.046939GB
torch.cuda.memory_reserved: 1.021484GB
torch.cuda.max_memory_reserved: 1.021484GB
```

#### gain
`1.021/1.638 * 100 = 62.33`

 `memory_reserved` is the total memory allocated during the entire life of the program. `memory_allocated` represent the current allocation of the program. I assumes, accumulated memory is released to the allocator after the `predict` function returns. Hence the no change in `memory_allocated`.  I think it's fair to use gain of `memory_reserved` metric to argue the benefit of this code change. Please correct me if something is wrong. 

---

## Checklist

---

Fixes PRI-33

### DIFF
--- a/src/tabpfn/regressor.py
+++ b/src/tabpfn/regressor.py
@@ -963,7 +963,7 @@ class TabPFNRegressor(RegressorMixin, BaseEstimator):
                     accumulated_logits = accumulated_logits + transformed
                 n_estimators += 1
 
-        assert n_estimators > 0, "n_estimators must be greater than 0"
+        assert n_estimators > 0
 
         if accumulated_logits is None:
             raise ValueError(


src/tabpfn/regressor.py
```diff
@@ -988,31 +994,12 @@ def predict(
 
         return logit_to_output(output_type=output_type)
 
-    def forward(
+    def _iter_forward_executor(
```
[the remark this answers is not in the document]
poonai: resolved

poonai:
@oscarkey friendly ping to review this PR. 

This is a small change and requires less time to review the PR. I know you may be busy with other work so no rush at all. Review like this encourages me to do more opensource contribution.

oscarkey:
hey @poonai , sorry for the delay. I'll get to both your prs on monday :)

poonai:
@gemini-code-assist review

src/tabpfn/regressor.py
```diff
@@ -913,29 +913,34 @@ def predict(
...
+                    averaged_logits = averaged_logits + transformed
+                n_estimators += 1
+
+        # Finalize averaging
         if self.average_before_softmax:
-            logits = stacked_logits.log().mean(dim=0).softmax(dim=-1)
+            logits = (averaged_logits / n_estimators).softmax(dim=-1)  # type: ignore
         else:
-            logits = stacked_logits.mean(dim=0)
+            logits = averaged_logits / n_estimators  # type: ignore
```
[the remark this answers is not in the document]
poonai: At least 1 estimator exist at a time. Hence, I think it's irrelevant. Correct me if I'm wrong

noahho:
@oscarkey ping here

src/tabpfn/regressor.py
```diff
@@ -913,29 +913,34 @@ def predict(
         )
 
         # Runs over iteration engine
+
+        n_estimators = 0
+        averaged_logits: torch.Tensor | None = None
```
oscarkey: maybe we should name this `accumulated_logits`?

src/tabpfn/regressor.py
```diff
@@ -913,29 +913,34 @@ def predict(
...
+                    to=self.znorm_space_bardist_.borders.to(output.device),
+                )
+
+                if self.average_before_softmax:
+                    transformed = transformed.log()
+
+                if averaged_logits is None:
+                    averaged_logits = transformed
+                else:
+                    averaged_logits = averaged_logits + transformed
```
oscarkey: what do you think about doing the accumulation in float64?
```
  if accumulated_logits is None:                                                                                                                                                                                                                                                                                              
      accumulated_logits = transformed.to(torch.float64)                                                                                                                                                                                                                                                                      
  else:
      accumulated_logits += transformed.to(torch.float64) 
```

poonai:
I’ve gone to a pilgrimage and return home by this Friday. I will address your comment by that time @oscarkey

oscarkey:
sounds great! No rush :)

src/tabpfn/regressor.py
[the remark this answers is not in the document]
noahho: should we add an assert for n_estimators?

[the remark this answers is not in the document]
poonai: it's better to have the assertion and I've added it.

src/tabpfn/regressor.py
poonai: currently the accumulated_logits datatype is same as incoming tensor's datatype. 

why do we have to change the datatype to float64 in the middle? Am I missing anything that I should be aware of?

src/tabpfn/regressor.py
poonai: done. good suggestion.

src/tabpfn/regressor.py
```diff
@@ -942,36 +942,38 @@ def predict(
...
+                else:
+                    accumulated_logits = accumulated_logits + transformed
+                n_estimators += 1
+
+        assert n_estimators > 0, "n_estimators must be greater than 0"
+
+        if accumulated_logits is None:
+            raise ValueError(
+                "Cannot make predictions, possibly due to `n_estimators=0`."
             )
```
oscarkey:
```suggestion
        assert n_estimators > 0 and accumulated_logits is not None
```
As we never expect this to happen, I think this less verbose check will suffice

src/tabpfn/regressor.py
oscarkey: I was worried that we might overflow now the sum/division are split into separate operations. But, thinking about it more, I think this is a pretty unlikely error that will be obvious if it does happen. So let's leave it how it is for now :)

src/tabpfn/regressor.py
poonai: The suggested code throws this `Assertion should be broken down into multiple parts` linter error. So I've removed the `and` statement to make the linter happy. I think that's okay since accumulated_logits can't be none when `n_estimator>0`

oscarkey: okay sounds good!

PR #747, opened 2026-01-23
Introduce revamped torch preprocessing pipeline

bejaeger:
Initial implementation of a cleaned-up torch preprocessing pipeline. 

Compared to the pipeline in `architectures/encoders` this new pipeline:
- Is modality-aware
- Removes dependency on nn.Module
- Makes state (fitted cache) handling explicit rather than implicit, which allows more flexible use.

The new pipeline can be used more easily in:
1. Before model forward pass for gpu-based fast preprocessing
2. In the model forward pass 

For inference, the torch pipeline is created inside `TabPFNEnsemblePreprocessor` and then passed to each `TabPFNPreprocessedEnsembleMember`. The naming of the latter might be adjusted when harmonizing the cpu based preprocessing.

### DIFF
--- a/src/tabpfn/preprocessing/torch/factory.py
+++ b/src/tabpfn/preprocessing/torch/factory.py
@@ -12,7 +12,7 @@ from tabpfn.preprocessing.torch.pipeline_interface import (
     TorchPreprocessingStep,
 )
 from tabpfn.preprocessing.torch.steps import (
-    TorchRemoveOutliersStep,
+    TorchSoftClipOutliersStep,
 )
 
 if TYPE_CHECKING:
@@ -30,7 +30,7 @@ def create_gpu_preprocessing_pipeline(
     if config.outlier_removal_std is not None:
         steps.append(
             (
-                TorchRemoveOutliersStep(n_sigma=config.outlier_removal_std),
+                TorchSoftClipOutliersStep(n_sigma=config.outlier_removal_std),
                 {FeatureModality.NUMERICAL},
             )
         )
--- a/src/tabpfn/preprocessing/torch/steps.py
+++ b/src/tabpfn/preprocessing/torch/steps.py
@@ -40,8 +40,8 @@ class TorchStandardScalerStep(TorchPreprocessingStep):
         return self._scaler.transform(x, fitted_cache=fitted_cache), None, None
 
 
-class TorchRemoveOutliersStep(TorchPreprocessingStep):
-    """Pipeline step wrapper for TorchRemoveOutliers."""
+class TorchSoftClipOutliersStep(TorchPreprocessingStep):
+    """Pipeline step wrapper for TorchSoftClipOutliers."""
 
     def __init__(self, n_sigma: float = 4.0) -> None:
         """Initialize the outlier removal step.
--- a/src/tabpfn/preprocessing/torch/__init__.py
+++ b/src/tabpfn/preprocessing/torch/__init__.py
@@ -11,7 +11,7 @@ from .pipeline_interface import (
     TorchPreprocessingStep,
     TorchPreprocessingStepResult,
 )
-from .steps import TorchRemoveOutliersStep, TorchStandardScalerStep
+from .steps import TorchSoftClipOutliersStep, TorchStandardScalerStep
 from .torch_soft_clip_outliers import TorchSoftClipOutliers
 from .torch_standard_scaler import TorchStandardScaler
 
@@ -21,8 +21,8 @@ __all__ = [
     "TorchPreprocessingPipelineOutput",
     "TorchPreprocessingStep",
     "TorchPreprocessingStepResult",
-    "TorchRemoveOutliersStep",
     "TorchSoftClipOutliers",
+    "TorchSoftClipOutliersStep",
     "TorchStandardScaler",
     "TorchStandardScalerStep",
     "create_gpu_preprocessing_pipeline",
--- a/tests/test_torch_preprocessing/test_pipeline_interface.py
+++ b/tests/test_torch_preprocessing/test_pipeline_interface.py
@@ -10,7 +10,7 @@ import torch
 from tabpfn.preprocessing.datamodel import (
     FeatureModality,
 )
-from tabpfn.preprocessing.torch import TorchRemoveOutliersStep
+from tabpfn.preprocessing.torch import TorchSoftClipOutliersStep
 from tabpfn.preprocessing.torch.pipeline_interface import (
     ColumnMetadata,
     TorchPreprocessingPipeline,
@@ -204,7 +204,7 @@ def test__call__no_num_train_rows_fits_on_all_data():
 
 def test__call__zero_num_train_rows():
     """Test that fit is skipped when num_train_rows is 0."""
-    step = TorchRemoveOutliersStep()
+    step = TorchSoftClipOutliersStep()
     pipeline = TorchPreprocessingPipeline(steps=[(step, {FeatureModality.NUMERICAL})])
     metadata = ColumnMetadata(
         indices_by_modality={FeatureModality.NUMERICAL: [0]},
--- a/tests/test_torch_preprocessing/test_torch_soft_clip_outliers.py
+++ b/tests/test_torch_preprocessing/test_torch_soft_clip_outliers.py
@@ -1,4 +1,4 @@
-"""Tests for TorchRemoveOutliers."""
+"""Tests for TorchSoftClipOutliers."""
 
 from __future__ import annotations
 


bejaeger:

- #747 ◀

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -0,0 +1,54 @@
...
+import dataclasses
+from enum import Enum
+
+import pandas as pd
+
+
+class FeatureModality(str, Enum):
+    """The modality of a feature.
+
+    Here we move between the way the data is stored, and what it actually
```
LeoGrin:
```suggestion
    This should be what the column actually represent, not how it is stored. For instance...
```

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+        keep_fitted_cache: bool = False,
+    ) -> None:
+        """Initialize with list of (step, target_modalities) pairs.
+
+        Args:
+            steps: List of (step, modalities) where modalities is a set of
+                FeatureModality values the step should be applied to.
+            keep_fitted_cache: Whether to keep the fitted cache between calls.
+                If True, the fitted cache will be reused in the transform step.
+                If False, the fitted cache will be fit on the training data for each
```
LeoGrin: not super clear IMO. Maybe something like: "If True, saves the fitted state of all steps on the first call, to be reused when passing use_fitted_cache in a second call."

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -36,10 +40,15 @@
...
+    """Holds preprocessed data, config, preprocessors for a single ensemble member.
+
+    The data is preprocessed on the CPU but this member also holds a torch preprocessor
+    pipeline to be run before inference on the GPU.
+    """
 
     config: EnsembleConfig
-    preprocessor: SequentialFeatureTransformer
+    cpu_preprocessor: SequentialFeatureTransformer
+    gpu_preprocessor: TorchPreprocessingPipeline | None
```
LeoGrin, line 51: nit: maybe torch_preprocessor, as it can in theory also run on cpu if that's the device? Not sure

src/tabpfn/architectures/encoders/__init__.py
```diff
@@ -1,3 +1,7 @@
+"""DEPRECATION WARNING: Please note that this module will be deprecated in future
+versions of TabPFN in favor of a new torch-based preprocessing pipeline.
+"""
```
alanprior, line 3: @bejaeger what is our deprecation strategy? (could be discussed internally)

src/tabpfn/preprocessing/torch/factory.py
```diff
@@ -0,0 +1,41 @@
...
+)
+from tabpfn.preprocessing.torch.steps import (
+    TorchRemoveOutliersStep,
+)
+
+if TYPE_CHECKING:
+    from tabpfn.preprocessing.configs import EnsembleConfig
+
+
+def create_gpu_preprocessing_pipeline(
```
alanprior, line 22: @bejaeger nice!!

src/tabpfn/preprocessing/torch/ops.py
```diff
@@ -0,0 +1,98 @@
...
+    nan_mask = torch.isnan(x)
+    if include_inf:
+        nan_mask = torch.logical_or(nan_mask, torch.isinf(x))
+
+    num_valid = torch.where(
+        nan_mask,
+        torch.zeros_like(x),
+        torch.ones_like(x),
+    ).sum(dim=axis)
+    value_sum = torch.where(nan_mask, torch.zeros_like(x), x).sum(dim=axis)
```
alanprior, line 60: @bejaeger reading through this, I was surprised we don't use `torch_nansum` in this function?

src/tabpfn/preprocessing/torch/ops.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    nan_mask = torch.isnan(x)
+    num_valid = torch.where(
+        nan_mask,
+        torch.zeros_like(x),
+        torch.ones_like(x),
+    ).sum(dim=axis)
+    value_sum = torch.where(nan_mask, torch.zeros_like(x), x).sum(dim=axis)
+
+    mean = value_sum / num_valid.clamp(min=1.0)
```
alanprior, line 83: @bejaeger same as above. shouldn't we directly use `torch_nansum` and `torch_nanmean` here?

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
+#  Copyright (c) Prior Labs GmbH 2025.
```
alanprior, line 1: nit: we are definitely not consistent re copyright thing

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+import dataclasses
+from typing_extensions import override
+
+import torch
+
+from tabpfn.preprocessing.datamodel import FeatureModality
+
+
+class TorchPreprocessingStep(abc.ABC):
+    """Base class for preprocessing steps that operate on specific columns.
```
alanprior: @bejaeger are we by definition limited over columns only?

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+    transformation logic. The base class handles column selection, tensor
+    cloning, and reassignment.
+    """
+
+    def fit_transform(
+        self,
+        x: torch.Tensor,
+        column_indices: list[int],
+        num_train_rows: int,
+        fitted_cache: dict[str, torch.Tensor] | None = None,
```
alanprior, line 33: @bejaeger I found this name confusing. `cache` is too much entangled with the kv cache IMO.

Also, I was expecting that all steps would just pass one input to the next, and we'd stop caring about the initial `x`. I guess this doesn't happen but would love to have some examples of why we need two inputs?

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+        Args:
+            x: Tensor of selected columns [num_train_rows, batch_size, num_cols].
+
+        Returns:
+            Cache dictionary with the cache for the transform step.
+        """
+        ...
+
+    @abc.abstractmethod
+    def _transform(
```
alanprior, line 79: @bejaeger do we have cases that change the existing modality? The easiest, currently unexisting case I can think of is a string given as date, then doing skrub's DateEncoder or such so now the inputs are numerical, but you don't want the original date anymore.

This might not be in torch though. Anyway, maybe for later

src/tabpfn/preprocessing/torch/pipeline_interface.py
alanprior: I'm a bit unsure about what's happening with the cache and how it is implemented tbh. Also, the word `cache` confuses me a bit as it has multiple potential meanings which I'm not sure what we want (e.g., fit vs transform, cache between every step, cache between model calls like the kv cache...).

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+            steps: List of (step, modalities) where modalities is a set of
+                FeatureModality values the step should be applied to.
+            keep_fitted_cache: Whether to keep the fitted cache between calls.
+                If True, the fitted cache will be reused in the transform step.
+                If False, the fitted cache will be fit on the training data for each
+                step.
+        """
+        super().__init__()
+
+        for step in steps:
```
alanprior: @bejaeger I guess that we are in purpose no passing objects here? Anyway, style, but I would wrap it as self._validate_steps(steps)

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+        x: torch.Tensor,
+        metadata: ColumnMetadata,
+        num_train_rows: int | None = None,
+        *,
+        use_fitted_cache: bool = False,
+    ) -> TorchPreprocessingPipelineOutput:
+        """Apply all steps to the input tensor.
+
+        Args:
+            x: Input tensor [num_rows, batch_size, num_columns] or
```
alanprior, line 144: nit: This notation is not super consistent with our usual `R`, `C/F`

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+                [num_rows, num_columns]. If 2D, a batch dimension is added
```
alanprior, line 145: @bejaeger preprocessing for all batches at the same time sounds painful. Imagine a step adds an extra column for some datasets at one step, but doesn't for others. You'll need to be controlling the padding in a very annoying way. Unsure if it's realistic in torch setup, but this would definitely happen in real cases (preprocessing).

src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -0,0 +1,291 @@
...
+                "use_fitted_cache=True is only supported if keep_fitted_cache=True "
+                "during initialization."
+            )
+
+        squeeze_batch_dim = False
+        if x.ndim == 2:
+            x = x.unsqueeze(1)
+            squeeze_batch_dim = True
+
+        num_columns = x.shape[-1]
```
alanprior, line 164: @bejaeger again style but I really prefer these validations to be one-liners so they don't distract the reader from the main logic

src/tabpfn/utils.py
```diff
@@ -379,35 +378,7 @@ def update_encoder_params(
             be differentiable with pt autograd. This disables non-differentiable
             encoder steps.
     """
-    if remove_outliers_std is not None and remove_outliers_std <= 0:
```
alanprior, line 382: @bejaeger unclear what happens here but so nice to see this being removed from here

src/tabpfn/preprocessing/torch/steps.py
```diff
@@ -0,0 +1,67 @@
...
+from tabpfn.preprocessing.torch.torch_remove_outliers import TorchRemoveOutliers
+from tabpfn.preprocessing.torch.torch_standard_scaler import TorchStandardScaler
+
+if TYPE_CHECKING:
+    import torch
+
+    from tabpfn.preprocessing.datamodel import FeatureModality
+
+
+class TorchStandardScalerStep(TorchPreprocessingStep):
```
alanprior, line 20: @bejaeger this is kind of a major comment seeing all this PR, but while I understand the pytorch preprocessing motivation, it still feels a bit dangerous. What happens if down the line you want to do non-torch operations that must happen after preprocessing?

src/tabpfn/preprocessing/torch/steps.py
```diff
@@ -0,0 +1,67 @@
...
+    def _transform(
+        self,
+        x: torch.Tensor,
+        fitted_cache: dict[str, torch.Tensor],
+    ) -> tuple[torch.Tensor, torch.Tensor | None, FeatureModality | None]:
+        """Transform columns using the fitted scaler."""
+        return self._scaler.transform(x, fitted_cache=fitted_cache), None, None
+
+
+class TorchRemoveOutliersStep(TorchPreprocessingStep):
```
alanprior: It feels like this step is very tied to StandardNormalization; the name is a bit too generic. For example, if we do MaxMin preprocessing or Quantiles, the logic would be different.

src/tabpfn/preprocessing/torch/torch_remove_outliers.py
```diff
@@ -0,0 +1,136 @@
+"""Torch implementation of outlier removal with NaN handling."""
+
+from __future__ import annotations
+
+import torch
+
+from tabpfn.preprocessing.torch.ops import torch_nanmean, torch_nanstd
+
+
+class TorchRemoveOutliers:
```
alanprior: The word removing is ambiguous: it wasn't clear to me after carefully reading the function whether we are going to remove these rows completely, or put Nulls instead their values, or clamp them.

src/tabpfn/preprocessing/torch/torch_standard_scaler.py
```diff
@@ -0,0 +1,88 @@
...
+        Args:
+            x: Input tensor with shape [T, ...] where T is the number of rows.
+
+        Returns:
+            Cache dictionary with the cache for the transform step.
+        """
+        mean = torch_nanmean(x, axis=0)
+        std = torch_nanstd(x, axis=0)
+
+        # Handle constant features (std=0) by setting std to 1
```
alanprior, line 29: @bejaeger Probably a nice guardrail although we should never really encounter such cases? We are going to pre-filter constant features. It's an existing `FeatureModality`.

src/tabpfn/preprocessing/torch/torch_standard_scaler.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        if "mean" not in fitted_cache or "std" not in fitted_cache:
+            raise ValueError("Invalid fitted cache. Must contain 'mean' and 'std'.")
+
+        mean = fitted_cache["mean"]
+        std = fitted_cache["std"]
+        x = (x - mean) / (std + 1e-16)
+
+        # Clip very large values
+        return torch.clip(x, min=-100, max=100)
```
alanprior, line 59: These numbers are a bit surprising, also unsure if they should be hardcoded. Isn't this directly related to `TorchRemoveOutliers`?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -0,0 +1,54 @@
...
+
+from __future__ import annotations
+
+import dataclasses
+from enum import Enum
+
+import pandas as pd
+
+
+class FeatureModality(str, Enum):
```
alanprior, line 11: @bejaeger thanks for moving them away! looking good

src/tabpfn/preprocessing/ensemble.py
alanprior: Right, so seems like @bejaeger is aiming that we could decide whether we are going to have any, or, put the torch steps as part of `SequentialFeatureTransformer`?

bejaeger: Hm...I was wondering if I should use torch_preprocessor but I liked the separation between `cpu` and `gpu`. Since we should always aim to run the `torch` pipeline on GPU I think `gpu` fits well here?

RE `SequentialFeatureTransformer`: This will be deprecated in favor of this new, more flexible pipeline.

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: You're right that `cache` is a very general term. Since this is only used internally of the preprocessing pipeline and I didn't want to use the word `state` (because I did not want to confuse it with something that is passed through the pipeline as the "data") I think it's okay here?

src/tabpfn/preprocessing/torch/ops.py
bejaeger: Good point. I used the old implementation and didn't want to change functionality. torch_nanmean also handles infs, torch_nansum doesn't.

src/tabpfn/preprocessing/torch/ops.py
bejaeger: Since we need the nan_mask anyway we would calculate it twice if we run the current `torch_nanmean` Would need to also parse the `nan_mask` which might then introduce more complexity? Unsure but I think it's okay to keep the old code.

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Added a "can" to "can operate". If there is a usecase for row-based transformations there should be nothing stopping us to do it in this pipeline

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Not sure what you mean by two inputs?
if it is `x` and `column_indices` the idea was to stay "compile-friendly" for these preprocessing classes. Having another wrapper might complicate things for the compiler.
Since it's only two inputs I think it's clean enough?

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: The text processing needs to happen before this pipeline, yup.

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Hm...it's a tradeoff. I didn't want to introduce yet another dataclass for this and I think the readability is okay like that. Added a `validate_steps` function.

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: I think it's fine because the context is clearly separated. Model and preprocessing. In the model we deal with a lot of shapes so having shortcuts for them makes more sense. Here, I think being more descriptive adds to the readability.

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Yeah, that's a very good point we should keep in mind. In the current pipeline it's not possible. For now, we don't have any steps that add features so should be okay.

src/tabpfn/preprocessing/torch/steps.py
bejaeger: Discussed internally, thanks

src/tabpfn/preprocessing/torch/steps.py
bejaeger: I agree! I would say this is the job of the user of the pipeline to decide how to use it?

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Done!

src/tabpfn/preprocessing/torch/torch_remove_outliers.py
bejaeger: Yes, good point! Unfortunately, we use `OUTLIER_REMOVAL_STD` in the config for this. However, that shouldn't hold us back from updating the name of this function! Called in `TorchSoftClipOutliers`

src/tabpfn/preprocessing/torch/torch_standard_scaler.py
bejaeger: I agree this is not the cleanest. It's just a "last resort" to avoid breaking the model when there are really large values for some reason. Which there shouldn't be anyways. This was used in the previous implementation so I would leave it as is.

bejaeger:
Thanks for the reviews! Adressed all of the comments above.

src/tabpfn/preprocessing/torch/pipeline_interface.py
alanprior: No, I'm fine with x and column_indices. I was unsure about why we need x and fitted_cache, probably as I misunderstood what it supposed to be.

src/tabpfn/preprocessing/torch/torch_standard_scaler.py
alanprior: Hmmm. Ideally if this piece of code is completely legacy and should never happen, I would aim to remove it? It just seems like something that doesn't belong. A safer way to assure that would be to log an event every time this is relevant. If no-one triggers this during training/inference, then probably it shouldn't exist.

src/tabpfn/preprocessing/torch/steps.py
alanprior: Right. I think you ended up changing the name which makes things better for that purpose as well.

src/tabpfn/preprocessing/torch/pipeline_interface.py
alanprior: It's fine. Reading through this I was unsure what exactly we are holding there but I trust it's fine. We'll bump into this code again.

PR #762, opened 2026-01-29
Add .python-version for Dependabot Python 3.14 resolution

LeoGrin:
## Summary
- Add `.python-version` file set to `3.14` so Dependabot resolves dependencies against Python 3.14

## Why
Dependabot uses the minimum Python version (3.9) from `requires-python` to evaluate dependency updates. This causes it to filter out packages that dropped Python 3.9 support (e.g. pandas 3.0 requires Python >= 3.11), resulting in "No update possible" even when the `widen_ranges` strategy should widen the upper bound.

With `.python-version` set to 3.14, Dependabot will see these newer versions as compatible and propose range widens. CI still tests both Python 3.9 and 3.14, so actual compatibility is verified on the PR.

## Test plan
- [x] Verify `.python-version` contains `3.14`
- [ ] After merge, trigger a Dependabot run and confirm it no longer filters out packages like pandas 3.0

---

> [!NOTE]
> **Low Risk**
> Adds a single repo config file and does not change runtime code paths; the main risk is only that some tooling may start defaulting to Python 3.14 locally.
> 
> **Overview**
> Adds a `.python-version` file pinned to `3.14` so tooling (notably Dependabot) resolves and evaluates Python dependencies against 3.14 rather than the minimum `requires-python` value, enabling updates that drop older Python support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a41adb1ad4c28f287430b88124eb46bdcc9ea744. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


LeoGrin:
Closing in favor of switching to Renovate, which doesn't filter by Python version by default. See replacement PR.

LeoGrin:
The issue is that to set the python version for dependabot resolution the only way AFAIK is to add a .python-version file which also impacts uv etc.

PR #764, opened 2026-01-29
remove all upper version limits

LeoGrin:
Context:

Right now we have an older version which didn't specify dependencies upper limits, so when a user install a new version of a dependencies which is above our current upper limit, they would get this older version, which is very confusing. Now they would get the newest version and a bug, which is probably less confusing?

Also as our CI tests the highest version possible, this means we would fail quickly when new incompatible versions come out. This should already have been caught with dependabot but wasn't always (TLDR: didn't work if the new version of the dependency didn't support our minimum python version = 3.9).

One potential downside of this PR: If we don't remove the upper limit, we might one day be able to yank the old versions without upper limits once we think they're not used anywhere, which would make these upper limit "work". After this PR we won't really be able to put back upper version limit safely if we ever want to. I think this is fine because we probably don't want to yank these old versions anyway but wanted to flag.

A bit more context on _sklearn_compat, for which we had to update our vendored version (it's used to be able to support multiple sklearn versions at the same time): https://github.com/sklearn-compat/sklearn-compat

## Issue

## Motivation and Context

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?

---

## Checklist

-   [ ] The changes have been tested locally.
-   [ ] The code follows the project's style guidelines.

---

> [!NOTE]
> **Medium Risk**
> Removing upper bounds can pull in future major releases (notably `scikit-learn`) with breaking behavior; the vendored sklearn compat layer is updated to mitigate some version drift but may not cover all edge cases.
> 
> **Overview**
> Removes *upper version limits* from core runtime dependencies in `pyproject.toml` (e.g., `torch`, `numpy`, `scikit-learn`, `scipy`, `pandas`, `einops`, `huggingface-hub`), allowing installs to resolve to newer major versions.
> 
> Updates the vendored `src/tabpfn/misc/_sklearn_compat.py` to `sklearn-compat` `0.1.4`, including new shims for scikit-learn `<1.8` (dataframe/series detection across pandas/polars/pyarrow and `_check_targets` sample-weight handling) plus small compatibility cleanups; adds a changelog entry noting the dependency bound change.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f60e5b04bb3d45f32217e80cfc078167d6cd9d99. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/changelog/764.changed.md
+++ b/changelog/764.changed.md
@@ -1,0 +1,1 @@
+Remove upper version limits on dependencies


brendan-priorlabs:
> After this PR we won't really be able to put back upper version limit safely if we ever want to.

This seems like a problem. What happens when a library comes out with a breaking change and we need to at least temporarily pin away from it?

LeoGrin:
> This seems like a problem. What happens when a library comes out with a breaking change and we need to at least temporarily pin away from it?

yeah it is a bit annoying but solving it would require yanking older tabpfn versions, which I don't think we want to do anyway?
Right now we already have this issue but it is hidden, as users with this new library version would just get an old tabpfn version.

brendan-priorlabs:
Fair point about the problem existing regardless. But wait... I think this PR is the solution too! After we release, if we do pin again in the future, then at least users will be able to get this version instead of the super old ones

brendan-priorlabs (approved):
I'll approve on that basis, lmk if the reality is drastically different

LeoGrin:
> Fair point about the problem existing regardless. But wait... I think this PR is the solution too! After we release, if we do pin again in the future, then at least users will be able to get this version instead of the super old ones

Yeah fully agree :) (with the caveat that if we add a pin because of a breaking change the user would get the tabpfn version just before the pin, but would still get a weird error instead of a "this version is not supported" error). But just makes it harder to go for the nuke "yanking old versions and putting back long term major version upper limit" solution because we would have to yank pretty new solutions.

PR #767, opened 2026-01-30
Introduce feature schema and use in preprocessing pipeline

bejaeger:
- Preprocessing code didn't allow to handle modalities other than categoricals and (implicitly) numericals. This PR introduces a feature schema object that's passed along the pipelines, together with an overhaul of the pipeline interfaces.
- Adds many tests for individual preprocessing steps
- Also introduces a small `TabPFNLabelEncoder` class for convenience and testing.

Functionality is unchanged with this PR

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?

---

## Checklist

-   [X] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.

---

### DIFF
--- a/src/tabpfn/preprocessing/pipeline_interface.py
+++ b/src/tabpfn/preprocessing/pipeline_interface.py
@@ -93,6 +93,12 @@ class PreprocessingStep:
       main output.
     """
 
+    n_added_columns_: int | None
+    """Number of added columns from `_transform`. Set during first transform call."""
+
+    modality_added_: FeatureModality | None
+    """Modality of added columns from `_transform`. Set during first transform call."""
+
     @abstractmethod
     def _fit(
         self,
@@ -162,6 +168,7 @@ class PreprocessingStep:
             PreprocessingStepResult with transformed data and feature schema.
         """
         result, X_added, modality_added = self._transform(X, is_test=is_test)
+        self._validate_added_data(X_added=X_added, modality_added=modality_added)
 
         return PreprocessingStepResult(
             X=result,
@@ -169,6 +176,30 @@ class PreprocessingStep:
             X_added=X_added,
             modality_added=modality_added,
         )
+
+    def _validate_added_data(
+        self,
+        X_added: np.ndarray | None,
+        modality_added: FeatureModality | None,
+    ) -> None:
+        """Validate consistency of added columns across train/test transforms."""
+        n_added = X_added.shape[1] if X_added is not None else None
+
+        store_expected_values_on_first_call = not hasattr(self, "n_added_columns_")
+        if store_expected_values_on_first_call:
+            self.n_added_columns_ = n_added
+            self.modality_added_ = modality_added
+        else:
+            if n_added != self.n_added_columns_:
+                raise ValueError(
+                    f"Inconsistent number of added columns: expected "
+                    f"{self.n_added_columns_}, got {n_added}"
+                )
+            if modality_added != self.modality_added_:
+                raise ValueError(
+                    f"Inconsistent modality for added columns: expected "
+                    f"{self.modality_added_}, got {modality_added}"
+                )
 
 
 class PreprocessingPipeline:
@@ -264,6 +295,10 @@ class PreprocessingPipeline:
         Returns:
             Tuple of (transformed array, updated feature schema).
         """
+        # Single copy to preserve immutability for the caller, avoiding N copies
+        # inside the loop for steps that target specific modalities.
+        X = X.copy() if isinstance(X, np.ndarray) else X.clone()
+
         for step, modalities in self.steps:
             if modalities:
                 indices = feature_schema.indices_for_modalities(modalities)
@@ -287,7 +322,6 @@ class PreprocessingPipeline:
                         f"modalities must return the same number of columns."
                     )
 
-                X = X.copy() if isinstance(X, np.ndarray) else X.clone()
                 assert X.dtype == result.X.dtype
                 X[:, indices] = result.X
 


bejaeger:

- #756
    - #767 ◀

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -52,3 +69,147 @@ def x_txt(self) -> pd.DataFrame:
 
     def _get_modality(self, modality: FeatureModality) -> pd.DataFrame:
         return self.X.loc[:, self.columns_by_modality[modality]]
+
+
+@dataclasses.dataclass
+class FeatureSchema:
```
alanprior, line 42: @bejaeger It's not super clear to me whether this is different from `DatasetView`. I understand that will allow us to be backwards compatible, but it seems like it's playing the same role? If so, wouldn't it be best to rename it to `LegacyDatasetView` or something like that?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -52,3 +69,147 @@ def x_txt(self) -> pd.DataFrame:
...
+    this mapping as preprocessing steps transform the data.
+
+    Attributes:
+        features: List of Feature objects where index = column position.
+    """
+
+    features: list[Feature] = dataclasses.field(default_factory=list)
+
+    @classmethod
+    def from_only_categorical_indices(
```
alanprior, line 57: @bejaeger is this supposed to replace `infer_categorical_features`? If so, this logic is much simplified?

src/tabpfn/finetuning/data_util.py
```diff
@@ -430,12 +431,16 @@ def __getitem__(self, index: int) -> ClassifierBatch | RegressorBatch:  # noqa:
         else:
             y_train = y_train_raw
 
+        num_columns = x_train_raw.shape[1]
+        feature_schema = FeatureSchema.from_only_categorical_indices(
+            cat_ix, num_columns
+        )
         itr = fit_preprocessing(
```
alanprior: Unclear to me what happens in these lines and what is `feature_schema_preprocessed`. It doesn't seem to be defined anywhere. At least with the github view it looks weird

Image

src/tabpfn/finetuning/data_util.py
```diff
@@ -479,14 +484,18 @@ def __getitem__(self, index: int) -> ClassifierBatch | RegressorBatch:  # noqa:
         x_test_raw = torch.from_numpy(x_test_raw)
         y_test_raw = torch.from_numpy(y_test_raw)
 
+        cat_indices = [
+            m.indices_for(FeatureModality.CATEGORICAL)
+            for m in feature_schema_preprocessed
+        ]
         # Return structured batch data using dataclasses for clarity
         if is_regression_task:
             return RegressorBatch(
```
alanprior, line 493: @bejaeger shouldn't we kill the `cat_indices` use inside?

src/tabpfn/preprocessing/datamodel.py
alanprior: Ok, we discussed internally that this is some mutation of it and we should remove `DatasetView`

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -23,8 +24,19 @@ def _get_least_common_category_count(x_column: np.ndarray) -> int:
     return int(counts.min())
 
 
-class EncodeCategoricalFeaturesStep(FeaturePreprocessingTransformerStep):
-    """Encode Categorical Features Step."""
+class EncodeCategoricalFeaturesStep(PreprocessingStep):
+    """Encode categorical features using ordinal or one-hot encoding.
+
+    When using with PreprocessingPipeline, register as a bare step (no modalities):
```
alanprior, line 30: Unclear why?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -52,3 +69,147 @@ def x_txt(self) -> pd.DataFrame:
...
+        cls,
+        categorical_indices: list[int],
+        num_columns: int,
```
alanprior, line 60: nit: I find it a bit more natural to pass X here, and cleaner from outside.

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -52,3 +69,147 @@ def x_txt(self) -> pd.DataFrame:
...
+        """
+        numerical_indices = [
+            i for i in range(num_columns) if i not in categorical_indices
+        ]
+        if not numerical_indices and not categorical_indices:
+            return cls(features=[])
+
+        features: list[Feature | None] = [None] * num_columns
+        for idx in categorical_indices:
+            features[idx] = Feature(name=None, modality=FeatureModality.CATEGORICAL)
```
alanprior, line 76: Using None feels a bit weird here and prone to errors, especially if we want to later use columns names (even if uninformative). I would suggest in such case to default into a more interpretable "col_{idx}" format or so.

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -52,3 +69,147 @@ def x_txt(self) -> pd.DataFrame:
...
+        return [f.name for f in self.features]
+
+    @property
+    def num_columns(self) -> int:
+        """Get the total number of columns."""
+        return len(self.features)
+
+    def indices_for(self, modality: FeatureModality) -> list[int]:
+        """Get column indices for a single modality."""
+        return [i for i, f in enumerate(self.features) if f.modality == modality]
```
alanprior, line 94: @bejaeger this feels a bit redundant. shouldn't we use just `indices_for_modalities`? I suggest the latter could get Iterable[FeatureModality] | FeatureModality and wrap inside for convenience.

src/tabpfn/preprocessing/steps/nan_handling_polynomial_features_step.py
```diff
@@ -9,13 +9,14 @@
...
-    FeaturePreprocessingTransformerStep,
+from tabpfn.preprocessing.datamodel import FeatureModality, FeatureSchema
+from tabpfn.preprocessing.pipeline_interface import (
+    PreprocessingStep,
 )
 from tabpfn.utils import infer_random_state
 
 
-class NanHandlingPolynomialFeaturesStep(FeaturePreprocessingTransformerStep):
+class NanHandlingPolynomialFeaturesStep(PreprocessingStep):
```
alanprior, line 19: @bejaeger is this something we still use?

One things that comes to my mind is that different versions of models will have specific preprocessors and specific layers. I'm unsure how much backwards compatible we plan to be. But if there are things which belong to older versions (and them only), it might be interesting in the future to move them aside a bit, just to have less clutter. Anyway, outside of the scope of this PR, but something worth having in mind.

Imagine us releasing V3, V4... but still supporting some ugly old bunch of code. So arguably now we have v2.5, how much do we care about v2? Could we have its "old" pieces moved to a dedicated folder? This would help much to find our way into the code and flow, expecting where to find files etc.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
```diff
@@ -2,18 +2,22 @@
...
+from tabpfn.preprocessing.pipeline_interface import (
+    PreprocessingStep,
 )
 
+if TYPE_CHECKING:
+    from tabpfn.preprocessing.datamodel import FeatureModality, FeatureSchema
 
-class RemoveConstantFeaturesStep(FeaturePreprocessingTransformerStep):
+
+class RemoveConstantFeaturesStep(PreprocessingStep):
```
alanprior, line 20: @bejaeger this confuses me. I don't have the full flow in mind but there are two things i don't understand:

1. The modality detector already detects constants. Why still using this? Is it some backwards compatible thing? I thought we are going to make this function go away.
2. The "removed indices" feels a bit of a waste, because we'll keep dragging columns in X which we'll remove. but probably code-wise it's much easier than having to change all the indices.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
```diff
@@ -36,15 +42,13 @@ def _fit(  # type: ignore
...
+        return feature_schema.remove_columns(removed_indices)
 
     @override
     def _transform(
         self, X: np.ndarray | torch.Tensor, *, is_test: bool = False
-    ) -> np.ndarray:
+    ) -> tuple[np.ndarray, np.ndarray | None, FeatureModality | None]:
         assert self.sel_ is not None, "You must call fit first"
-        return X[:, self.sel_]
+        return X[:, self.sel_], None, None
```
alanprior, line 54: it's a bit hard to know exactly what is this and remember the structure of the 3 objects that we return. Maybe we could call these objects in a certain way, so instead of the signature being:

`tuple[np.ndarray, np.ndarray | None, FeatureModality | None]`

it could be something like 'tuple[X, AddedCol]'... or something. Unsure if that's helpful. But I feel that an object here would have been a much better choice TBH

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -126,8 +127,35 @@ def _skew(x: np.ndarray) -> float:
     return float(3 * (np.nanmean(x, 0) - np.nanmedian(x, 0)) / np.std(x, 0))
 
 
-class ReshapeFeatureDistributionsStep(FeaturePreprocessingTransformerStep):
-    """Reshape the feature distributions using different transformations."""
+class ReshapeFeatureDistributionsStep(PreprocessingStep):
+    """Reshape feature distributions using various transformations.
+
+    This step should receive ALL columns (not modality-sliced) because it:
+    1. Handles feature subsampling when too many features exist
```
alanprior, line 134: @bejaeger interesting and a bit unexpected for me, why and how does it happen? I thought we had a max capping?

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -126,8 +127,35 @@ def _skew(x: np.ndarray) -> float:
     return float(3 * (np.nanmean(x, 0) - np.nanmedian(x, 0)) / np.std(x, 0))
 
 
-class ReshapeFeatureDistributionsStep(FeaturePreprocessingTransformerStep):
-    """Reshape the feature distributions using different transformations."""
+class ReshapeFeatureDistributionsStep(PreprocessingStep):
+    """Reshape feature distributions using various transformations.
+
+    This step should receive ALL columns (not modality-sliced) because it:
```
alanprior, line 133: @bejaeger I read the docstrings and I'm unconvinced. It feels like a monster function that could have been broken down. But I didn't read the logic too carefully

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -126,8 +127,35 @@ def _skew(x: np.ndarray) -> float:
     return float(3 * (np.nanmean(x, 0) - np.nanmedian(x, 0)) / np.std(x, 0))
 
 
-class ReshapeFeatureDistributionsStep(FeaturePreprocessingTransformerStep):
-    """Reshape the feature distributions using different transformations."""
+class ReshapeFeatureDistributionsStep(PreprocessingStep):
```
alanprior, line 130: @bejaeger I don't like this function at all. When implementing a new modality, e.g text, I'll need to be super careful with these types of functions. Maybe there's no choice, but I feel that this is exactly the things I would love to avoid - I'm missing here some "main stream" which a feature can go directly to the end without going through these manipulations

src/tabpfn/preprocessing/steps/shuffle_features_step.py
```diff
@@ -53,16 +56,14 @@ def _fit(
         else:
             self.index_permutation_ = index_permutation
 
-        return [
-            new_idx
-            for new_idx, idx in enumerate(index_permutation)
-            if idx in categorical_features
-        ]
+        return feature_schema.apply_permutation(index_permutation)
```
alanprior, line 59: @bejaeger well done! quite elegant

src/tabpfn/preprocessing/clean.py
alanprior, line 66: @bejaeger I would have expected this function to change a lot, and also to get the detected features and not cat indices. for example if there are constant or text features, what would happen here? I'm fundamentally missing something

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -151,8 +140,24 @@ def _detect_numeric_as_categorical(
...
+    try:
+        return s.nunique(dropna=False)
+    except TypeError as e:
+        # The sklearn test is inserting a dict ({"foo": "bar"}) into the data to verify
+        # that the estimator raises a TypeError with a specific message pattern
+        # ("argument must be .* string.* number"). However, when pandas tries to
+        # compute nunique() on a Series containing a dict, it fails with "unhashable
+        # type: 'dict'" which doesn't match sklearn's expected error pattern.
+        raise TypeError(
+            f"argument must be a string or a number (columns must only contain strings "
```
alanprior, line 148: @bejaeger now sure what is this, but what about dates?

src/tabpfn/base.py
```diff
@@ -350,7 +351,7 @@ def create_inference_engine(  # noqa: PLR0913
         return InferenceEngineBatchedNoPreprocessing(
             X_trains=X_train,  # pyright: ignore[reportArgumentType]
             y_trains=y_train,  # pyright: ignore[reportArgumentType]
-            cat_ix=cat_ix,  # pyright: ignore[reportArgumentType]
+            feature_schema=feature_schema,  # pyright: ignore[reportArgumentType]
```
alanprior, line 354: What are these ugly # pyright: ignore[reportArgumentType]?

alanprior (approved):
I'm approving not to block anything, but I must say that it was very hard to follow, the PR was too big. The fatigue as a reviewer is felt. Also, I think that I have missed a longer summary of what the PR is doing, what are the main decisions taken etc. Probably my fault but I just randomly start reading the changes, which might not be the best way to review.

There are a few things I feel are worth noting:

1. It was unclear to me why `cat_indices` still lives and is passed quite a lot in the code. I would have expected this term to go away completely. At the same time, some functions seem to ignore the fact that new modalities are being introduced.
2. There are at least two "monster functions" which I believe should have been broken down. I'm unsure whether it's intentional or out of scope of this PR.
3. The code is "too much". We have so many utilities and stuff we do, sometimes optionally... To the very least, it would be cool to "hide" away code that belongs to optional features or older versions, just so the main common flow is visible from all the clutter. 

Ofc these comments are a bit vague and general, but I literally put a few hours into this review and I'm unsure how helpful it really is. I also did not review all of it - skipped the tests completely and skimmed through the last files. Obviously this is not your fault but simply a combination of the code being complex, the logic being rich, me still not knowing it enough.

src/tabpfn/finetuning/data_util.py
bejaeger: yeah it's confusing formatting. It comes from `) = list(zip(*itr))` 
Renamed `itr` to `preprocessing_iterator`

src/tabpfn/finetuning/data_util.py
bejaeger: Yes we should propagate the changes to the finetuning pipeline. Since there would be a whole bunch of more files touched, I put this on our list of things to do and would tackle it in a follow up

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
bejaeger: Just backwards compatibility. One-hot encoding wouldn't work without further changes.
Something we hopefully get around harmonizing going forward

src/tabpfn/preprocessing/steps/nan_handling_polynomial_features_step.py
bejaeger: Yes, this is still an option in our inference config. It's just off per default.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
bejaeger: The problem is that we need the info during `predict` as well and then the pipeline is the safest thing because we run it during `fit` and `predict`. We could simply use the `feature_schema` information in this `RemoveConstantFeaturesStep` which would be clean. What's preventing this atm is that we support `torch.Tensor` here which doesn't run the `cleaning` first. We can do it once we decouple completely the torch.Tensor path here.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
bejaeger: Hmm...I partially agree. It's a bit of a tradeoff, though. If we have a typevar of `AddedCol` it would also not be not immediately clear what this is and whether e.g. it's a `torch.Tensor`, an `np.ndarray` and whether it can be null. You would still need to hover over that typevar to see it.

The docstring of the base class should explain the return type so when hovering over the function the Return values should be explained to make sense out of them?

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: Yep, we need to decouple this. I started it in follow-up PRs already. This step should be configured to ONLY receive "numericals" (or potentially "categoricals"). Then we can exactly do what you say. E.g. text features not being touched by anything.

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: I should have added a clear TODO here. (Or avoid adding this doc string in the first place).
I didn't want to make the PR even larger. This will be decoupled in follow-up PRs

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: that's exactly what's used here

src/tabpfn/preprocessing/clean.py
bejaeger: I consciously didn't update this function. This PR is about the pipeline interface.
Thanks for the pointer. This function would need to change for supporting text features.

src/tabpfn/preprocessing/datamodel.py
bejaeger: I'd rather keep it at `None` and add a default whenever we need a string. This way, we know exactly whether the feature name was "user provided" or not. If we add a default here we wouldn't know downstream whether the feature name was a user provided "col_1" or a default.

src/tabpfn/preprocessing/datamodel.py
bejaeger: if we pass X we would only use it for getting `.shape[1]` and therefore make an assumption about the shape of X. Also it's unclear from the outside whether X is used in some other way internally. Passing `num_columns` makes it clear that the content of `X` is not needed, just the shape.

src/tabpfn/preprocessing/datamodel.py
bejaeger: This is a function for backwards compatibility, as mentioned in the docstring. 
There are still parts in the pipeline I didn't change and I think it's useful to keep this for now.

src/tabpfn/preprocessing/datamodel.py
bejaeger: `indices_for("categoricals")` reads quite nice and is shorter?
I'd rather have a single function get a single type of input rather than if/else logic inside functions. IMO makes it easier to read.

src/tabpfn/preprocessing/modality_detection.py
bejaeger: This is a sklearn thing. A date should also be either a string or a number? 
It's just rephrasing the error message so our sklearn interface tests are happy.

src/tabpfn/base.py
bejaeger: Just to silence ruff. The typing is not ideal in the `create_inference_engine` function but nothing I introduced here. Would be a bigger refactor to make this proper that's not worth it atm I think.

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
bejaeger: I did refactor the steps a bit during data cleaning prompted by this comment!

bejaeger:
Makes total sense what you say. It was hard to completely avoid a big PR for this type of work. 

The problem is that with small changes the intend of it may be unclear, and with bigger changes things are hard to review.
I decided to change the interface in this PR and leaving existing functionality as is, to avoid making this an even bigger PR. Most of the added lines are actually adding tests, the rest is changing the interface and refactoring.

A lot of your comments were on existing functionality or code that I actually haven't touched (except maybe adding doc strings). They'll be definitely useful to keep in mind for follow-ups, so thanks a lot for the review!

src/tabpfn/preprocessing/datamodel.py
alanprior: Right... it's a bit surprising though as its neither the new function, nor the old big one. I understand it probably represents some minor par of the code.

src/tabpfn/finetuning/data_util.py
alanprior: Ok, think it could be written a bit more readable but it is out of scope.

src/tabpfn/preprocessing/datamodel.py
alanprior: Hmmm. Debatable but fair

src/tabpfn/preprocessing/steps/nan_handling_polynomial_features_step.py
alanprior: Ok. The high level comment is something worth keeping in mind though, for any v2 pieces floating around

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
alanprior: Happy to hear that it helped a bit :)
Right, so IIUC in the future we will not need this as feature_schema will know from the detection what is constant and won't propagate it?

alanprior:
> Makes total sense what you say. It was hard to completely avoid a big PR for this type of work.
> 
> The problem is that with small changes the intend of it may be unclear, and with bigger changes things are hard to review. I decided to change the interface in this PR and leaving existing functionality as is, to avoid making this an even bigger PR. Most of the added lines are actually adding tests, the rest is changing the interface and refactoring.
> 
> A lot of your comments were on existing functionality or code that I actually haven't touched (except maybe adding doc strings). They'll be definitely useful to keep in mind for follow-ups, so thanks a lot for the review!

Yeah makes sense. I think that this PR's scope was a bit unclear for me. Also, it's crazy how much work was done 😅 Nice job

src/tabpfn/preprocessing/modality_detection.py
alanprior: Hmm, there is a datetime object in pandas, so I'm insure about that sentence. We currently have no proper test-coverage for that to be honest

src/tabpfn/preprocessing/steps/remove_constant_features_step.py
alanprior: I would have sticked to an object here tbh, but I think there were good reasons not to, so let's move on :)

src/tabpfn/classifier.py
```diff
@@ -580,7 +581,9 @@ def _initialize_for_differentiable_input(
         )
 
         # Minimal preprocessing for prompt tuning
-        self.inferred_categorical_indices_ = []
+        n_features = X.shape[1]
+        features = [Feature(name=None, modality=FeatureModality.NUMERICAL)] * n_features
```
LeoGrin, line 592: nit: we could fail if self.categorical_features_indices is not empty here

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -23,32 +25,158 @@ class FeatureModality(str, Enum):
...
+        Args:
+            modality: The modality for the new columns.
+            num_new: Number of new columns to add.
+            names: Names for the new columns. If None, uses "added_0", "added_1", ...
+
+        Returns:
+            New FeatureSchema instance with added features.
+        """
+        if names is None:
+            names = [f"added_{i}" for i in range(num_new)]
```
LeoGrin: it feels maybe a bit dangerous, as if you append twice you would get duplicate names?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -23,32 +25,158 @@ class FeatureModality(str, Enum):
...
+    def remove_columns(self, indices_to_remove: list[int]) -> FeatureSchema:
+        """Return new schema with specified columns removed."""
+        remove_set = set(indices_to_remove)
+        return FeatureSchema(
+            features=[f for i, f in enumerate(self.features) if i not in remove_set]
+        )
 
-    def _get_modality(self, modality: FeatureModality) -> pd.DataFrame:
-        return self.X.loc[:, self.columns_by_modality[modality]]
+    def apply_permutation(self, permutation: list[int]) -> FeatureSchema:
```
LeoGrin, line 179: nit: should we check this is indeed a permutation? or change the name if we want to support broader things.

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+@dataclasses.dataclass
+class PreprocessingStepResult:
+    """Result of a feature preprocessing step.
+
+    Attributes:
+        X: Transformed array. For steps registered with specific modalities,
+            this is only the transformed columns (not the full array).
+            The shape should match the input shape unless columns are removed.
```
LeoGrin, line 24: I think if columns are removed checks below would fail right? "if result.X.shape[1] != len(indices):"

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -23,8 +24,20 @@ def _get_least_common_category_count(x_column: np.ndarray) -> int:
...
+        pipeline = PreprocessingPipeline(steps=[EncodeCategoricalFeaturesStep()])
+
+    NOT as a modality-targeted step:
+        pipeline = PreprocessingPipeline(steps=[
+            (EncodeCategoricalFeaturesStep(), {FeatureModality.CATEGORICAL})
+        ])
+
+    This is needed for the pipeline with onehot encoding to work.
+    It will be updated in future versions.
```
LeoGrin, line 39: does it mean using it as a modality-targeted step will work in the future for one-hot?

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -0,0 +1,357 @@
...
+import numpy as np
+import torch
+
+from tabpfn.preprocessing.datamodel import FeatureModality, FeatureSchema
+
+StepWithModalities: TypeAlias = tuple["PreprocessingStep", set[FeatureModality]]
+
+
+@dataclasses.dataclass
+class PreprocessingStepResult:
```
LeoGrin, line 18: a few questions about this:
- do I understand correctly that we separate the transformed X from the added X, so that we can have a 1-to-1 match between original X and transformed X column?
- what if I want to add a text embedding preprocessing step? I would do a step which returns an empty X, and then put the embeddings in X_added, and modality_added = NUMERICAL. Is that correct?
- what if I want to something like a rotation of a group of column. My understanding is that I should also return an empty X and only use X_added
- am I correct that the X result is reserved for transforms that operate on a column-wise basis? Maybe we can make this a bit clearer. I think it might be clearer to actually separate between column-wise transformers and "multi-column" transformers?

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -0,0 +1,357 @@
...
+
+        Args:
+            X: array of shape (n_samples, n_features). For steps registered
+                with specific modalities, this is only the relevant columns.
+            is_test: Whether this is test data (used for AddFingerPrint step).
+
+        Returns:
+            PreprocessingStepResult with transformed data and feature schema.
+        """
+        result, X_added, modality_added = self._transform(X, is_test=is_test)
```
LeoGrin, line 170: nit: not sure how to do that but I feel like there should be a check that this is consistent, i.e X_added has the same number of columns and modality is the same for each individual step. Maybe they could have a n_added_columns_ and modality_added_ attributes?

src/tabpfn/preprocessing/steps/nan_handling_polynomial_features_step.py
```diff
@@ -76,17 +84,22 @@ def _fit(self, X: np.ndarray, categorical_features: list[int]) -> list[int]:
                     continue
                 self.poly_factor_2_idx[i] = rng.choice(list(indices_))
 
-        return categorical_features
+        # Polynomial features are appended as new numerical columns
+        return feature_schema.append_columns(
```
LeoGrin: IIUC this is already handled by the pipeline no? with this we would count them twice?

changelog/767.changed.md
```diff
@@ -0,0 +1 @@
+Public method `ensure_compatible_fit_inputs()` now returns the name of the y data if present.
```
noahho: I'd expand, maybe sth like this (this is just LLM, so please improve)
```suggestion
Refactoring of the Preprocessing Pipeline
* **Preprocessing Architecture:** Overhauled the preprocessing pipeline to use a robust `FeatureSchema` system instead of passing raw lists of categorical indices.
    * Introduced `FeatureSchema`, `Feature`, and `FeatureModality` data models to track column metadata through transformations.
    * Introduced `PreprocessingPipeline` and `PreprocessingStep` interfaces for modular, modality-aware transformations.
    * Refactored `TabPFNClassifier` and `TabPFNRegressor` to use `FeatureSchema` for internal state tracking (replacing `inferred_categorical_indices_` with `inferred_feature_schema_`).
    * Updated all preprocessing steps (e.g., `AddFingerprintFeaturesStep`, `EncodeCategoricalFeaturesStep`) to the new interface.
* **Validation:** `ensure_compatible_fit_inputs()` now returns the original target name (if present) to support better label encoding metadata.
* **Label Encoding:** Introduced `TabPFNLabelEncoder` to centralize label validation and metadata extraction.
```

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -0,0 +1,357 @@
...
+
+                if result.X.shape[1] != len(indices):
+                    raise ValueError(
+                        f"Step {step.__class__.__name__} registered with modalities "
+                        f"{modalities} received {len(indices)} columns but returned "
+                        f"{result.X.shape[1]} columns. Steps registered with "
+                        f"modalities must return the same number of columns."
+                    )
+
+                X = X.copy() if isinstance(X, np.ndarray) else X.clone()
```
noahho: For every step in the pipeline that targets specific modalities, you are cloning the entire dataset X just to update a subset of columns. If a pipeline has 10 steps and the dataset is large (e.g., 500MB), you are allocating 5GB of memory and spending significant CPU time copying data that wasn't touched.

tests/test_preprocessing/test_pipeline_consistency.py
```diff
@@ -25,14 +25,18 @@
...
 
 try:
-    from tabpfn.preprocessing.datamodel import ColumnMetadata, FeatureModality
+    from tabpfn.preprocessing.datamodel import FeatureModality, FeatureSchema
 
     NEW_PIPELINE_IMPLEMENTATION = True
 except ImportError:
+    try:
+        from tabpfn.preprocessing.pipeline_interface import TransformResult
+    except ImportError:
```
noahho, line 38: Should this be removed?

src/tabpfn/preprocessing/pipeline_interface.py
noahho, line 1: The rename from interfaces to interface seems a bit unnecessary?

src/tabpfn/classifier.py
bejaeger: Good idea, Done!

src/tabpfn/preprocessing/datamodel.py
bejaeger: Hmm...I don't think we would have duplicated names. But thanks for spotting this. I did want to use `None` whenever the feature name is not "user provided" (or something we decide to use when we append certain types of features. So I changed this to
`chosen_names = [None] * num_new if names is None else names`

src/tabpfn/preprocessing/datamodel.py
bejaeger: Yes! Added a safeguard

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Yeah, I noticed that too. This is one of the reasons we still allow for steps to be registered in the pipeline without modalities. Then we simply take the transformed `X` as the new `X` for the next step. 
The plan is to get rid of these two modes; but haven't done it in this PR yet because it requires a whole bunch of other changes in several of the current steps.

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
bejaeger: That's the goal; passing all modalities to `encode_categorical.*` doesn't really make much sense anyway.

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: I like the idea of separating `column-wise` and `multi-column` transformers. Not sure if we need to separate their API's, but at least logically that's a good way of thinking about it.

As mentioned, right now we're in a bit of a hybrid state, where we can register steps with and without modalities. This should definitely be better documented and made clearer.

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Hmm...not sure what you mean by "modality is the same for each individual step?". The step decided internally what modality it returns. Right now, each step can only add columns that have the same modality, so modality_added is just the feature modality. `X_added` can actually have more or less columns than the original `X`. Seems like we need to improve this interface a bit to cover all cases.

src/tabpfn/preprocessing/steps/nan_handling_polynomial_features_step.py
bejaeger: Really nice spot! added a test for it and fixed it!

changelog/767.changed.md
bejaeger: Okay, wasn't sure if we want to mention all of the internals. Made it a bit more concise than what you suggested, thanks!

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Yes, this becomes heavy for large datasets, indeed. It should match what sklearn's `ColumnTransformer` is doing internally, though (with e.g. `remainder=passthrough`), and I wanted to be safe for now and prioritize immutability. In the current form this might do more mem-copying than the preivous implementation but I'm not actually sure. Do you want me to test this?

In any case, if we move over the full pipeline to transforming only specific slices per step; we actually have a handle on this and can likely replace some it with in-place operations and improve on what ColumnTransformer is doing. Alternatively, we might split modalities early on and do one concatenation at the end.
Since this is a bit risky I didn't want to do this optimization here.

tests/test_preprocessing/test_pipeline_consistency.py
bejaeger: For convenience, I'd leave this in for now. In case we have to double-check some previous behavior this is handy. I put it on a "clean-up" list of things I anyway wanted to do after many of the preprocessing changes.

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Hm...yeah I did it in a way so it didn't show the diff. The diff wouldn't have been useful anyways though.

I renamed it to be harmonized with `src/tabpfn/preprocessing/torch/pipeline_interface.py`.

bejaeger:
Thanks a lot for the reviews! I should have addressed everything, PTAL.

src/tabpfn/preprocessing/pipeline_interface.py
LeoGrin: sorry wasn't clear! I meant that we add the same modality / number of columns in each transform call (mainly if we call transform on train and on test). Does that make sense?

LeoGrin (approved):
Thanks!!

src/tabpfn/preprocessing/pipeline_interface.py
noahho: Could cloning once at the start of _process_steps (if is_fitting or simply always) be sufficient to preserve immutability for the caller?

We can do current_X = X.copy() once, and then apply transforms in-place on current_X.

That seems like a good immediate optimization that avoids the N-copies issue without risky refactoring.

If this doesn't work, we could also add a TODO note here and do this later.

noahho (approved):
Great stuff!

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Yep, changed it to doing that!

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: Right, added a validation for it!

PR #768, opened 2026-02-01
Refactor SVD features preprocessing step

bejaeger:
Refactors SVD feature processing so that it is a separate step in the preprocessing pipeline.

---

## Public API Changes

-   [ ] No Public API changes
The "scaler" option for the preprocessing config `global_transformer_name` has been removed.


bejaeger:

- #756
    - #767
        - #768 ◀

src/tabpfn/preprocessing/steps/add_svd_features_step.py
```diff
@@ -0,0 +1,123 @@
...
+from tabpfn.preprocessing.pipeline_interface import PreprocessingStep
+from tabpfn.preprocessing.steps.utils import make_standard_scaler_safe
+from tabpfn.utils import infer_random_state
+
+if TYPE_CHECKING:
+    import numpy as np
+
+
+class AddSVDFeaturesStep(PreprocessingStep):
+    """Adds SVD features to the data."""
```
alanprior: @bejaeger can we elaborate slightly more the docstrings? What we do here is add, on top of the raw X, also more features that are just a compressed version of them? If so - I get it for numerical features, but a bit weird for categorical, and non-applicable for text?

src/tabpfn/preprocessing/steps/add_svd_features_step.py
```diff
@@ -0,0 +1,123 @@
...
+    else:
+        raise ValueError(f"Invalid global transformer name: {global_transformer_name}.")
+
+    n_components = max(1, min(n_samples // 10 + 1, n_features // divisor))
+    return Pipeline(
+        steps=[
+            (
+                "save_standard",
+                make_standard_scaler_safe(
+                    ("standard", StandardScaler(with_mean=False)),
```
alanprior: @bejaeger at this step we expect many columns to already be normalized, right? and this is general is to learn a more balanced projection, but unrelated to the original features?

src/tabpfn/preprocessing/steps/add_svd_features_step.py
```diff
@@ -0,0 +1,123 @@
...
+        """Return the number of added features."""
+        if n_features < 2:
+            return 0
+
+        transformer = get_svd_features_transformer(
+            self.global_transformer_name,
+            n_samples,
+            n_features,
+        )
+        return next(
```
alanprior: @bejaeger I find this syntax hard to read. Why do we have next here? I'm confused

src/tabpfn/preprocessing/configs.py
```diff
@@ -99,7 +99,6 @@ class PreprocessorConfig:
     max_features_per_estimator: int = 500
     global_transformer_name: (
         Literal[
-            "scaler",
```
alanprior, line 102: What did scaler mean here?

bejaeger: this was just a standardscaler which didn't really make sense here

src/tabpfn/preprocessing/steps/add_svd_features_step.py
bejaeger: `next` just takes the first element of an iterator. quite a neat structure in python. For this one, I made things more explicit now without using `next`

src/tabpfn/preprocessing/steps/add_svd_features_step.py
bejaeger: This imputes NaNs for the SVD features to work.

src/tabpfn/preprocessing/steps/add_svd_features_step.py
bejaeger: True! it would only make sense for categoricals encoded as one-hot or target encoding or so.

bejaeger: I will follow-up with a PR that changes the order of categorical encoding and svd features, also making the consistency tests cover more cases.

src/tabpfn/preprocessing/steps/add_svd_features_step.py
alanprior: Yeah I'm familiar with the syntax, it was a bit surprising when reading the code but since you already changed it then cool :)

src/tabpfn/preprocessing/steps/add_svd_features_step.py
alanprior: Hmmm OK! It wasn't clear from reading, also if you aren't really re-normalizing then it would make more sense both in terms fo readability and efficiency to impute missing values? Anyway not super crucial, but might be worth a comment

bejaeger: Not sure what you mean, we do this imputation only to get the svd features. The default dataset will still have nans.

alanprior: Very bad phrasing above :) 

What I meant was that if you are already operating on a a dataset which is already scaled, but has missing values, I would expect that rather doing this:

```make_standard_scaler_safe(("standard", StandardScaler(with_mean=False))```

You would have replaced `StandardScaler` with some `ImputeMissing` of sort. Because no point of scaling again, if that's what happening here. But perhaps that would be bad as if the data won't be scaled for some reason, this would break, so maybe my original comment is bad.

Anyway - if it doesn't make sense, let's just go ahead with that.

PR #771, opened 2026-02-02
Option to fix preprocessing seed in finetuning

bejaeger:
Fixing the seed will, e.g., keep column permutations the same across batches. This is expected to improve results in finetuning.

## Todo 
- [x] Update consistency tests


bejaeger:

- #756
    - #767
        - #771 ◀

src/tabpfn/classifier.py
```diff
@@ -723,15 +725,18 @@ def fit(self, X: XType, y: YType) -> Self:
                 "batched",
             ] = "fit_preprocessors"
 
-        byte_size, rng = self._initialize_model_variables()
+        static_seed, rng = infer_random_state(self.random_state)
+        byte_size = self._initialize_model_variables()
         ensemble_configs, X, y = self._initialize_dataset_preprocessing(X, y, rng)
```
psinger-prior: Should this maybe use `static_seed` instead of `rng` also?

bejaeger: Yes I think that makes sense, changed it.

src/tabpfn/finetuning/data_util.py
```diff
@@ -770,10 +773,7 @@ def get_preprocessed_dataset_chunks(
         y_raw = [y_raw]
     assert len(X_raw) == len(y_raw), "X and y lists must have the same length."
 
-    if not hasattr(calling_instance, "models_") or calling_instance.models_ is None:
-        _, rng = calling_instance._initialize_model_variables()
-    else:
-        _, rng = infer_random_state(calling_instance.random_state)
+    calling_instance._initialize_model_variables()
```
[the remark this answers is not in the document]
bejaeger: Fixed!

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -181,6 +187,22 @@ def __init__(  # noqa: PLR0913
...
+            == self.n_estimators_validation
+            == self.n_estimators_final_inference
+        ):
+            warnings.warn(
+                "`use_fixed_preprocessing_seed` is only supported if "
+                "`n_estimators_finetune` == `n_estimators_validation` == "
+                "`n_estimators_final_inference`. We are setting the number of "
+                "estimators for validation and final inference to the same value "
+                f"as `n_estimators_finetune`(={self.n_estimators_finetune}).",
+                UserWarning,
```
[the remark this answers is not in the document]
bejaeger: Done!

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -181,6 +187,24 @@ def __init__(  # noqa: PLR0913
...
         self.meta_batch_size = META_BATCH_SIZE
+        self.use_fixed_preprocessing_seed = use_fixed_preprocessing_seed
+
+        if self.use_fixed_preprocessing_seed and not (
+            self.n_estimators_finetune
+            == self.n_estimators_validation
+            == self.n_estimators_final_inference
+        ):
+            warnings.warn(
+                "`use_fixed_preprocessing_seed` is only supported if "
```
psinger-prior: Could you please elaborate why this is only supported in this case? We could still train on less than we validate for example. Maybe a warning is sufficient instead of overwriting?
Actually the current defaults would even run into this case I believe.

src/tabpfn/preprocessing/initialization.py
```diff
@@ -0,0 +1,75 @@
...
+        A tuple containing the data, the ordinal encoder, and the inferred feature
+        modalities.
+    """
+    feature_schema = detect_feature_modalities(
+        X=X,
+        feature_names=feature_names,
+        min_samples_for_inference=min_samples_for_inference,
+        max_unique_for_category=max_unique_for_category,
+        min_unique_for_numerical=min_unique_for_numerical,
+        reported_categorical_indices=provided_categorical_indices,
```
psinger-prior: I think this one should be `provided_categorical_indices`

src/tabpfn/finetuning/finetuned_base.py
bejaeger: Hmm...the idea was to keep things consistent for validation because of early stopping.
I don't have a strong opinion on overwriting this though. What I now did is to set `n_estimators_final_inference = 2` as well and ONLY print out a warning here. We can change the default if wer find other settings are better.

src/tabpfn/preprocessing/initialization.py
bejaeger: Thanks for spotting this. This entire file is deprecated and not used anywhere. Removed it.

PR #773, opened 2026-02-03
Limit MPS memory to not crash entire Computer on OOM

klemens-floege:
Following: https://docs.pytorch.org/docs/stable/generated/torch.mps.set_per_process_memory_fraction.html

This Script now fails gracefully without crashing MacBook with message: 

```

from sklearn.datasets import make_classification

# Config
N_TRAIN, N_TEST, N_FEATURES, N_CLASSES = 1000, 25_000, 20, 5
DEVICE = "mps"  # Change to "cuda" for NVIDIA GPUs

# Generate data
X_train, y_train = make_classification(
    n_samples=N_TRAIN, n_features=N_FEATURES, n_informative=10,
    n_classes=N_CLASSES, random_state=42
)
X_test, _ = make_classification(
    n_samples=N_TEST, n_features=N_FEATURES, n_informative=10,
    n_classes=N_CLASSES, random_state=123
)


# Fit and predict
clf = TabPFNClassifier(device=DEVICE)
clf.fit(X_train, y_train)

predictions = clf.predict_proba(X_test)

```

Output:

`RuntimeError: MPS backend out of memory (MPS allocated: 2.28 GiB, other allocations: 8.55 GiB, max allowed: 11.20 GiB). Tried to allocate 608.83 MiB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).`

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -6,6 +6,7 @@ import pytest
 import torch
 
 from tabpfn.settings import TabPFNSettings, TestingSettings
+from tests.utils import get_pytest_devices
 
 
 def test__load_settings__env_file_contains_variables_for_other_apps__does_not_crash(
@@ -32,7 +33,7 @@ def test__ci_env_false_sets_false(monkeypatch: pytest.MonkeyPatch) -> None:
     assert testing_settings.ci is False
 
 
-@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="MPS not available")
+@pytest.mark.skipif("mps" not in get_pytest_devices(), reason="MPS not available")
 def test__mps_memory_limit__low_limit__causes_oom() -> None:
     """Test that a very low MPS memory limit causes OOM on large allocations."""
     TabPFNSettings(mps_memory_fraction=0.001)  # 0.1% - triggers model_post_init
@@ -40,15 +41,3 @@ def test__mps_memory_limit__low_limit__causes_oom() -> None:
 
     with pytest.raises(RuntimeError, match="out of memory"):
         _ = [torch.randn(5000, 5000, device="mps") for _ in range(10)]
-
-
-@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="MPS not available")
-def test__mps_memory_limit__reasonable_limit__no_oom() -> None:
-    """Test that a reasonable MPS memory limit allows normal allocations."""
-    TabPFNSettings(mps_memory_fraction=0.7)  # 70% - triggers model_post_init
-    torch.mps.empty_cache()
-
-    x = torch.randn(1000, 1000, device="mps")
-    y = x @ x.T
-    assert y.shape == (1000, 1000)
-    del x, y


src/tabpfn/architectures/base/memory.py
```diff
@@ -175,3 +178,34 @@ def _get_num_cells(
...
+    References:
+        https://pytorch.org/docs/stable/generated/torch.mps.set_per_process_memory_fraction.html
+    """
+    if not torch.backends.mps.is_available():
+        return
+
+    if memory_fraction is None:
+        memory_fraction = float(
+            os.environ.get("TABPFN_MPS_MEMORY_FRACTION", MPS_DEFAULT_MEMORY_FRACTION)
+        )
```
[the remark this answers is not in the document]
klemens-floege: done

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [6.3.2] - 2026-02-03
+
+### Added
+
+- Add MPS memory limiting to prevent macOS system crashes when using Apple Silicon GPUs. Memory is automatically limited to 70% of recommended max on import. Configurable via `TABPFN_MPS_MEMORY_FRACTION` environment variable.
```
bejaeger: This shouldn't be necessary anymore. The files in `changelog/` will be automatically compiled and added here.

src/tabpfn/architectures/base/memory.py
```diff
@@ -175,3 +178,35 @@ def _get_num_cells(
     n_train, n_features = X_train_shape
     n_test, _ = X_test_shape
     return (n_train + n_test) * n_features
+
+
+def configure_mps_memory_limits(
```
bejaeger: Should we add this somewhere more central? Maybe `src/tabpfn/settings.py`? It's not really architecture related.

src/tabpfn/__init__.py
```diff
@@ -1,5 +1,14 @@
 from importlib.metadata import version
 
+# Configure MPS memory limits early to prevent macOS system crashes
```
bejaeger: Do we need to run this before the imports below?

src/tabpfn/__init__.py
```diff
@@ -1,5 +1,14 @@
...
+# This must happen before any MPS tensor operations
+import torch
+
+if torch.backends.mps.is_available():
+    from tabpfn.architectures.base.memory import configure_mps_memory_limits
+
+    configure_mps_memory_limits()
```
bejaeger: I'm wondering if it's cleaner to add this logic to another file, e.g. `settings.py` as mentioned above?
Then we can just do:

```
from tabpfn.settings import configure_mps_memory_limits
configure_mps_memory_limits()
```

here. The `torch.backend.mps` check is also handled by the function itself.

src/tabpfn/__init__.py
```diff
@@ -1,5 +1,14 @@
 from importlib.metadata import version
 
+# Configure MPS memory limits early to prevent macOS system crashes
+# This must happen before any MPS tensor operations
```
oscarkey: Did you test this? The pytorch docs don't mention it, but maybe it's hidden behaviour.

src/tabpfn/architectures/base/memory.py
oscarkey: we can also use the pydantic settings system in `settings.py` rather than parsing the environment variable manually.

CHANGELOG.md
klemens-floege: Done

src/tabpfn/__init__.py
klemens-floege: Now moved everything into the `model_post_init` of `TabPFNSettings`, seemed neater to stick with the pedantic classes in settings, but also happy to define a new function inside settings

src/tabpfn/__init__.py
klemens-floege: > Did you test this? The pytorch docs don't mention it, but maybe it's hidden behaviour.
What do you mean exactly? I can add a test for the memory setting; I guess we do need to set the limits before calling the fit/predict methods

klemens-floege: I now added these tests

```
@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="MPS not available")
def test__mps_memory_limit__low_limit__causes_oom() -> None:
    """Test that a very low MPS memory limit causes OOM on large allocations."""
    TabPFNSettings(mps_memory_fraction=0.001)  # 0.1% - triggers model_post_init
    torch.mps.empty_cache()

    with pytest.raises(RuntimeError, match="out of memory"):
        tensors = [torch.randn(5000, 5000, device="mps") for _ in range(10)]
        del tensors


@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="MPS not available")
def test__mps_memory_limit__reasonable_limit__no_oom() -> None:
    """Test that a reasonable MPS memory limit allows normal allocations."""
    TabPFNSettings(mps_memory_fraction=0.7)  # 70% - triggers model_post_init
    torch.mps.empty_cache()

    x = torch.randn(1000, 1000, device="mps")
    y = x @ x.T
    assert y.shape == (1000, 1000)
    del x, y
```

klemens-floege:
@oscarkey I believe I have addressed all your comments; lmk what you think :)

src/tabpfn/__init__.py
oscarkey: ahhh sorry I misunderstood "This must happen before any MPS tensor operations" as "The limit won't apply if it is set after MPS is initialised" (which is how some of the cuda settings work), rather than "We want to set this limit on initialisation so it applies to all MPS operations in the code base".

PR #780, opened 2026-02-10
Fix infinite loop in fingerprint hashing with inf/large floats

LeoGrin:
## Summary
- **Root cause fix**: `_float_hash_arr` now appends the collision counter as separate bytes instead of adding it to float values. This prevents infinite loops when rows contain `inf` or very large floats (`>= ~1e16`) where `row + k == row` due to float64 precision limits.
- **Safety net**: Added `_MAX_COLLISION_RETRIES = 1000` — the collision loop now raises `RuntimeError` instead of hanging if collisions can't be resolved.

## Test plan
- [x] All 7 existing fingerprint tests pass

---

> [!NOTE]
> **Medium Risk**
> Touches core preprocessing that affects model inputs, so predictions and golden fixtures change; logic is contained but could impact downstream reproducibility and collision behavior.
> 
> **Overview**
> Fixes `AddFingerprintFeaturesStep` collision handling to avoid hangs on rows containing `inf` or very large floats by making the rehash offset affect the byte stream (rather than `row + k`), adding rounding for stable hashing, and enforcing a bounded retry limit that raises on unresolved collisions.
> 
> Fingerprint generation is now fully deterministic: the `random_state` parameter is removed, a dataset-derived salt is used instead, pipeline wiring is updated, and reference prediction/preprocessing fixtures are regenerated (with slightly relaxed cross-device equality tolerance).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a6426ee18642168fb8879ebc94449a72db57aa35. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/changelog/780.changed.md
+++ b/changelog/780.changed.md
@@ -1,1 +1,1 @@
-Remove `random_state` parameter from `AddFingerprintFeaturesStep`; fingerprint hashing is now fully deterministic and no longer uses a random salt.
+Remove `random_state` parameter from `AddFingerprintFeaturesStep`; fingerprint hashing is now fully deterministic and no longer uses a random salt. Predictions will differ slightly from previous versions due to the changed fingerprint values.
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -87,8 +87,7 @@ def test__to__between_fits__outputs_equal(
     np.testing.assert_array_almost_equal(
         prediction_1,
         prediction_2,
-        # Relaxed comparison as cross-device (e.g. CPU vs CUDA) floating-point
-        # can differ at the ~1e-5 level due to different reduction orders.
+        # Use a slightly relaxed comparison as comparing between devices.
         decimal=4,
     )
 


src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -94,13 +98,20 @@ def _transform(  # type: ignore
                 # Start checking from the last known count for this row content
                 add_to_hash = hash_counter[h_base]
 
-                h = _float_hash_arr(row + add_to_hash)
+                h = _float_hash_arr(row, add_to_hash)
```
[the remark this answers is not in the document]
LeoGrin: I think it's fine?

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -16,10 +16,14 @@
...
 
 _CONSTANT = 10**12
+_MAX_COLLISION_RETRIES = 1000
 
 
-def _float_hash_arr(arr: np.ndarray) -> float:
-    _hash = int(hashlib.sha256(arr.tobytes()).hexdigest(), 16)
+def _float_hash_arr(arr: np.ndarray, counter: int = 0) -> float:
+    data = arr.tobytes()
+    if counter != 0:
```
priorphil: Maybe add a small note explaining what happens and why (since + on a numpy array typically means addition)? I.e., that `+` here is on bytes so the offset is *appended* as additional bytes, to avoid precision issues when adding offsets to floats directly.
Also, how about renaming counter --> offset?

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -16,10 +16,14 @@
...
+        data += counter.to_bytes(8, "little", signed=True)
```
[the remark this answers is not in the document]
priorphil: no strong feelings here, but it's a little bit cleaner to convert to signed=False? (And might prevent AI reviewers from complaining in the future?)

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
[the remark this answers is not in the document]
priorphil: Do you know why we have the salted_x in the first place?
Is it just to prevent long loops in case we have the same sample over and over? If it's just that and given that we have the lookup already, how about removing the salt altogether?
This has the very nice side effect of removing all random components, which makes this much more portable across OSs (which is now more relevant given that it's active in training).

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -94,13 +98,20 @@ def _transform(  # type: ignore
...
+                h = _float_hash_arr(row, add_to_hash)
 
                 # Resolve remaining collisions (if row+k accidentally collides with
                 # another row)
+                retries = 0
                 while h in seen_hashes and not np.isnan(row).all():
                     add_to_hash += 1
-                    h = _float_hash_arr(row + add_to_hash)
+                    retries += 1
+                    if retries > _MAX_COLLISION_RETRIES:
```
priorphil, line 115: With the above fix, we shouldn't have that many collisions anymore - should we make _MAX_COLLISION_RETIRES smaller, maybe ~100 (or even better, figuring out what the expected collision probability of our hashing function is :D)

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -16,10 +16,14 @@
...
 
 
-def _float_hash_arr(arr: np.ndarray) -> float:
-    _hash = int(hashlib.sha256(arr.tobytes()).hexdigest(), 16)
+def _float_hash_arr(arr: np.ndarray, counter: int = 0) -> float:
+    data = arr.tobytes()
+    if counter != 0:
+        data += counter.to_bytes(8, "little", signed=True)
+    _hash = int(hashlib.sha256(data).hexdigest(), 16)
     return _hash % _CONSTANT / _CONSTANT
```
priorphil: Nit: given that we execute this for each row, how about making the constant a power of 2, so modulo becomes and with a bitmask? Not sure if python is smart enough to figure this out on its own like a  C++ compiler, so maybe we can do that manually, i.e., 
`_CONSTANT = 2**64 - 1  # Use this to as efficiently compute modulo 2** 64`
and then `(_hash & _CONSTANT) / _CONSTANT`  # modulo _CONSTANT + 1.

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
```diff
@@ -16,10 +16,14 @@
 from tabpfn.utils import infer_random_state
 
 _CONSTANT = 10**12
```
priorphil:
```suggestion
_CONSTANT = 2**64 - 1  # Use this to efficiently compute modulo 2** 64```

File: src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py (hunk 5 above)
Comment (priorphil) [inline src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py]:
```suggestion
    return (_hash & _CONSTANT) / _CONSTANT
```

src/tabpfn/preprocessing/steps/add_fingerprint_features_step.py
LeoGrin: put it to 100!

LeoGrin:
> Do you know why we have the salted_x in the first place?
Is it just to prevent long loops in case we have the same sample over and over? If it's just that and given that we have the lookup already, how about removing the salt altogether?
This has the very nice side effect of removing all random components, which makes this much more portable across OSs (which is now more relevant given that it's active in training).

Not sure, maybe to give more control to the user by being able to change the fingerprint features by changing random_state? I would also agree that it's fine to remove so I removed the salt and the random state dependency. Maybe @LennartPurucker you have an opinion?

LeoGrin:
Thanks for the review :)

LennartPurucker:
To my understanding:

Without salting, the fingerprint features could be identical across training datasets, leading the model to overfit to this.
Now the likelihood of this is very small for many-feature datasets. However, for few-feature datasets, rows can appear identical across datasets, even when their distributions differ.

LennartPurucker:
We could use the number of rows or features, or cells as salt, or something similar, instead of a random number to reduce the likelihood further even for small data, while having it deterministic

PR #784, opened 2026-02-16
Fixed Thinking tokens in kv-cache get added only during fit

MagnusBuehler:
## Issue
Added a small fix, which is part of the solution to the issue:  https://github.com/PriorLabs/TabPFN/issues/631 . 
The issue describes that different `fit_modes` result in different prediction results. 

This PR fixes the bug, that with `fit_mode="fit_with_cache"` (= kv-cache) the thinking tokens were added twice during the `fit` and `predict` call.  

Notes: 
- This is not the full solution to the problem that different fit modes result in different predictions. On my local tiny experiment I still get: 
    Mode: fit_with_cache ROC AUC: 0.9955593931170593
    Mode: fit_preprocessors ROC AUC: 0.9955593931170593
    Mode: low_memory ROC AUC: 0.9965461946466017

-> low memory is still different. 
- Further investigation that caught my attention and might be worth to look at are:
    - different fit modes initialize the Inference engines (and consequently the outer-model-preprocessing) at different times. The random seed (or random generators) thus are in a different state. -> preprocessings get different initial random state
    - the `torch.nn.functional.scaled_dot_product_attention` inside the `full_attention.py` behaves slightly differently (probably rounding precision error) with different context lengths. The context lengths are different for different fit_modes 

- Additional: did not re-enable the tests `tests/test_classifier_interface.py::test_fit_modes_all_return_equal_results` and `tests/test_regressor_interface.py::test_fit_modes_all_return_equal_results` as not yet fully solved. 

## Motivation and Context

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
- tiny local test
---

## Checklist

---

### DIFF
--- a/changelog/784.fixed.md
+++ b/changelog/784.fixed.md
@@ -1,0 +1,1 @@
+Fixed the bug, that with fit_mode="fit_with_cache" (= kv-cache) the thinking tokens were added twice during the fit and predict call.


src/tabpfn/architectures/base/transformer.py
```diff
@@ -523,7 +523,11 @@ def forward(  # noqa: PLR0912, C901
             )
         del embedded_y, embedded_x
 
-        if self.add_thinking_tokens is not None:
+        is_kv_cache_prediction = (
+            self.cache_trainset_representation and single_eval_pos == 0
+        )
+
+        if self.add_thinking_tokens is not None and not is_kv_cache_prediction:
```
oscarkey, line 530: Would you add a quick comment that the thinking tokens are added when the cache is populated, so we don't need to add them again when using the cache?

oscarkey (approved):

oscarkey:
hey @MagnusBuehler , I hope you don't mind but I picked up this fix in https://github.com/PriorLabs/TabPFN/pull/852 to get it in. (i added you as an author)

PR #798, opened 2026-02-24
Add browser-based auth and improved error handling for model loading

brendan-priorlabs:
ref ENG-516


USER_TESTING.md
```diff
@@ -0,0 +1,107 @@
+# User Testing
```
brendan-priorlabs: This should be deleted before merge.

tests/test_browser_auth.py
```diff
@@ -0,0 +1,447 @@
+"""Tests for the browser-based license acceptance module."""
```
brendan-priorlabs, line 1: These are probably not high quality tests to be honest.

brendan-priorlabs:
@ggprior, the diff on this one is kind of misleading. More than half of that is boilerplate tests + script + temporary README, so hopefully this won't be so bad to review.

brendan-priorlabs:
Note: We need to turn off gating in HF + deploy the API + deploy the GUI before merging (definitely before releasing) this.

brendan-priorlabs:
If you want to test this locally, which I'd certainly appreciate, you can remove the cached data models/key with:

```
rm -f ~/Library/Caches/tabpfn/tabpfn-v2.5-* && rm ~/.cache/tabpfn/auth_token
```

Also be sure to set HF_TOKEN to simulate the actual ungating of HF (which is a change we'll make there when merging this PR).

ggprior:
thanks for your comments on testing; taking a look now.

src/tabpfn/browser_auth.py
```diff
@@ -0,0 +1,396 @@
...
+    3. ``~/.tabpfn/token`` (tabpfn-client's cache)
+    """
+    env_token = os.environ.get("TABPFN_TOKEN")
+    if env_token:
+        return env_token.strip()
+
+    for path in (_TOKEN_FILE, _CLIENT_TOKEN_FILE):
+        if path.is_file():
+            token = path.read_text().strip()
+            if token:
```
ggprior: minor: wondering if we should also check that license is well-formed -- but I see you take care here not to add deps (JWT parsing would likely add a dependency or home-brew code we wouldn't want to maintain here -- perhaps a simple check that length is greater than zero

src/tabpfn/browser_auth.py
```diff
@@ -0,0 +1,396 @@
...
+    try:
+        with urllib.request.urlopen(req, timeout=10) as resp:  # noqa: S310
+            return resp.status == 200
+    except urllib.error.HTTPError as exc:
+        if exc.code in (401, 403):
+            return False
+        logger.warning("Unexpected HTTP %s from token verification endpoint", exc.code)
+        return None
+    except Exception:  # noqa: BLE001
+        logger.debug("Token verification endpoint unreachable", exc_info=True)
```
ggprior: nit: maybe this should be raised up to `error`

src/tabpfn/browser_auth.py
```diff
@@ -0,0 +1,396 @@
...
+        License has not been accepted (or token invalid).
+    None
+        Server is unreachable — cannot verify.
+    """
+    url = f"{api_url.rstrip('/')}/tabpfn/license/"
+    req = urllib.request.Request(url, headers={"Authorization": f"Bearer {token}"})
+    try:
+        with urllib.request.urlopen(req, timeout=10) as resp:  # noqa: S310
+            data = json.loads(resp.read())
+            return data.get("accepted", False)
```
ggprior, line 160: let's consider the case where license version is updated -- would we want to trigger a new acceptance flow? In this case, simple check for accept is probably not enough and we would need an additional check if accepted license version matches current license version

src/tabpfn/browser_auth.py
```diff
@@ -0,0 +1,396 @@
...
+
+    Returns ``True`` on success.
+
+    Raises
+    ------
+    TabPFNLicenseError
+        If the license cannot be accepted (no browser, server unreachable
+        without cached token, etc.).
+    """
+    from tabpfn.errors import TabPFNLicenseError  # noqa: PLC0415
```
ggprior: nit: wondering if these imports can move to top

ggprior (commented):
LGTM! Gonna proceed with testing locally now.

tests/quick_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+# from tabpfn_client import UserDataClient
+from tabpfn import TabPFNClassifier, TabPFNRegressor
+from tabpfn.constants import ModelVersion
+
+logging.basicConfig(level=logging.INFO)
+
+
+if __name__ == "__main__":
+    use_server = True
```
ggprior: minor: don't think this is used

ggprior (approved):
Flow here looks good!

src/tabpfn/browser_auth.py
brendan-priorlabs: Done!

src/tabpfn/browser_auth.py
brendan-priorlabs: Good call, done

src/tabpfn/browser_auth.py
brendan-priorlabs: I think this is the current behavior provided that we're okay with users always simply accepting the latest license. In full generality I can imagine having users that accept some licenses, but not others (and the db does have enough info to support this if we want to go there), but for now this simpler approach seems sufficient

src/tabpfn/browser_auth.py
brendan-priorlabs: Done

tests/quick_test.py
brendan-priorlabs: fixed

ggprior:
@brendan-priorlabs this is pretty much feature complete now from my side, supports 2.5/2.6 licenses, redirects working, just waiting for @noahho to give a quick pass and then we're GTG here

src/tabpfn/browser_auth.py
ggprior: note: discussed with Diana last week, the license acceptance flow is expected depending on version. so that's now implemented

ggprior: to clarify, if you instantiate a 2.6 model and you only accepted terms for 2.5, you'll see a new acceptance flow with the 2.6 license

noahho (approved):
LGTM!

Issue #608, opened 2025-11-06
Prints "Warning: PYTORCH_CUDA_ALLOC_CONF is deprecated" every time


oscarkey:
### Describe the bug

I assume due to pytorch upgrade. Happens on 2.8, unsure about earlier.

Full message:
```
[W1106 11:25:56.492120877 AllocatorConfig.cpp:28] Warning: PYTORCH_CUDA_ALLOC_CONF is deprecated, use PYTORCH_ALLOC_CONF instead (function operator())
```

Comes from this line: https://github.com/PriorLabs/TabPFN/blob/2487f5a56aba1d071534ed83084e2667b583c730/src/tabpfn/architectures/base/memory.py#L17

### Steps/Code to Reproduce

Run `python examples/tabpfn_for_multiclass_classification.py`

### Expected Results

No warning

### Actual Results

Warning

### Versions

```shell
Latest
```

klemens-floege:
@oscarkey what is the state of this?

klemens-floege:
Just synced with @oscarkey offline, and this still needs to be fixed.

arthur-priorlabs:
Closing as discussed with @oscarkey, as this is present only on `2.8 <= torch < 2.12`, and `2.12` is the version of PyTorch now installed by default by `uv sync`.

PR #801, opened 2026-02-27
PYTORCH_CUDA_ALLOC_CONF deprecated fix

klemens-floege:
Issue: https://github.com/PriorLabs/TabPFN/issues/608

Replace deprecated PYTORCH_CUDA_ALLOC_CONF environment variable with PYTORCH_ALLOC_CONF to fix warning on PyTorch >= 2.8.


klemens-floege:
Synced with @oscarkey and the fix is not quite as simple + require testing to make sure everything works with different torch version. We will move it to the backlog for now due to capacity issue.

PR #802, opened 2026-03-01
Re-enable tests that low_memory and fit_preprocessors are equal.

egeonur:
## Issue

PriorLabs/TabPFN#631 fix for this issue<br>I tried to fix the precision issue.First fix was come from [#784](<https://github.com/PriorLabs/TabPFN/issues/784>) which doesn't add thinking tokens so that single eval pos stays zero and kv cache can be used during prediction. I casted all tensors for the given dtype so results became

no_cache vs repeat: 0.0<br>no_cache vs fit_preprocessors: 0.0<br>no_cache vs fit_with_cache: 0.0

only caveat is that when I run script with float32 there were still some inconsistencies like:

no_cache vs repeat: 5.3390077e-06<br>no_cache vs fit_preprocessors: 5.3390077e-06<br>no_cache vs fit_with_cache: 5.5486857e-06

but I am guessing it might be related to low precision 64 vs 32.

Also tests/test_consistency.py fails locally but I assume they are stored for future comparisons if sth deviates

## Motivation and Context

code to run on local machine for above results:

```python
import random
from sklearn.datasets import load_diabetes
import numpy as np
import torch

X, y = load_diabetes(return_X_y=True)
    X, y, test_size=0.33, random_state=42
)

def _set_seeds() -> None:
    torch.manual_seed(0)
    np.random.seed(0)
    random.seed(0)

_set_seeds()
reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache = reg.predict(X_test)

reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache_repeat = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_preprocessors", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_cache_preproc = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_with_cache", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_kv_cache = reg.predict(X_test)

def _max_diff(a: np.ndarray, b: np.ndarray) -> float:
    return np.max(np.abs(a - b) / np.abs(a))

print("max relative diffs")
print("no_cache vs no_cache_repeat:", _max_diff(preds_no_cache, preds_no_cache_repeat))
print("no_cache vs cache_preproc:", _max_diff(preds_no_cache, preds_cache_preproc))
print("no_cache vs kv_cache:", _max_diff(preds_no_cache, preds_kv_cache))
```

## Public API Changes

- \[ X\] No Public API changes

---

## How Has This Been Tested?

Tested locally without GPU only on macbook cpu. <br>Collecting system and dependency information...<br>PyTorch version: 2.10.0<br>CUDA used to build PyTorch: None<br>ROCM used to build PyTorch: N/A

OS: macOS 15.7.3 (arm64)<br>GCC version: Could not collect<br>Clang version: 17.0.0 (clang-1700.0.13.5)<br>CMake version: version 3.31.1<br>Libc version: N/A

Python version: 3.11.9 (main, Nov 22 2024, 14:33:40) \[Clang 14.0.3 (clang-1403.0.22.14.1)\] (64-bit runtime)<br>Python platform: macOS-15.7.3-arm64-arm-64bit<br>Is CUDA available: False<br>CUDA runtime version: No CUDA<br>CUDA_MODULE_LOADING set to: N/A<br>GPU models and configuration: No CUDA<br>Nvidia driver version: No CUDA<br>cuDNN version: No CUDA<br>HIP runtime version: N/A<br>MIOpen runtime version: N/A<br>Is XNNPACK available: True

CPU:<br>Apple M1 Max

## Dependency Versions:

tabpfn: 6.4.1<br>torch: 2.10.0<br>numpy: 2.4.2<br>scipy: 1.17.1<br>pandas: 2.3.3<br>scikit-learn: 1.8.0<br>typing_extensions: 4.15.0<br>einops: 0.8.2<br>huggingface-hub: 1.5.0

## Checklist

- \[ X\] The changes have been tested locally.
- [ ] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.
- \[X \] The code follows the project's style guidelines.
- [ ] I have considered the impact of these changes on the public API.

---


egeonur:
@klemens-floege hey I checked the failing tests but it is not sth I changed in my pr. #757  after this change this test can be failed but my local changes don't touch modality_detection.py or test_modality_detection.py. what would you suggest? macos test I can fix. it is consistency test so no big issue I already mentioned it in pr description but other platform errors shouldn't depend on my changes. The test seems flaky or at least environment-sensitive.

klemens-floege:
@egeonur could you pls try a simple rebase to main? For the changelog test you need to add an .md file in the changelog folder. Thank you :)

egeonur:
@klemens-floege I rebased, fixed the consistency test and added .md file. Can you run again to see whether the tests are fixed? 🤞

egeonur:
@klemens-floege I get the same errors again but #815 had the same failing tests. my local python version is 3.11 and I guess this error happens since CI paths use Python 3.14 so I guess behaviour of the of pd.to_numeric or s.isna() changed in Python 3.14

klemens-floege (commented):
Hi, thanks again for contributing! I ran the reproduction script locally from your branch (on CPU) and the inconsistencies are still present:

- no_cache vs cache_preproc:  0.013
- no_cache vs kv_cache:       1.423
 

A few thoughts:

On the consistency tests: I'd prefer not to adjust the random dataset inside the estimators to fix the test failures — those tests are there to verify that the internal model behaviour hasn't changed, so relaxing them isn't a great signal. If adjusting the random state inside the estimators is truly unavoidable to fix the inconsistency, that may be acceptable, but it should be a deliberate decision.

What I think would be valuable: Could you add a test to test_regressor.py and test_classifier.py that explicitly verifies the different fit_mode options produce consistent predictions? Something like:

```
def test_fit_mode_consistency(regressor_or_classifier):
      # assert predictions from low_memory, fit_preprocessors, fit_with_cache
      # are all within float tolerance of each other
```

This would both document the expected behaviour and catch regressions going forward.

egeonur:
```
import random
from sklearn.datasets import load_diabetes
import numpy as np
import torch

X, y = load_diabetes(return_X_y=True)
    X, y, test_size=0.33, random_state=42
)

def _set_seeds() -> None:
    torch.manual_seed(0)
    np.random.seed(0)
    random.seed(0)

_set_seeds()
reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64, device="cpu")
reg.fit(X_train, y_train)
preds_no_cache = reg.predict(X_test)

reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64, device="cpu")
reg.fit(X_train, y_train)
preds_no_cache_repeat = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_preprocessors", inference_precision=torch.float64, device="cpu")
reg.fit(X_train, y_train)
preds_cache_preproc = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_with_cache", inference_precision=torch.float64, device="cpu")
reg.fit(X_train, y_train)
preds_kv_cache = reg.predict(X_test)

def _max_diff(a: np.ndarray, b: np.ndarray) -> float:
    return np.max(np.abs(a - b) / np.abs(a))

print("max relative diffs")
print("no_cache vs no_cache_repeat:", _max_diff(preds_no_cache, preds_no_cache_repeat))
print("no_cache vs cache_preproc:", _max_diff(preds_no_cache, preds_cache_preproc))
print("no_cache vs kv_cache:", _max_diff(preds_no_cache, preds_kv_cache))
```
when I run this on with uv run my local CPU I get this result: 
no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc: 0.0
no_cache vs kv_cache: 0.0
@klemens-floege If I may ask which script did you run? egeonur:ege/fix-fit-mode this is the branch name where I fixed the issue. If you run it on the main I get the same errors as you do but that is before my fix. If it is another script can I try it as well?  the override_random_state change isn't to make consistency tests pass. it's to fix the underlying inconsistency between fit modes. without that change
 no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc: 0.013375835
no_cache vs kv_cache: 0.013375835 they were still inconsistent so I changed it to fix not to past consistency test
| Comparison                      | main   | dtype fix only | dtype fix + random state fix |
|---------------------------------|--------|----------------|------------------------------|
| no_cache vs no_cache_repeat     | 0.0%   | 0.0%           | 0.0%                         |
| no_cache vs cache_preproc       | 1.3%   | 1.3%           | 0.0%                         |
| no_cache vs kv_cache            | 142.3% | 1.3%           | 0.0%                         |

klemens-floege:
@egeonur I ran you script with device set to cpu: 

```
max relative diffs
no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc:   0.013377033
no_cache vs kv_cache:        1.4230262
```

```
tabpfn: 6.4.1 
numpy: 2.3.3
pandas: 2.3.3
scikit-learn: 1.6.1
scipy: 1.16.2
torch: 2.10.0
```


It could be that the pandas version makes a difference will do more digging, I think a concrete good next step in this PR is to add a new test to `test_consistency.py` that specifically tests this behavior: 


```
  @pytest.mark.parametrize("estimator_cls,data_fn", [
      (TabPFNRegressor, _get_tiny_regression_data),
      (TabPFNClassifier, _get_tiny_classification_data),
  ])
  def test__fit_modes__produce_consistent_predictions(estimator_cls, data_fn):
      """All fit_mode values should produce numerically equivalent predictions."""
      X_train, y_train, X_test = data_fn()
      fit_modes = ["low_memory", "fit_preprocessors", "fit_with_cache"]
      preds = {}
      for mode in fit_modes:
          model = estimator_cls(**DEFAULT_CONFIG, fit_mode=mode)
          model.fit(X_train, y_train)
          if isinstance(model, TabPFNClassifier):
              preds[mode] = model.predict_proba(X_test)
          else:
              preds[mode] = model.predict(X_test)

      reference = preds["low_memory"]
      for mode in ["fit_preprocessors", "fit_with_cache"]:
          np.testing.assert_allclose(
              preds[mode], reference, rtol=1e-3, atol=1e-3,
              err_msg=f"fit_mode='{mode}' predictions differ from 'low_memory'",
          )
```

oscarkey:
hey @egeonur , I'm just merging https://github.com/PriorLabs/TabPFN/pull/852 which makes `fit_preprocessors` and `fit_with_cache` consistent. However, `low_memory` is still inconsistent, do you think this change might help with that?

Re the tests, we have `test__fit_preprocessors_and_low_memory_produce_equal_results` in both `test_regressor_interface.py` and `test_classifier_interface.py`. These are currently disabled, because the fit modes are not consistent. So if this PR works we should be able to re-enable them!

egeonur:
Hey @oscarkey — I was about to write the same update.

I tested my fork again on a different MacBook with an M3 chip yesterday, and I still got zero diffs. That matches what I saw before.

What confuses me is that @klemens-floege said he still did not see improvements, and reported these max relative diffs:

```text
no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc:   0.013377033
no_cache vs kv_cache:        1.4230262
```

So from his side it looks like the fix is not taking effect.

I am wondering whether this could be a branch/version mismatch somewhere, although I am not certain. I tested with the same library versions he mentioned:

* tabpfn: 6.4.1
* numpy: 2.3.3
* pandas: 2.3.3
* scikit-learn: 1.6.1
* scipy: 1.16.2
* torch: 2.10.0

On my side, with float64 precision, I again got zero diffs.

The key thing I found is this change in PR #802:
https://github.com/PriorLabs/TabPFN/pull/802/changes#diff-261bbedda6d582b2e6a903d83c193c83d082dd4d6b8513305b998ca732f6ce73L366

That part seems to solve the issue by overriding the random state setting to `self.static_seed`.

So I am not sure why Klemens still sees no improvement with my local version, especially because my solution was built on top of your solution. Even if it did not fully fix everything on his side, I would have expected at least some improvement. If you have time to check whether switching to this PR changes the results when you run it, I’d really appreciate it. Then I can pull your latest changes from main and try again with the seed improvement.

oscarkey:
hey @egeonur . I applied your static seed fix and ran `test_regressor_interface.py:test__fit_preprocessors_and_low_memory_produce_equal_results` and `test_classifier_interface.py:test__fit_preprocessors_and_low_memory_produce_equal_results`, and they passed. So I think you should be good to rebase the PR!

I think it should also be safe to update `tests/test_consistency.py` to stop specifying the fit_mode entirely and just test the default one, as we'll now have tests asserting that all the fit modes produce the same answers.

egeonur:
@oscarkey I removed the `skip` markers from the `fit_preprocessors` vs `low_memory` equivalence tests in `tests/test_classifier_interface.py` and `tests/
  test_regressor_interface.py`, and they now pass locally. I also regenerated the consistency snapshots and `tests/test_consistency.py` passes locally.

  On your earlier comment about simplifying `tests/test_consistency.py`: do you mean removing the fit-mode-specific cases in `TEST_CASES` and just testing
  the default estimator behavior there, while relying on the interface tests for fit-mode equivalence? The blocks I think you mean are at:
  - `tests/test_consistency.py:109`
  - `tests/test_consistency.py:124`
  - `tests/test_consistency.py:137`
  - `tests/test_consistency.py:152`

  If that’s what you had in mind, I’m happy to make that cleanup too. I just wanted to confirm exactly what you’d prefer to remove.

  Other than that, I think this issue is fixed. All relevant tests passed locally. I’m still seeing some unrelated mypy errors in `tests/
  test_browser_auth.py`, but they don’t seem related to this PR.

src/tabpfn/inference.py
```diff
@@ -809,14 +810,20 @@ def __init__(  # noqa: PLR0913
...
+                force_inference_dtype
+                if force_inference_dtype is not None
+                else torch.float32
+            )
+
+            X = torch.as_tensor(X, dtype=inference_dtype, device=device)
             X = X.unsqueeze(1)
-            if not isinstance(y, torch.Tensor):
-                y = torch.as_tensor(y, dtype=torch.float32, device=device)
+            y = torch.as_tensor(y, dtype=inference_dtype, device=device)
```
oscarkey: I think this is already happening a few lines below?

src/tabpfn/inference.py
```diff
@@ -874,7 +881,16 @@ def iter_outputs(
...
+            inference_dtype = (
+                self.force_inference_dtype
+                if self.force_inference_dtype is not None
+                else torch.float32
+            )
+            X_test = torch.as_tensor(
+                X_test,
+                dtype=inference_dtype,
+                device=self.device,
+            )
```
oscarkey: Same again? This already happens below?

oscarkey (commented):
> do you mean removing the fit-mode-specific cases in `TEST_CASES`

yes exactly! To update the reference predictions, I would suggest renaming `[prefix]_fit_preprocessors.json` to just `[prefix].json`, and deleting the `low_memory`/`fit_with_cache` reference predictions. You can also revert all other reference prediction changes, as the predictions should not have changed.

src/tabpfn/inference.py
egeonur: The lower cast only overlaps partially. My change makes the KV-cache path normalize inputs up front with inference_dtype = force_inference_dtype or
  torch.float32, even when X and y already arrive as tensors. Before that, this branch only cast non-tensor inputs to float32, so tensor inputs could retain
  an arbitrary dtype, and force_inference_dtype was not applied consistently during cache building. The later block only re-applies the forced dtype after
  GPU preprocessing, so it does not replace the up-front normalization or the default float32 path.

  With the seed held fixed, that difference was observable in output consistency. Before this change, I saw:

  

> max relative diffs
>   no_cache vs no_cache_repeat: 0.0
>   no_cache vs cache_preproc: 0.0
>   no_cache vs kv_cache: 1.191185e-06

  After normalizing the KV-cache path to use self.force_inference_dtype instead of defaulting to float32, the discrepancy went to zero:

> 
>   max relative diffs
>   no_cache vs no_cache_repeat: 0.0
>   no_cache vs cache_preproc: 0.0
>   no_cache vs kv_cache: 0.0  

When I last checked TabPFN-2.6 doesn't support currently fit_with_cache fit_mode so these were the results from  TabPFN-2.5. but if you say they are redundant I can revert them back but without that change small difference will stay imo.

egeonur:
@oscarkey I updated as you sugessted. I simplified tests/test_consistency.py to snapshot only the default estimator behavior, removed the fit-mode-specific consistency cases, renamed
  the surviving *_fit_preprocessors.json references to the plain filenames, and deleted the obsolete low_memory / fit_with_cache reference predictions. I
  also rebased onto current main, preserved the deterministic seed fix in inference.py, and re-ran uv run pytest tests/test_consistency.py tests/
  test_inference.py; both pass. also for the other change I replied to your pr review but can you look into that as wellso that if you find it maybe we can merge this pr? otherwise I can revert that change and see any cleanup is required. Thanks :)

src/tabpfn/inference.py
oscarkey: ahhh I see my mistake! I think there might still be some redundancy (see my other comments), but if we fix that I think we're ready to go :)

src/tabpfn/inference.py
oscarkey, line 971: Same as above, hopefully we can delete this line.

src/tabpfn/inference.py
```diff
@@ -825,7 +832,16 @@ def iter_outputs(
         for ensemble_member, model in zip(self.ensemble_members, self.models):
             model.to(self.device)
             X_test = ensemble_member.transform_X_test(X)
-            X_test = torch.as_tensor(X_test, dtype=torch.float32, device=self.device)
+            inference_dtype = (
+                self.force_inference_dtype
+                if self.force_inference_dtype is not None
+                else torch.float32
+            )
```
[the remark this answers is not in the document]
oscarkey: same again.

src/tabpfn/inference.py
```diff
@@ -768,14 +769,20 @@ def __init__(  # noqa: PLR0913
...
             X = ensemble_member.X_train
             y = ensemble_member.y_train
 
-            if not isinstance(X, torch.Tensor):
-                X = torch.as_tensor(X, dtype=torch.float32, device=device)
+            inference_dtype = (
+                force_inference_dtype
+                if force_inference_dtype is not None
+                else torch.float32
+            )
```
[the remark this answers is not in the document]
oscarkey: would you make this change? This way it will match what we do in `_prepare_model_inputs()`.

src/tabpfn/inference.py
oscarkey, line 896: Hopefully we can delete these lines as we do it above? I guess there's a risk `_maybe_run_gpu_preprocessing()` doesn't preserve the type. But, judging by the other inference engines, it looks like it does?

oscarkey (commented):
hey, thanks for making the changes, I think we're nearly there!

I think the changes in `reference_predictions` can be reverted? The only thing that should change is `*_fit_preprocessors.json` should be renamed to `*.json`, but the contents of the files shouldn't change.

egeonur:
@oscarkey Sorry for the slightly delayed update. I had a rough week and got to this a bit later than planned.

I rebased my branch again and I updated the dtype selection to match `_prepare_model_inputs()` and removed the later redundant `.type(...)` conversions. All tests passed, and I did not notice any behavior changes, so my guess is that `_maybe_run_gpu_preprocessing()` already preserves types, but it would be good to confirm by running the full test suite on GPU as well.

I also cleaned up the reference prediction diff, so the remaining snapshot changes are now just pure `*_fit_preprocessors.json -> *.json` renames with unchanged contents.

I reran:

`uv run pytest tests/test_consistency.py`

and got: `11 passed`
locally all tests passes and if you think everything is fine we can run all tests and maybe merge this pr finally :) thank you very much for your feedbacks

egeonur:
@oscarkey I rebased again. There were additional upstream changes, but it should work now.

All tests are passing. I also rolled back the changes I had made, since upstream already addresses the consistency issue.

In particular, deleting the type casts after `_maybe_run_gpu_preprocessing()` causes problems with the maximum relative differences:

* `no_cache` vs `no_cache_repeat`: `0.0`
* `no_cache` vs `cache_preproc`: `0.0`
* `no_cache` vs `kv_cache`: `0.0029411074`

With TabPFN v2.5, the difference between `no_cache` and `kv_cache` does not go to zero. Given the upstream changes, I think it is better to keep the casts in place.

At this point, this PR confirms that the precision issue appears to be resolved. if you want you can run the all tests again

egeonur:
@oscarkey #884 also made almost the exact same changes I made, so I accepted those changes. Locally, I’m again getting these max relative diffs:

no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc: 0.0
no_cache vs kv_cache: 0.0

However, as I mentioned in my previous comment, removing the type casts after _maybe_run_gpu_preprocessing() causes issues with the maximum relative differences:

no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc: 0.0
no_cache vs kv_cache: 0.0029411074

So I decided to keep them.

I also rebased again after #886, since it had conflicts with the tests, which we had agreed to remove. Other than that, all other tests are passing right now.

If you have time today, could you take a look at this PR? I think it seems ready, and if it looks good to you, I can avoid doing one more rebase and appreciate it a lot :)

Thanks!

egeonur:
@oscarkey hey I think all checks has passed your last review is required. thanks

oscarkey (approved):
lgtm! Thanks so much for your contribution. Sorry this has been quite a long journey.

egeonur:
@oscarkey yes it has been a long journey but thanks for your support as well. Hopefully see you for next contributions :)

PR #805, opened 2026-03-02
More informative Memory Error Msg

klemens-floege:
New error message:

```
(TabPFN) klemensfloege@Klemenss-MacBook-Pro TabPFN % python3 test_oom_message.py
TabPFNMPSOutOfMemoryError:

MPS out of memory with 5,000 test samples.

This is issue is usually caused by one of the following two reasons:

1) Large test set — split into batches:

    predictions = []
    for i in range(0, len(X_test), 100):
        pred = model.predict_proba(X_test[i:i + 100])
        predictions.append(pred)
    predictions = np.vstack(predictions)

2) Large training set — batching won't help.
   You need subsampling or ensembling, see:
   https://github.com/PriorLabs/tabpfn-extensions/blob/main/examples/large_datasets/large_datasets_example.py

Your sizes: 2,000 train / 5,000 test samples, 30 features.
Not sure which? If model.predict_proba(X_test[:1]) also fails, it's (2).

Original error: MPS backend out of memory (MPS allocated: 1.09 GiB, other allocations: 10.07 GiB, max allowed: 11.20 GiB). Tried to allocate 236.21 MiB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

Script on MacOS: 

"""Force an OOM through TabPFN on MPS to see the real error message.

Usage:
    PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.01 python test_oom_message.py
"""

```

import numpy as np

rng = np.random.default_rng(42)

X_train = rng.standard_normal((2000, 30))
y_train = rng.integers(0, 3, size=2000)

model = TabPFNClassifier(device="mps")
model.fit(X_train, y_train)

X_test = rng.standard_normal((5000, 30))

try:
    model.predict_proba(X_test)
except Exception as e:
    print(f"{type(e).__name__}:\n")
    print(e)
```

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -106,7 +106,7 @@ jobs:
       id-token: write
     steps:
       - name: Download build artifacts
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: dist
           path: dist/
@@ -160,7 +160,7 @@ jobs:
       id-token: write
     steps:
       - name: Download build artifacts
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: dist
           path: dist/
@@ -175,7 +175,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Download release notes
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: release_notes
           path: .


src/tabpfn/errors.py
```diff
@@ -69,13 +69,26 @@ def __init__(
...
+            f"resolves the issue:\n\n"
+            f"    batch_size = 1000  # adjust for your hardware\n"
             f"    predictions = []\n"
             f"    for i in range(0, len(X_test), batch_size):\n"
             f"        batch = model.{predict_method}(X_test[i:i + batch_size])\n"
             f"        predictions.append(batch)\n"
-            f"    predictions = np.vstack(predictions)"
+            f"    predictions = np.vstack(predictions)\n\n"
+            f"If even batch_size=1 causes OOMs, your training set is too large — "
+            f"TabPFN's attention mechanism scales quadratically with the number "
```
priorphil: This statement might be confusing - Memory doesn't scale quadratically since we use flash attention, only compute is quadratic.

klemens-floege: ofc thanks for the correction! For I just rephrased to "increases" with training and test set size. in case you still find it confusing lmk :)

src/tabpfn/errors.py
```diff
@@ -69,13 +69,27 @@ def __init__(
...
         message = (
             f"{self.device_name} out of memory{size_info}.\n\n"
-            f"Solution: Split your test data into smaller batches:\n\n"
-            f"    batch_size = 1000  # depends on hardware\n"
+            f"TabPFN loads the entire training set into memory "
+            f"at prediction time alongside each test batch to "
+            f"apply its In-Context Learning capability, so memory "
+            f"scales with both training and test set size.\n\n"
+            f"If your test set is large compared to the training "
+            f"set size, splitting into smaller batches often "
```
priorphil:
```suggestion
            f"set size, splitting test into smaller batches often "
```

src/tabpfn/errors.py
```diff
@@ -69,13 +69,27 @@ def __init__(
 
         message = (
             f"{self.device_name} out of memory{size_info}.\n\n"
-            f"Solution: Split your test data into smaller batches:\n\n"
-            f"    batch_size = 1000  # depends on hardware\n"
+            f"TabPFN loads the entire training set into memory "
```
noahho: I'd suggest this message:

```
message = (
    f"{self.device_name} out of memory{size_info}.\n\n"
    f"This is usually one of two things:\n\n"
    f"1) Large test set — split into batches:\n\n"
    f"    predictions = []\n"
    f"    for i in range(0, len(X_test), 100):\n"
    f"        pred = model.{predict_method}("
    f"X_test[i:i + 100])\n"
    f"        predictions.append(pred)\n"
    f"    predictions = np.vstack(predictions)\n\n"
    f"2) Large training set — batching won't help.\n"
    f"   You need subsampling or ensembling, see:\n"
    f"   https://github.com/PriorLabs/tabpfn-extensions/"
    f"blob/main/examples/large_datasets/"
    f"large_datasets_example.py\n\n"
    f"Your sizes: {n_train_samples:,} train / "
    f"{n_test_samples:,} test samples, "
    f"{n_features} features.\n"
    f"Not sure which? If model.{predict_method}(X_test[:1]) "
    f"also fails, it's (2)."
)
```

klemens-floege: Done!

klemens-floege: The diff just got a bit larger because I had to add these attributed to the model: 

```
n_train_samples=getattr(self, "n_train_samples_", None),
n_features=getattr(self, "n_features_in_", None),
```

Issue #807, opened 2026-03-06
Auto mode using all GPUs


psinger-prior:
### Describe the workflow you want to enable

I believe the `auto` device mode should use all available GPUs, not just the first one. This is common utility across current GPU based ML libraries.

### Describe your proposed solution

The current `auto` mode detects and uses all GPUs available. Can still be manually changed and specified as before.

### Impact

None

psinger-prior:
closed via https://github.com/PriorLabs/TabPFN/pull/808

PR #808, opened 2026-03-06
By default use all gpus available in auto mode

psinger-prior:
## Issue
https://github.com/PriorLabs/TabPFN/issues/807

## Motivation and Context

I believe the `auto` device mode should use all available GPUs, not just the first one. This is common utility across current GPU based ML libraries.

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
Using the test cases and manually running examples on multi gpu machine.

---

## Checklist

---

### DIFF
--- a/src/tabpfn/inference_tuning.py
+++ b/src/tabpfn/inference_tuning.py
@@ -90,23 +90,31 @@ class ClassifierEvalMetrics(str, Enum):
 
 
 METRIC_NAME_TO_OBJECTIVE = {
-    "f1": lambda y_true, y_pred: -f1_score(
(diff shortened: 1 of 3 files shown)


oscarkey:
It was initially set like this while we were testing if multi-gpu inference worked properly, and then we never changed it.

It does make inference on small datasets slower, but the conclusion [on slack](https://prior-labs.slack.com/archives/C08R6A4N54H/p1764834405835659) was that it's better to be slower on small datasets, which already have fast runtime, than slower on big datasets. So lgtm!

Issue #809, opened 2026-03-09
DDP for Finetuning


psinger-prior:
### Describe the workflow you want to enable

Finetuning currently does not support distributed training

### Describe your proposed solution

Implement DDP for distributed finetuning

### Impact

None

psinger-prior:
closed via https://github.com/PriorLabs/TabPFN/pull/812

PR #812, opened 2026-03-09
DDP for FineTuning

psinger-prior:
## Issue
Closing https://github.com/PriorLabs/TabPFN/issues/809

## Motivation and Context

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

Running example scripts on single and multi gpu nodes.

---

## Checklist

---

### DIFF
--- a/examples/finetune_classifier.py
+++ b/examples/finetune_classifier.py
@@ -3,6 +3,8 @@
 Note: We recommend running the fine-tuning script on a CUDA-enabled GPU with 80 GB of VRAM.
 
 Multi-GPU: torchrun --nproc-per-node=N examples/finetune_classifier.py
+Note: Only call fit() once per torchrun session. For multiple finetuning runs, use
+separate torchrun invocations.
 """
 
 import gc
--- a/src/tabpfn/finetuning/finetuned_base.py
+++ b/src/tabpfn/finetuning/finetuned_base.py
@@ -994,7 +994,8 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
         if is_main_process:
             logger.info("--- ✅ Fine-tuning Finished ---")
 
-        self._setup_inference_model(final_inference_eval_config)
+        if is_main_process:
+            self._setup_inference_model(final_inference_eval_config)
 
         self.is_fitted_ = True
         return self
--- a/examples/finetune_regressor.py
+++ b/examples/finetune_regressor.py
@@ -3,6 +3,8 @@
 Note: We recommend running the fine-tuning script on a CUDA-enabled GPU with 80 GB of VRAM.
 
 Multi-GPU: torchrun --nproc-per-node=N examples/finetune_regressor.py
+Note: Only call fit() once per torchrun session. For multiple finetuning runs, use
+separate torchrun invocations.
 """
 
 import gc


alanprior:
@psinger-prior sorry, I've completely missed this PR, probably due to the war. @anuragg1209 would you feel OK to review it?

anuragg1209:
> @psinger-prior sorry, I've completely missed this PR, probably due to the war. @anuragg1209 would you feel OK to review it?

Hi @alanprior, yes, this PR review is under my To-Dos. Please feel free to unsubscribe.

anuragg1209:
Hi @psinger-prior,
Thanks for the PR. I tested the DDP on 4 H100s and everything worked correctly. I wrote test scripts for classifier DDP, regressor DDP, and gradient synchronization verification (parameter difference across ranks: 0.00e+00). All the tests passed, and the core implementation looks solid.

I just have a few comments to add.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -720,47 +935,64 @@ def _fit(  # noqa: C901,PLR0912
...
-            self.finetuned_estimator_ = best_model
+        if self.early_stopping and best_model_state is not None:
+            self.finetuned_estimator_.model_.load_state_dict(best_model_state)
+
+        # --- DDP cleanup ---
+        self._ddp_module_ = None
+        if using_ddp:
+            dist.barrier()
+            if created_process_group:
+                dist.destroy_process_group()
```
anuragg1209, line 992: **Edge case: Multiple fit() calls in one torchrun session can deadlock**

- After `fit()` completes, `dist.destroy_process_group()` is called in `finetuned_base.py`
- If the user calls `fit()` again on a new model, all ranks must re-enter `dist.init_process_group()` at line 79 simultaneously                                                              
  - But rank 0 may be delayed in doing prediction/evaluation (only rank 0 runs `predict()`), while rank 1 already reaches `init_process_group()` and blocks waiting                               
- Result: deadlock until the 30-min NCCL timeout. 

Potential Remedy:
Add a warning in the docs/examples: "Do not call fit() on multiple finetuning instances within a single torchrun session. Use separate scripts or separate torchrun invocations."  

Wdyt?

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -720,47 +935,64 @@ def _fit(  # noqa: C901,PLR0912
...
 
-        logger.info("--- ✅ Fine-tuning Finished ---")
+        if is_main_process:
+            logger.info("--- ✅ Fine-tuning Finished ---")
 
         self._setup_inference_model(final_inference_eval_config)
```
anuragg1209: Minor: `_setup_inference_model()` runs on all ranks. Only rank 0 uses the inference model for prediction. Should we consider guarding with `if is_main_process`: to avoid redundant model cloning + fitting on non-main ranks ?

examples/finetune_classifier.py
```diff
@@ -1,9 +1,13 @@
 """Example of fine-tuning a TabPFN classifier using the FinetunedTabPFNClassifier wrapper.
 
 Note: We recommend running the fine-tuning script on a CUDA-enabled GPU with 80 GB of VRAM.
+
+Multi-GPU: torchrun --nproc-per-node=N examples/finetune_classifier.py
```
anuragg1209, line 5: Minor: I noticed there are no tests for the DDP finetuning code path (existing `test_finetuning_classifier.py` and `test_finetuning_regressor.py` don't cover distributed). What do you think about adding at least a basic smoke test? I understand DDP tests are tricky to run in CI (need multi-GPU), but maybe we could add a test that at least validates the setup/cleanup logic with mocked dist calls, or mark them as `@pytest.mark.skipUnless(torch.cuda.device_count() > 1)`. Happy to add the tests if you think it makes sense. Open to your thoughts!

src/tabpfn/finetuning/finetuned_base.py
psinger-prior: I would say this is an extreme edge case that is not really common, I still added a note to the examples.

examples/finetune_classifier.py
psinger-prior: I think mocked tests are not too useful here, we would need real tests on multi-gpu runners. I would rely on the manual testing for now and fix issues along the way as this is also not core functionality of the package.

anuragg1209: Sounds good!

PR #829, opened 2026-03-20
Add v2.6 model interface

bejaeger:
Merge once we make the model available.


bejaeger:
Close and re-open to get correct github diff

PR #832, opened 2026-03-20
Cache loaded models across fit calls

psinger-prior:
Currently, we reload the weights from local disk each time we fit a model. As the weights are rather small, we can cache the results. This is particularly helpful for repeated hyperparameter tuning runs on smaller data, as well as bagging or agentic workflows.

Summary:
  - Every fit() call currently runs torch.load() to read the model checkpoint from disk, even when the same checkpoint was just loaded. This is redundant disk I/O.
  - This PR adds @functools.cache on the checkpoint loading, so repeated fit() calls with the same model path skip disk I/O entirely. Each call still builds a fresh model instance from the cached checkpoint, so there are no shared mutable state concerns.
 - Cache can be cleared via _load_checkpoint.cache_clear() for long-running processes.

For repeated benchmark on smaller data saw gains of up to 30-40% runtime.

There might be a better solution, open to any suggestions.


src/tabpfn/model_loading.py
```diff
@@ -823,26 +832,24 @@ def load_model(
...
+    checkpoint = _load_checkpoint(str(path))
 
     try:
         architecture_name = checkpoint["architecture_name"]
     except KeyError:
         architecture_name = "base"
     architecture = ARCHITECTURES[architecture_name]
-    state_dict = checkpoint["state_dict"]
+    # Shallow-copy so the pop() below doesn't mutate the cached checkpoint.
+    state_dict = dict(checkpoint["state_dict"])
```
bejaeger: Not the biggest fan of creating a shallow-copy here and then popping the criterion below. 

Could we remove this shallow-copy and instead of popping below be more declarative:

```
criterion_state = {
            k.replace("criterion.", ""): full_state.get(k) 
            for k in criterion_state_keys
        }
```

and then after the `if isinstance(loss_criterion, FullSupportBarDistribution): ... else ....` case have a 

```
model_state = {
        k: v for k, v in full_state.items() 
        if k not in criterion_state_keys
    }
...
model.load_state_dict(model_state)
```

I think that would be a bit more clear.

src/tabpfn/model_loading.py
```diff
@@ -823,26 +832,24 @@ def load_model(
...
+    The raw checkpoint is cached so that repeated calls with the same path
+    skip disk I/O.  Each call returns a fresh model instance so callers can
+    mutate it freely (finetuning, differentiable input, KV caching, etc.).
+
     Args:
         path: Path to the checkpoint
         cache_trainset_representation: If True, the model will cache the
             trainset representation. Forwarded to get_architecture.
     """
-    # Catch the `FutureWarning` that torch raises. This should be dealt with!
```
bejaeger, line 831: should we add back this comment so we don't forget about it and know why we ignore this warning here?

src/tabpfn/model_loading.py
```diff
@@ -819,26 +832,24 @@ def load_model(
...
-    # The warning is raised due to `torch.load`, which advises against ckpt
-    # files that contain non-tensor data.
-    # This `weightes_only=None` is the default value. In the future this will default to
-    # `True`, dissallowing loading of arbitrary objects.
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore", category=FutureWarning)
-        checkpoint: dict = torch.load(path, map_location="cpu", weights_only=None)
+    checkpoint = _load_checkpoint(str(path))
```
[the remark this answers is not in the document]
bejaeger: I think that's a good point.

psinger-prior:
Thanks @bejaeger - should be addressed!

PR #833, opened 2026-03-22
Introduce cardinality limit on One Hot encodings

klemens-floege:
Disabled by default.

### DIFF
--- a/src/tabpfn/preprocessing/configs.py
+++ b/src/tabpfn/preprocessing/configs.py
@@ -37,7 +37,8 @@ class PreprocessorConfig:
         global_transformer_name: Name of the global transformer to use.
         max_onehot_cardinality: Maximum number of unique values a categorical feature
             can have to be one-hot encoded. Features with higher cardinality are passed
-            through unchanged. If None, all categorical features are one-hot encoded.
+            through unchanged to ordinal encoding. If None, all categorical features
+            are one-hot encoded.
     """
 
     name: Literal[


src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -98,6 +100,13 @@ def _get_transformer(
             return ct, categorical_features
 
         if self.categorical_transform_name == "onehot":
+            # Only one-hot encode features with cardinality <= max_onehot_cardinality
+            if self.max_onehot_cardinality is not None:
+                categorical_features = [
+                    i
+                    for i in categorical_features
+                    if len(np.unique(X[:, i])) <= self.max_onehot_cardinality
+                ]
```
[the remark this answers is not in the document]
klemens-floege: done

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -17,6 +17,29 @@
...
+    n_output_features: int,
+    all_cat_indices: list[int],
+) -> list[int]:
+    """Return output indices of all categorical features after one-hot encoding.
+
+    Includes both one-hot encoded columns and high-cardinality categoricals
+    that were passed through in the remainder.
+    """
+    onehot_input_cols = set(ct.transformers_[0][2])
+    onehot_out = list(range(n_output_features))[ct.output_indices_["one_hot_encoder"]]
```
bejaeger: nit: could define 
```
ONE_HOT_ENCODER_NAME="one_hot_encoder"
```
on top of the file to avoid hardcoded strings.

src/tabpfn/preprocessing/pipeline_factory.py
```diff
@@ -93,6 +94,20 @@ def create_preprocessing_pipeline(
...
+        # Second subsampling after expansion to cap blowup from one-hot/SVD
+        max_after = pconfig.max_features_after_expansion
+        if max_after is None:
+            max_after = 2 * pconfig.max_features_per_estimator
+        steps.append(
+            ReshapeFeatureDistributionsStep(
+                transform_name="none",
+                append_to_original=False,
+                max_features_per_estimator=max_after,
+                apply_to_categorical=True,
```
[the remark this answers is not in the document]
bejaeger: I agree that this is an issue.

If we want to add a second resampling, I think we need to first extract the subsampling in its own preprocessing step and don't do it as part of `ReshapeFeatureDistributionStep` (anyway something that's overdue).
However, we can discuss whether we want a second subsampling here. I think it would only be useful if it's some smart subsampling and not only a randomized one. Naive subsampling can easily loose all info of one-hot encoded features.

bejaeger (commented):
Thanks!
I think the "max cardinality one-hot-encoding" is a good addition, but I think we should remove the logic for the second resampling for now.

src/tabpfn/preprocessing/configs.py
```diff
@@ -104,6 +104,7 @@ class PreprocessorConfig:
         ]
         | None
     ) = None
+    max_onehot_cardinality: int | None = None
```
bejaeger, line 111: sorry one nit: could you add to the doc string above?

klemens-floege: done

PR #838, opened 2026-03-25
Fix float overflow in Yeo-Johnson inverse transform

simo-prior:
## Summary

- **Fix `RuntimeWarning: overflow encountered in cast`** in `_yeojohnson_inverse_transform` that occurs when the inverse transform produces float64 values exceeding the input array's dtype range (e.g., float32)
- Use `np.log1p(v)` instead of `np.log(v + 1)` for better numerical stability near zero, consistent with the forward transform which already uses `log1p`/`expm1` pairs

## Problem

The Yeo-Johnson inverse transform *expands* values (unlike the forward transform which compresses via `log`), so for small `lmbda` or large transformed values, the result of `np.expm1(np.log(...) / lmbda)` can exceed the float32 representable range. Since `x_inv` was allocated with the input's dtype, assigning float64 intermediate results back to a float32 array triggered an overflow-on-cast warning, producing `inf` values.

These `inf`s were silently caught downstream by `~np.isfinite(borders_t)` in `transform_borders_one()` and discarded — causing silent loss of regression border resolution rather than a visible error.

## Fix

1. **Compute in float64 internally** — allocate `x_inv` as float64 regardless of input dtype, avoiding overflow during intermediate computation
2. **Use `np.log1p` instead of `np.log(... + 1)`** — more numerically stable for small arguments (both positive and negative branches)
3. **Clip to dtype range before casting back** — values exceeding the target dtype's representable range are clamped to `finfo.max`/`finfo.min` instead of overflowing to `inf`, preserving usable finite borders

### DIFF
--- a/src/tabpfn/preprocessing/steps/safe_power_transformer.py
+++ b/src/tabpfn/preprocessing/steps/safe_power_transformer.py
@@ -141,32 +141,31 @@ def _yeojohnson_inverse_transform(x: np.ndarray, lmbda: float) -> np.ndarray:
     transform with parameter lambda.
     """
     dtype = x.dtype if np.issubdtype(x.dtype, np.floating) else np.float64
-    # Compute in at least float64 to avoid overflow when the inverse produces
-    # values exceeding a lower-precision dtype's range (e.g., float32), while
-    # preserving higher precision (e.g., float128) if the input provides it.
-    compute_dtype = np.promote_types(dtype, np.float64)
-    x_inv = np.zeros_like(x, dtype=compute_dtype)
+    x_inv = np.zeros_like(x, dtype=dtype)
     pos = x >= 0
+
+    # Clip expm1 arguments to prevent overflow in the output dtype.
+    max_arg = np.log(np.finfo(dtype).max)
 
     # when x >= 0
     if abs(lmbda) < np.spacing(1.0):
-        x_inv[pos] = np.expm1(x[pos])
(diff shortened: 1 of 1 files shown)


src/tabpfn/preprocessing/steps/safe_power_transformer.py
```diff
@@ -141,24 +141,32 @@ def _yeojohnson_inverse_transform(x: np.ndarray, lmbda: float) -> np.ndarray:
     transform with parameter lambda.
     """
     dtype = x.dtype if np.issubdtype(x.dtype, np.floating) else np.float64
-    x_inv = np.zeros_like(x, dtype=dtype)
+    # Compute in at least float64 to avoid overflow when the inverse produces
+    # values exceeding a lower-precision dtype's range (e.g., float32), while
+    # preserving higher precision (e.g., float128) if the input provides it.
+    compute_dtype = np.promote_types(dtype, np.float64)
+    x_inv = np.zeros_like(x, dtype=compute_dtype)
```
bejaeger: I actually doubt we actually need this resolution. If it's really just to silence the warning, doesn't some clipping like below do the job?
In any case, happy to merge this, but could you do a quick check how much slower this transform is when running in float64?

simo-prior: @bejaeger makes sense, addressed 👍

PR #840, opened 2026-03-25
Make 2.6 the default model and improve HF auth error message

bejaeger:


bejaeger:

- #840 ◀

src/tabpfn/architectures/tabpfn_v2_5.py
```diff
@@ -866,7 +866,7 @@ def get_architecture(
     return TabPFNV2p5(
         config=config,
         n_out=n_out,
-        task_type="multiclass" if config.max_num_classes > 2 else "regression",
+        task_type="multiclass" if config.max_num_classes >= 2 else "regression",
```
psinger-prior: is this change intended?

src/tabpfn/model_loading.py
```diff
@@ -624,13 +624,15 @@ def load_model_criterion_config(
...
             )
             if res != "ok":
-                repo_type = "clf" if which == "classifier" else "reg"
-                raise RuntimeError(
-                    f"Failed to download model to {path}!\n\n"
-                    f"For offline usage, please download the model manually from:\n"
-                    f"https://huggingface.co/Prior-Labs/TabPFN-v2-{repo_type}/resolve/main/{resolved_model_names[i]}\n\n"
-                    f"Then place it at: {path}",
-                ) from res[0]
+                if _version_has_direct_download_option(model_version):
```
psinger-prior, line 626: This function appears to be a bit redundant as it only checks for v2. Not sure we need it.

bejaeger: This is mostly a fn for readibility

src/tabpfn/architectures/tabpfn_v2_5.py
bejaeger: Yep! doesn't change any behavior but better safe than sorry.

bejaeger:
Thanks a lot for the quick review, @psinger-prior . There were a few hiccups with the tests I had to fix. Should be good to go now but you may want to have another look.

PR #843, opened 2026-03-27
Fix differentiable input & update consistency tests for v2.6

### LINKED ISSUE #842: tabpfn_v2_6 breaks for `differentiable_input=True`
### Describe the bug

Running V2_6 with `differentiable_input=True` will fail in `remove_non_differentiable_preprocessing_from_models`, since the single file implementation of V2_6 has `target_embedder` instead of `y_encoder` and thus we can't iterate over y_encoder here:
https://github.com/PriorLabs/TabPFN/blob/f22aaf77f22c4b9bdbdd6b48b1cc9e0128c615b3/src/tabpfn/utils.py#L373-L377
The single file implementation of V2_5 also has `target_embedder` instead `y_encoder`, but running the code above will not fail for `ModelVersion.V2_5` since the old multi-file implementation is still being used.

Adding this
```py
        if not hasattr(model, "y_encoder"):
            continue
```
or skipping the call here
https://github.com/PriorLabs/TabPFN/blob/f22aaf77f22c4b9bdbdd6b48b1cc9e0128c615b3/src/tabpfn/classifier.py#L884
based on the version would fix that, I can create a PR if you tell me which fix you would prefer.

btw the test doesn't catch this since it only tests V2_5
https://github.com/PriorLabs/TabPFN/blob/f22aaf77f22c4b9bdbdd6b48b1cc9e0128c615b3/tests/test_consistency.py#L139-L144

### Steps/Code to Reproduce

```python
import torch
from sklearn.datasets import load_breast_cancer
from tabpfn.constants import ModelVersion

X, y = load_breast_cancer(return_X_y=True)
X, y = torch.as_tensor(X, dtype=torch.float32), torch.as_tensor(y, dtype=torch.int64)

model = TabPFNClassifier.create_default_for_version(
    ModelVersion.V2_6,
    n_estimators=1,
    device="cpu",
    differentiable_input=True
)

model.fit_with_differentiable_input(X, y)
```

### Expected Results

No error thrown

### Actual Results

```
Traceback (most recent call last):
  File "/Users/pfeffera/Downloads/repos/TabPFN/src/repro_bug.py", line 16, in <module>
    model.fit_with_differentiable_input(X, y)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/pfeffera/Library/Python/3.13/lib/python/site-packages/tabpfn_common_utils/telemetry/core/decorators.py", line 320, in wrapper
    return _safe_call_with_telemetry(
        func, args, kwargs, model_method, param_names
    )
  File "/Users/pfeffera/Library/Python/3.13/lib/python/site-packages/tabpfn_common_utils/telemetry/core/decorators.py", line 364, in _safe_call_with_telemetry
    result = func(*args, **kwargs)
  File "/Users/pfeffera/Downloads/repos/TabPFN/src/tabpfn/classifier.py", line 884, in fit_with_differentiable_input
    remove_non_differentiable_preprocessing_from_models(models=self.models_)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pfeffera/Downloads/repos/TabPFN/src/tabpfn/utils.py", line 376, in remove_non_differentiable_preprocessing_from_models
    for module in model.y_encoder:
                  ^^^^^^^^^^^^^^^
  File "/Users/pfeffera/Library/Python/3.13/lib/python/site-packages/torch/nn/modules/module.py", line 1962, in __getattr__
    raise AttributeError(
        f"'{type(self).__name__}' object has no attribute '{name}'"
    )
AttributeError: 'TabPFNV2p6' object has no attribute 'y_encoder'
```

### Versions

```shell
Dependency Versions:
--------------------
tabpfn: 7.0.1
torch: 2.11.0
numpy: 2.4.3
scipy: 1.17.1
pandas: 3.0.1
scikit-learn: 1.8.0
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 1.8.0
```

bejaeger:
Fixes #842


tests/test_consistency.py
oscarkey, line 1: should we keep the v2 tests?

bejaeger: Yep!

oscarkey (approved):
thanks!

PR #844, opened 2026-03-27
Forward the task type to forward.

priorphil:


src/tabpfn/model_loading.py
```diff
@@ -896,6 +906,19 @@ def load_model(
...
+    )
+
+    if "test_targets_MB" in inspect.signature(model.forward).parameters:
+        # The model computes the loss internally. Support for this was only added after
+        # v2.5, so we can safely assume that the inference config is stored in the
+        # checkpoint.
+        model.load_state_dict(full_state)
+        model.eval()
+        inference_config = InferenceConfig(**(checkpoint["inference_config"]))
+        return model, None, model_config, inference_config
```
oscarkey: maybe
```python
criterion = None
return model, criterion, ...
```

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "tabpfn"
-version = "7.0.1"
+version = "7.0.2"
```
oscarkey: should this bump be here?

priorphil: I thought I'd bump given the API change? But can revert if you prefer.

oscarkey: usually we do the bump when we release the package, but I'm not sure if that's the correct way to do things...

priorphil: Okay, will revert :)

PR #847, opened 2026-03-30
UV exclude newer

psinger-prior:
Adds `exclude-newer = "7 days"` to `pyproj.toml` for ignoring fresh packages on pypi. 

Only real caveat is that the relative date will only work with newer `uv` versions.


adrian-prior (approved):
LGTM!

PR #851, opened 2026-04-01
Balanced feature subsampling

bejaeger:
- Refactor preprocessing pipeline creation; everything is handled in TabPFNEnsemblePreprocessor now.
- Adds new feature subsampling methods.
- Still defaults to random feature subsampling.


src/tabpfn/finetuning/data_util.py
```diff
@@ -436,29 +434,26 @@ def __getitem__(self, index: int) -> ClassifierBatch | RegressorBatch:  # noqa:
...
+        ensemble_preprocessor = TabPFNEnsemblePreprocessor(
             configs=conf,
-            X_train=x_train_raw,
-            y_train=y_train,
+            n_samples=x_train_raw.shape[0],
             feature_schema=feature_schema,
             random_state=self.random_state,
             n_preprocessing_jobs=self.n_preprocessing_jobs,
-            parallel_mode="block",
         )
```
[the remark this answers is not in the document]
bejaeger: Should be okay.

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -76,41 +84,56 @@ def __init__(
...
+        self.random_state = random_state
+        self.static_seed, self.rng = infer_random_state(random_state)
+
+        pipeline_seeds = self.rng.integers(0, np.iinfo(np.int32).max, len(self.configs))
+        self.pipelines = [
+            create_preprocessing_pipeline(config, random_state=int(seed))
+            for config, seed in zip(self.configs, pipeline_seeds)
+        ]
+
+        max_features = self.configs[0].preprocess_config.max_features_per_estimator
```
LeoGrin: we have some presets (like 2.6 with different max_features_per_estimator for each config), here we would rely on the first one. If it's annoying, maybe we just don't allow this for now and change it in 2.6?

changelog/851.added.md
```diff
@@ -0,0 +1 @@
+Add balanced feature subsampling across ensemble members when the number of features exceeds `max_features_per_estimator`, ensuring each feature appears approximately equally across estimators.
```
LeoGrin: maybe say this is opt-in and explain how to turn it on?

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -153,6 +178,7 @@ def fit_transform_ensemble_members_iterator(
                 X_train=X_train_preprocessed,
                 y_train=y_train_preprocessed,
                 feature_schema=feature_schema_preprocessed,
+                feature_indices=self.subsample_feature_indices[i],
```
LeoGrin: as in the other PR, this will break for `parallel_mode="as-ready"`

tests/test_preprocessing/test_ensemble.py
```diff
@@ -52,3 +71,238 @@ def test__get_subsample_indices_for_estimators():
...
+        rng=rng,
+        feature_subsampling_method=FeatureSubsamplingMethod.RANDOM,
+    )
+
+    assert len(result) == 2
+    assert result[0] is None
+    assert result[1] is None
+
+
+def test__get_subsample_feature_indices__subsampling_needed():
```
LeoGrin, line 97: nit: might be interesting to have a slightly stronger test here in addition, e.g     - 8 estimators × size 60 over 100 features → assert each feature's appearance count is in {4, 5}

LeoGrin: nit: might also be nice to have another test asserting that, when two BALANCED runs are seeded identically, the outputs match (reproducibility), and that different seeds give different distributions.

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
LeoGrin: this can be removed now right?

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -332,6 +358,244 @@ def _generate_class_permutations(
...
+
+    # Check if all features already fit.
+    total = n_total + pipeline.num_added_features(n_samples, feature_schema)
+    if total <= max_features_per_estimator:
+        return n_total
+
+    lo, hi = 0, n_total
+    while lo < hi:
+        mid = (lo + hi + 1) // 2
+        sliced_schema = feature_schema.slice_for_indices(list(range(mid)))
```
LeoGrin: nit: I think for now it's fine with our configs option, but the way we compute this here might become an issue when we have transform which create more feature depending on the feature type.
For instance, let's say the first features of X are all categorical, and the last features are numerical. If we use one-hot (and count their features), we would underestimate max_input_features here (compared to what the estimator will be able to handle when he gets an input with shuffled features).

One slightly better solution might be to shuffle the schema in this function so we're less likely to get pathological cases. But there might be a better way still?

src/tabpfn/preprocessing/transform.py
```diff
@@ -49,28 +50,37 @@ def _fit_preprocessing_one(
...
+        feature_schema = feature_schema.slice_for_indices(feature_indices.tolist())
     if not isinstance(X_train, torch.Tensor):
         X_train = X_train.copy()
         y_train = y_train.copy()
 
-    preprocessor = create_preprocessing_pipeline(config, random_state=static_seed)
-    res = preprocessor.fit_transform(X_train, feature_schema)
+    static_seed, _ = infer_random_state(random_state)
+
+    if pipeline is None:
```
LeoGrin: I think we might not need that? If we make it mandatory we can clean up some more stuff:

  - transform.py: remove the branch, random_state param, seeds, and the pipeline is None fallback. Make pipeline and pipelines required.
  - ensemble.py: remove override_random_state param and self.static_seed.                                                                                                                                           
  - inference.py: remove the override_random_state=... kwarg.

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -76,41 +84,56 @@ def __init__(
...
-        during the fit_transform*() methods. Currently it is only used
-        in the InferenceEngineOnDemand class.
-        """
-        static_seed, _ = infer_random_state(self.random_state)
-        return static_seed
+        self.random_state = random_state
+        self.static_seed, self.rng = infer_random_state(random_state)
+
+        pipeline_seeds = self.rng.integers(0, np.iinfo(np.int32).max, len(self.configs))
+        self.pipelines = [
```
LeoGrin, line 119: now that we're defining these pipelines there and potentially  reusing them several time (IIUC in `fit_mode="low_memory"` at least?), we rely on our transforms being refit-safe (i.e giving the same results if you fit twice). I think we should be able to rely on that (sklearn contract), but apparently it's not the case for all our preprocessing transforms:
```
  For ReshapeFeatureDistributionsStep (#/Users/leo/VSCProjects/TabPFN/src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py#L186), append_to_original starts as the config value, which may be "auto". On fit, the step computes the decision for the current dataset and then
  overwrites self.append_to_original with a concrete True or False. After that, the original "auto" is gone. So if the same pipeline is fit again on a different dataset, the second fit no longer recomputes the auto decision; it just reuses the stale bool from the first fit. That is
  how I got second_reused=1000 vs second_fresh=500: the first dataset had few effective features after constant-feature removal, so "auto" became True and stayed True; the second dataset should have decided False, but the reused step still appended originals and doubled the feature
  count.

  For AddSVDFeaturesStep (#/Users/leo/VSCProjects/TabPFN/src/tabpfn/preprocessing/steps/add_svd_features_step.py#L66), the same pattern exists with self.is_no_op. If one fit sees fewer than 2 input features, it sets is_no_op = True. A later fit on richer data never resets that flag to
  False. Then _transform() checks the stale flag and skips adding SVD columns entirely, even though the current fit built a valid transformer. That is how I got second=2 vs fresh=3: first fit collapsed to one effective feature, marking the step no-op; second fit had two varying
  features and should have added one SVD component, but the reused step still short-circuited.
```

ideally we could also tests that all our estimators are refit safe?

LeoGrin (approved):
Looks great thanks a lot!

Nit: maybe we could add an end to end test that our estimator works with balanced feature subsampling on, and with features > max_features per estimators?

src/tabpfn/preprocessing/ensemble.py
bejaeger: good one! passing a list of max_features to the subsample method now.

changelog/851.added.md
bejaeger: changed it! I'm wondering if we should make "balanced" the default?
I'll leave as opt-in for now but when we test this a bit we should probably choose one of the better methods.

src/tabpfn/preprocessing/ensemble.py
bejaeger: Thanks! changed

tests/test_preprocessing/test_ensemble.py
bejaeger: Done!

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: Yep, slightly scary but should be okay.

src/tabpfn/preprocessing/ensemble.py
bejaeger: Yeah, the one-hots are a really pain here. Also the binary search was flawed in regard to the `append_original` flag. I changed it!

For one-hot, it would be a bit of a bigger change because we need information about the content of the features. This would also complicate the round-robin selection quite a bit! depending on which feature we sample, the output_features will be different :-/
I would leave that for follow-up work for now.

src/tabpfn/preprocessing/transform.py
bejaeger: Yep, better to be explicit! thanks

src/tabpfn/preprocessing/ensemble.py
bejaeger: Nice one!
Made them refit-safe and added tests for it that they are.

PR #852, opened 2026-04-01
Fix two KV cache fit mode bugs so output equal to fit_preprocessors.

oscarkey:
1. Magnus's work (see https://github.com/PriorLabs/TabPFN/pull/784): only add thinking tokens during fit, rather than fit and predict
2. If `force_inference_dtype` is set, cast the model during fit rather than just during predict
  
Re-enable the tests that `fit_preprocessors` and `fit_with_cache` are equal.

Part of PRI-154
Fixes PRI-260


bejaeger (approved):
LGTM!

PR #861, opened 2026-04-06
[WIP] Add ensemble_temperature to fix distributional calibration

noahho:
## Context

@Jonaslandsgesell shared a paper and benchmark ([scoringbench.bolt.host](https://scoringbench.bolt.host)) evaluating TabPFN on proper scoring rules (NLL/log-score, CRPS, CRLS, IS95). The results were surprisingly bad on distributional metrics — TabPFN v2.6 ranked poorly on log-score and CRPS despite strong RMSE/R2 numbers. This motivated a closer look at why.

In the follow-up email thread, Jonas pointed out that TabPFN v2.6's **sharpness decreased** compared to earlier versions (i.e. predictions got wider/more uncertain), which hurt log-score (which rewards sharp, correct distributions). At the same time, RMSE improved — a classic sign of conflicting inductive biases across scoring rules, as described in Jonas's paper ([arxiv 2603.08206](https://arxiv.org/abs/2603.08206)).

## Root cause

The default `softmax_temperature=0.9` **sharpens** each estimator's predicted distribution before ensembling. This was tuned for RMSE quality — it helps ensemble members mix better and improves point estimates. But it systematically **over-sharpens** the final predictive distribution, breaking calibration for all distributional metrics that depend on shape (NLL, interval coverage, PIT uniformity).

## Experimental results (regression)

Controlled experiment on 100-train / 200-test Gaussian linear regression, 8 estimators:

| Config | NLL | CRPS | CRLS | IS95 | Cov95 | KS p-val | RMSE |
|--------|-----|------|------|------|-------|----------|------|
| **Old default** `sm_t=0.9, ens_t=1.0` | 1.933 | 0.614 | 1.273 | 5.060 | 0.950 | 0.30 | **1.056** |
| Neutral `sm_t=1.0, ens_t=1.0` | 1.846 | 0.614 | 1.230 | 5.098 | 0.955 | 0.52 | 1.059 |
| **New default** `sm_t=0.9, ens_t=1/0.9` | **1.845** | 0.614 | **1.230** | 5.103 | 0.955 | **0.52** | 1.060 |

Key findings:
- **NLL improves ~4.5%** (1.933 → 1.845): the main signal that temperature is miscalibrated
- **KS p-value doubles** (0.30 → 0.52): PIT uniformity — calibrated distributions should be uniform, 0.30 is poor
- **CRPS is essentially flat** (~0.614 across all configs): CRPS is not sensitive to temperature in this setting because the bin-distribution representation is wide enough that the sharpening/flattening effect on CRPS is negligible at this scale. NLL and KS_p are the meaningful calibration signals here.
- **RMSE cost is negligible**: 0.004 difference, within noise on real benchmarks

## Solution

Add `ensemble_temperature` (default `1/0.9 ≈ 1.111`) as a final post-hoc recalibration step applied to the ensemble-averaged log-probabilities:

```python
# In predict(), after existing averaging logic:
if self.ensemble_temperature != 1.0:
    logits = logits / self.ensemble_temperature
```

- `softmax_temperature=0.9` is kept unchanged — it still does per-estimator sharpening for better ensemble mixing and RMSE
- `ensemble_temperature=1/0.9` compensates for the over-sharpening at the distribution level
- Users can tune `ensemble_temperature` independently to target calibration vs sharpness tradeoffs

**Alternative**: simply change `softmax_temperature` default to `1.0` and remove the complexity entirely. The RMSE difference is tiny (0.004). Simpler, but a breaking change for existing users.

## FYI: Classification calibration looks fine (and interesting)

We ran the same temperature sweep on a 3-class Gaussian classification problem to check if the same issue exists there:

| Config | NLL | Brier | ECE | Acc |
|--------|-----|-------|-----|-----|
| `sm_t=0.9` (default) | 0.558 | 0.285 | 0.244 | 0.884 |
| `sm_t=1.0` (neutral) | 0.574 | 0.295 | 0.259 | 0.884 |
| `sm_t=0.7` (sharper) | **0.529** | **0.264** | **0.213** | 0.884 |
| `sm_t=1.5` (flatter) | 0.647 | 0.346 | 0.322 | 0.874 |

The picture is the **opposite** of regression: for classification, lower temperature consistently improves NLL, Brier score, and ECE. The model is *under-confident* in classification (predicting 0.7 when it should predict 0.9 for well-separated classes), and `sm_t=0.9` corrects this in the right direction.

**Note:** the classifier already has an opt-in `calibrate_temperature=True` via `tuning_config` that finds the optimal temperature data-adaptively on a holdout set. The regressor currently has no equivalent — this PR's `ensemble_temperature` is a fixed heuristic. A future improvement would be to add the same data-driven calibration to the regressor (see checkbox below).

Also worth considering: **adding NLL (log-loss) to the meta-loss for classification** — the results above suggest we're systematically under-confident and could improve calibration directly in training rather than correcting post-hoc.

## FYI @LeoGrin: relevance for time series

Time series forecasting benchmarks almost universally evaluate on **CRPS** as the primary probabilistic metric. If TabPFN is being used or benchmarked as a time series foundation model, the regression miscalibration fixed here directly affects those rankings.

The `ensemble_temperature` knob introduced here (or better: a data-driven version of it, see checkbox below) could be a lightweight way to optimise for CRPS specifically at inference time without retraining — similar to how the classifier's `calibrate_temperature` already tunes for log-loss on a holdout split.

## WIP notes / open questions

- [ ] **Implementation is minimal** — needs review on whether this interacts correctly with fine-tuning paths
- [ ] **Alternative**: simply set `softmax_temperature=1.0` as default — cleaner but a breaking change
- [ ] @LennartPurucker: could we add NLL/log-loss (and ideally CRPS) to the meta-loss evaluation? Both for regression (would directly catch the calibration issue fixed here) and for classification (the results above suggest the model is systematically under-confident — worth optimising for directly rather than correcting post-hoc)
- [ ] **Data-driven `ensemble_temperature` calibration for the regressor**: mirror the classifier's `calibrate_temperature=True` / `find_optimal_temperature()` pattern — use a holdout split to find the `ensemble_temperature` that minimises NLL or CRPS, and expose it via `tuning_config`. This would replace the current fixed `1/0.9` heuristic with a principled, dataset-adaptive value — especially valuable for time series use cases where CRPS is the target metric

cc @LeoGrin @bejaeger @LennartPurucker @Jonaslandsgesell

### DIFF
--- a/scripts/check_distributional_calibration.py
+++ b/scripts/check_distributional_calibration.py
@@ -1,0 +1,163 @@
+"""Diagnostic script to verify whether softmax_temperature and average_before_softmax
+affect distributional calibration metrics.
+
+Hypothesis (from Jonas Landsgesell paper + email thread):
+- softmax_temperature=0.9 sharpens distributions -> hurts log-score / CRPS / CRLS
+- averaging probabilities vs logits before ensembling may matter for calibration
+
+Metrics:
+  NLL   - Negative log-likelihood (log score), sensitive to sharpness
+  CRPS  - Continuous Ranked Probability Score, integral over quantile losses
+  CRLS  - Continuous Ranked Log Score = CRPS + NLL / 2 (Brehmer & Gneiting 2021)
+  IS95  - Interval Score at 95%: penalises width + miscoverage (lower = better)
+  Cov95 - Empirical coverage of 95% prediction interval (target: 0.95)
+  Sharp - Mean width of 95% PI (lower = sharper, but only good if well-calibrated)
+  MACE  - Mean Absolute Calibration Error from PIT (lower = better)
+  KS_p  - p-value of KS test for PIT uniformity (higher = better calibrated)
+  RMSE  - Root mean squared error of the mean prediction
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import torch
+from scipy.stats import kstest
+
+from tabpfn import TabPFNRegressor
+
+
+# ---------------------------------------------------------------------------
+# Metric helpers
+# ---------------------------------------------------------------------------
+
+def compute_pit(criterion, logits: torch.Tensor, y: np.ndarray) -> np.ndarray:
+    """P(Y <= y_true) under predicted distribution. Uniform => calibrated."""
+    y_t = torch.as_tensor(y, dtype=logits.dtype, device=logits.device).unsqueeze(-1)
+    return criterion.cdf(logits, y_t).squeeze(-1).cpu().detach().numpy()
+
+
+def compute_nll(criterion, logits: torch.Tensor, y: np.ndarray) -> float:
+    """Mean negative log-likelihood (log score)."""
+    y_t = torch.as_tensor(y, dtype=logits.dtype, device=logits.device)
+    return criterion(logits, y_t).mean().item()
+
+
+def compute_crps(criterion, logits: torch.Tensor, y: np.ndarray) -> float:
+    """CRPS via quantile decomposition: E_q[(F^{-1}(q) - y)*(q - 1{y<=F^{-1}(q)})]."""
+    quantile_levels = np.linspace(0.01, 0.99, 99)
+    crps_sum = 0.0
+    for q in quantile_levels:
+        q_pred = criterion.icdf(logits, q).cpu().detach().numpy()
+        indicator = (y <= q_pred).astype(float)
+        crps_sum += np.mean((indicator - q) ** 2)
+    return crps_sum / len(quantile_levels)
+
+
+def compute_crls(crps: float, nll: float) -> float:
+    """Continuous Ranked Log Score (Brehmer & Gneiting 2021).
+    Combines sharpness of log score with calibration of CRPS:
+      CRLS = (CRPS + NLL) / 2
+    """
+    return (crps + nll) / 2
+
+
+def compute_is95(criterion, logits: torch.Tensor, y: np.ndarray) -> tuple[float, float, float]:
+    """Interval Score at 95% PI.
+    IS_alpha = (u - l) + (2/alpha) * [max(0, l-y) + max(0, y-u)]
+    """
+    alpha = 0.05
+    l = criterion.icdf(logits, alpha / 2).cpu().detach().numpy()
+    u = criterion.icdf(logits, 1 - alpha / 2).cpu().detach().numpy()
+    width = u - l
+    penalty = (2 / alpha) * (np.maximum(0, l - y) + np.maximum(0, y - u))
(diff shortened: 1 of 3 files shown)


noahho:
@LeoGrin @bejaeger what do you think about the proposed change? Essentially we are trading off a small difference in RMSE for better calibration of the quantiles - something that we didn't really consider before.

Potentially the best solution is keeping the code as is but considering calibration when we create the final ensemble configuration.

bejaeger:
IMO it makes sense to change the default to 1.0 and avoid adding extra complexity with an additional temperature parameter. 

As you say, the change in mse is very small (also what I have seen in the past) and we can promote the temperature tuning workflow a bit more.

Also, it would be interesting to see how the calibration changes for models trained with different losses.

noahho:
@Jonaslandsgesell would you be able to re-run your scoringbench evaluation with \`softmax_temperature=1.0\`? Based on our experiments, this should significantly improve the distributional calibration metrics (NLL, KS p-value) with only a negligible RMSE cost.

You can set it like this:

```python

model = TabPFNRegressor(softmax_temperature=1.0)
```

Context: the default \`softmax_temperature=0.9\` was tuned for RMSE but systematically over-sharpens the predictive distribution, which hurts log-score/NLL and interval calibration. Setting it to 1.0 is effectively neutral and avoids this bias.

Would be very helpful to see how this changes the scoringbench results!

scripts/check_distributional_calibration.py
```diff
@@ -0,0 +1,163 @@
...
+        crps_sum += np.mean((indicator - q) ** 2)
+    return crps_sum / len(quantile_levels)
+
+
+def compute_crls(crps: float, nll: float) -> float:
+    """Continuous Ranked Log Score (Brehmer & Gneiting 2021).
+    Combines sharpness of log score with calibration of CRPS:
+      CRLS = (CRPS + NLL) / 2
+    """
+    return (crps + nll) / 2
```
jonaslandsgesell, line 61: Different to definition of CRLS (aka Exceedance Probability Score) https://link.springer.com/article/10.1080/15598608.2012.695663

jonaslandsgesell:
@noahho, when rerunning ScoringBench with TabPFN v2.6 using softmax_temperature=1.0, the performance is essentially on par with the default v2.6 on the ~100 OpenML datasets. The adapted pipeline ranks very closely in the critical difference (CD) diagrams (CRPS, R^2 and others). I can provide the CD plots if you’d like to see them, though the results are effectively the same.

noahho:
Thanks @Jonaslandsgesell! Really appreciate you running the benchmark so quickly.

PR #863, opened 2026-04-07
Smarter model caching - reload when file metadata changes

bejaeger:
Since we have pipelines (finetuning, training, ...) where the name of the checkpoint stays the same while the underlying weights change, we want to load the checkpoint again even though it has the same file path. This adds a metadata check to the model loading cache to avoid using a cached model in cases where the model file was updated.


bejaeger:

- #863 ◀

LeoGrin (approved):
LGTM, thanks!

psinger-prior:
Thanks @bejaeger - I wonder if the is some (future) way of having this still enabled automatically but do not check the filename of the weights, but via hash or similar approach.

bejaeger:
Thanks @psinger-prior for the suggestion. Changed the PR to do exactly that and make the caching smarter. 
PTAL

psinger-prior (approved):
Thanks a lot @bejaeger - lgtm!

One small comment (I think we can leave it as-is): the cache in theory can accumulate many weights if we replace them many times, we could safeguard against that with a max size, but calling the clear cache manually also solves it, don't think this is a realistic concern.

bejaeger:
Actually I think that's a good point, thanks @psinger-prior . Restricted the cache size to 1 because in the scenario where we don't want to cache (e.g. during finetuning) we don't need to go back to a previous checkpoint and especially don't need it to be in the cache.

PR #867, opened 2026-04-08
Add speed metrics to inference engines to measure execution speed

bejaeger:


bejaeger:

- #867 ◀

src/tabpfn/inference.py
```diff
@@ -384,28 +387,43 @@ def iter_outputs(
                 override_random_state=np.random.default_rng(self.static_seed),
             )
         )
+        preprocessed = [
+            (em, em.transform_X_test(X)) for em in ensemble_members_iterator
+        ]
```
oscarkey: Does this list mean the evaluation of the preprocessing and model is no longer overlapped? Maybe the overlap wasn't working anyway...

I'm wondering if some sort of `TimedIterator` wrapper class might be helpful for both avoiding this problem and reducing boilerplate in some of the other spots?

bejaeger: Good idea!
Changed it so the preprocessing overlaps with forward again and added a `TimedIterator`. I decided to drop measuring the `predict_preprocessing_seconds` because it would have created some awkward time subtraction but I think this is negligible anyways so it's fine.

src/tabpfn/inference.py
```diff
@@ -94,6 +118,7 @@ def __init__(
         self.save_peak_mem = save_peak_mem
         self.dtype_byte_size = dtype_byte_size
         self.force_inference_dtype = force_inference_dtype
+        self._speed_metrics: dict[str, float] = {}
```
oscarkey, line 121: should `speed_metrics` be public? Atm there's no way to access it.

src/tabpfn/inference.py
```diff
@@ -794,10 +842,13 @@ def __init__(  # noqa: PLR0913
 
         ens_models: list[Architecture] = []
         ensemble_members: list[TabPFNEnsembleMember] = []
+        fit_preprocess_time = 0.0
+        fit_forward_time = 0.0
 
         for ensemble_member in ensemble_members_iterator:
             ensemble_members.append(ensemble_member)
 
+            preprocess_start = time.perf_counter()
```
oscarkey: just checking that this deliberately includes the model copy time? Fine by me if it's meant to.

oscarkey (approved):
nice!

src/tabpfn/inference.py
bejaeger: I made it private on purpose because it's more of a dev feature. Happy to change it if you think it's better.

oscarkey: ahh i see! This way is also fine for me then :)

src/tabpfn/inference.py
bejaeger: Thanks! Also noticed that this wasn't measuring the cpu preprocessing timing correctly. Fixed it and merging now.

PR #879, opened 2026-04-14
Add torch compile performance option.

priorphil:


src/tabpfn/architectures/interface.py
```diff
@@ -92,6 +92,14 @@ class PerformanceOptions:
     resident in memory at once.
     """
 
+    enable_torch_compile: bool = False
+    """If set to True, the model may decide to compile all or selected parts.
+
+    Setting this to `True` can enable speedups for repeated inference but may
+    result in longer inference time for the first forward pass, during which
+    compile and autotune will be run. Tuning results are cached, so should
+    persist across runs."""
```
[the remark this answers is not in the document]
priorphil: 1. I'm currently implementing this as having two functions, a compiled one and a non-compiled one, taking different paths in the model, so no state change.
2. The field has a default and will be enabled in future architectures.
3. We already need > 2.0 for sdpa

changelog/879.added.md
```diff
@@ -0,0 +1 @@
+Add enable_torch_compile to PerformanceOptions.
```
bejaeger, line 1: since this doesn't do anything at the moment, should we defer this changelog to the PR that activates it?

bejaeger (approved):
Please see comment, otherwise LGTM!

changelog/879.added.md
priorphil: We could, but not sure that would change anything? Still not all models would support this flag / change behaviour,?

priorphil: Also, I can't submit this without a changelog  due to the check ;)

PR #884, opened 2026-04-20
Add gpu preprocessing

bejaeger:
Also adds new inference engine for new kv cache interface.


src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -235,6 +243,24 @@ def _create_transformers_and_new_schema(
             num_columns=n_output_features,
         )
 
+        if self.schedule_quantile_for_gpu:
+            if self.append_to_original:
```
LeoGrin: it should be self.append_to_original_decision_, otherwise we have an issue if `self.append_to_original` is "auto" but resolves to False.

IIUC in this case it fails silently:

>   What the model gets depends on the preset, but the common outcome is:

>  - The CPU reshape step is switched from the intended quantile transform to "none" in src/tabpfn/preprocessing/pipeline_factory.py:80.
>  - Because no columns are marked in src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py:246, the GPU pipeline does not add TorchSelectiveQuantileTransformerStep.
>  - So the quantile transform is skipped entirely for both train and test data.

LeoGrin: (might also make sense to add a test which would have caught this?)

src/tabpfn/inference_config.py
```diff
@@ -128,9 +128,20 @@ class InferenceConfig:
             indices are repeated for the remaining estimators.
     """
 
+    ENABLE_GPU_PREPROCESSING: bool = False
```
LeoGrin, line 131: default here doesn't match the docstring

LeoGrin (approved):
LGTM thanks! Found one issue I think.

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: Thanks for catching this! fixed it

PR #887, opened 2026-04-22
Make GPU transforms more memory efficient

bejaeger:
image

### DIFF
--- a/src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
+++ b/src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
@@ -178,8 +178,7 @@ class TorchQuantileTransformer:
         """Transform a single chunk, preserving NaN positions."""
         nan_mask = torch.isnan(x)
         result = self._interpolate(x, quantiles, references)
-        nan_fill = torch.tensor(float("nan"), device=x.device, dtype=x.dtype)
-        return torch.where(nan_mask, nan_fill, result)
+        return torch.where(nan_mask, float("nan"), result)
 
     def _get_transform_chunk_size(self, x: torch.Tensor) -> int:
         """Compute a row-chunk size that keeps intermediate memory bounded.
--- a/src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
+++ b/src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
@@ -62,7 +62,7 @@ class TorchQuantileTransformer:
             0, 1, n_quantiles_effective, device=x.device, dtype=compute_dtype
         )
 
-        x_compute = x.to(compute_dtype) if x.dtype != compute_dtype else x
+        x_compute = x.to(compute_dtype)
         quantiles = self._nanquantile_chunked(x_compute, references)
 
         # Ensure monotonicity (handle floating point issues)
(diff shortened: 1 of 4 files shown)


bejaeger:

- #887 ◀

src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
```diff
@@ -61,20 +61,79 @@ def fit(self, x: torch.Tensor) -> dict[str, torch.Tensor]:
         references = torch.linspace(
             0, 1, n_quantiles_effective, device=x.device, dtype=compute_dtype
         )
-        quantiles = torch.nanquantile(x.to(compute_dtype), references, dim=0)
+
+        x_compute = x.to(compute_dtype) if x.dtype != compute_dtype else x
```
oscarkey: iiuc `.to()` already does this if statement internally

src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
```diff
@@ -61,20 +61,79 @@ def fit(self, x: torch.Tensor) -> dict[str, torch.Tensor]:
...
+        n_features = x_flat.shape[1] if x_flat.ndim > 1 else 1
+        chunk_cols = self._get_fit_chunk_cols(x_flat)
+        process_all_at_once = n_features <= chunk_cols
+
+        if process_all_at_once:
+            quantiles = torch.nanquantile(x, references, dim=0)
+        else:
+            chunks = []
+            for col_start in range(0, n_features, chunk_cols):
+                col_end = min(col_start + chunk_cols, n_features)
```
oscarkey: I think you can just slice to  `col_start + chunk_cols` and torch will do the min internally. But good with me if you prefer to be more explicit

oscarkey:
if you're maxed out then totally fine to keep it like this, but did you consider using `torch.vmap(chunk_size=blah)` rather than implementing the chunking manually? I just checked and it seems to be supported on torch==2.5

bejaeger:
Yep, using it in torch svd now to reduce line count! :) 
For torch quantile I stick to the manual way because "PyTorch emits a 'performance drop' warning and falls back to a slower path".

src/tabpfn/preprocessing/torch/torch_svd.py
```diff
@@ -252,40 +302,40 @@ def transform(
...
         orig_dtype = x.dtype
         compute_dtype = std.dtype
-        x_compute = x.to(compute_dtype) if x.dtype != compute_dtype else x
-
-        # Replace inf with nan before scaling
-        x_safe = torch.where(
-            torch.isinf(x_compute),
-            torch.tensor(float("nan"), device=x.device, dtype=compute_dtype),
-            x_compute,
+        chunk_size = self._get_transform_chunk_size(x)
```
oscarkey: do you need to compute this after casting to the compute dtype?

tests/test_torch_preprocessing/test_torch_quantile_transformer.py
oscarkey, line 1: should there be a `TestChunkedTorchSafeStandardScalerEquivalence`?

oscarkey (approved):
great savings!!

src/tabpfn/preprocessing/torch/torch_svd.py
bejaeger: passing the compute element size now, thanks!

tests/test_torch_preprocessing/test_torch_quantile_transformer.py
bejaeger: done!

src/tabpfn/preprocessing/torch/torch_quantile_transformer.py
bejaeger: done

PR #888, opened 2026-04-22
[PRI-277] Fix v2.6 low accuracy on MPS

adrian-prior:
## Summary

On MPS, the default v2.6 classifier/regressor produced near-random outputs (e.g. Iris multiclass accuracy 0.48 vs 0.98 on CPU). Root cause is a PyTorch MPS bug: `F.scaled_dot_product_attention` silently returns wrong values when given non-contiguous q/k/v produced by a `permute`. Every v2.6 attention block passes permute views straight into SDPA, so the bug fires on every forward.

v2.5 isn't affected because its default checkpoint loads via the older `PerFeatureTransformer`, which uses a different attention implementation.

**Verified against both task types** — the fix resolves the accuracy regression for classification *and* regression (details below).

## Fix

Two-line guard in `_batched_scaled_dot_product_attention` that calls `.contiguous()` on the permuted q/k/v only when the device is MPS. CUDA and CPU paths are untouched.

Upstream PyTorch bug with minimal reproducer: [pytorch/pytorch#181133](https://github.com/pytorch/pytorch/issues/181133). Once fixed upstream, we can drop this guard (gated on torch/macOS version).

Resolves [PRI-277](https://linear.app/priorlabs/issue/PRI-277/v26-has-low-accuracy-on-mps).

## Verification

Both examples were run on MPS (patched) and compared against CPU (`TABPFN_EXCLUDE_DEVICES=mps`):

- **Classification** — `examples/tabpfn_for_multiclass_classification.py` on MPS: Accuracy `0.48 → 0.98`, matches CPU.
- **Regression** — `examples/tabpfn_for_regression.py` on MPS: R² and all MAE values match CPU to 4+ decimals (pre-fix: MSE ~10% worse on MPS per the ticket).
- `tests/test_model/test_attention.py` — all pass.

## Test plan
- [ ] CI green on macOS runner (if one exists)
- [x] Spot-check Iris / regression example on a reviewer's Mac

### DIFF
--- a/changelog/888.fixed.md
+++ b/changelog/888.fixed.md
@@ -1,0 +1,1 @@
+Fix v2.6 producing near-random outputs on Apple Silicon (MPS). `F.scaled_dot_product_attention` on MPS silently returns wrong values for non-contiguous q/k/v (upstream: pytorch/pytorch#181133); we now force contiguity before the call. Iris multiclass accuracy on MPS: 0.48 → 0.98.
--- a/src/tabpfn/architectures/tabpfn_v2_6.py
+++ b/src/tabpfn/architectures/tabpfn_v2_6.py
@@ -282,6 +282,14 @@ def _batched_scaled_dot_product_attention(
     k_BJSD = k_BSJD.permute(0, 2, 1, 3)
     v_BJSD = v_BSJD.permute(0, 2, 1, 3)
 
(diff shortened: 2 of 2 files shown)


bejaeger (approved):
LGTM!

Issue #898, opened 2026-04-26
Add a progress bar


randommm:
### Describe the workflow you want to enable

A progress bar would be very useful when running with large dbs or many n_estimators.

### Describe your proposed solution

A progress bar would be very useful when running with large dbs or many n_estimators.

### Impact

None

PR #899, opened 2026-04-26
add progress bar

randommm:
## Issue
#898

## Motivation and Context

A progress bar would be very useful when running with large dbs or many n_estimators.

## Public API Changes

-   [ ] No Public API changes

Changes the public API but with changes that are backward and forward compatible.

## How Has This Been Tested?

Ran a regressor locally.

## Checklist

---

Fixes PRI-282

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -31,6 +31,7 @@ import torch
 from sklearn import config_context
 from sklearn.base import BaseEstimator, ClassifierMixin, check_is_fitted
 from tabpfn_common_utils.telemetry import track_model_call
+from tqdm.auto import tqdm
 
 from tabpfn.base import (
     ClassifierModelSpecs,
@@ -230,6 +231,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
         differentiable_input: bool = False,
         eval_metric: str | ClassifierEvalMetrics | None = None,
         tuning_config: dict | ClassifierTuningConfig | None = None,
+        show_progress_bar: bool = False,
     ) -> None:
         """Construct a TabPFN classifier.
 
@@ -453,6 +455,9 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
                 `eval_metric`. See
                 [tabpfn.inference_tuning.ClassifierTuningConfig][] for details
                 and options.
+
+            show_progress_bar:
+                Whether to show a progress bar during inference. Defaults to False.
         """
         super().__init__()
         self.n_estimators = n_estimators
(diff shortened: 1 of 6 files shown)


src/tabpfn/classifier.py
```diff
@@ -222,6 +223,7 @@ def __init__(  # noqa: PLR0913
             "fit_with_cache",
             "batched",
         ] = "fit_preprocessors",
+        show_progress_bar: bool = False,
```
oscarkey: Would you add this to the end of the parameter list? Just to avoid *args accidents, and to match the doc string.

src/tabpfn/classifier.py
```diff
@@ -1442,10 +1434,16 @@ def forward(  # noqa: C901, PLR0912
...
+        for output, config in tqdm(
+            self.executor_.iter_outputs(
+                X,
+                autocast=self.use_autocast_,
+                task_type="multiclass",
+            ),
+            total=self.n_estimators,
+            desc="TabPFN inference",
+            unit="estimator",
+            disable=not getattr(self, "show_progress_bar", False),
```
oscarkey:
```suggestion
            disable=self.show_progress_bar
```
is there ever a case where this attribute isn't set?

src/tabpfn/regressor.py
oscarkey, line 1: same comments as in `classifier.py`

tests/test_regressor_interface.py
```diff
@@ -40,6 +41,39 @@
...
+    model = TabPFNRegressor(show_progress_bar=True)
+    assert model.show_progress_bar is True
+    assert model.get_params()["show_progress_bar"] is True
+
+    default_model = TabPFNRegressor()
+    assert default_model.show_progress_bar is False
+    assert default_model.get_params()["show_progress_bar"] is False
+
+
+def test__forward__passes_progress_bar_flag(monkeypatch: pytest.MonkeyPatch) -> None:
```
oscarkey, line 70: Would you add tests for both the classifier and regressor that `.predict()` doesn't crash on a tiny dataset when `show_progress_bar=True`? I think we don't need to bother mocking tqdm as it's low enough stakes that the complexity isn't worth it.

oscarkey (changes requested):
Hey! Thanks so much for the contribution. This looks pretty good to me, I just have a few suggestions.

src/tabpfn/classifier.py
randommm: it was mostly for compatibility with past pickled models, but i guess it's unnecessary.

randommm:
this should probably have addressed all comments

oscarkey (approved):
lgtm! Thank you very much.

I think you'll need to rename your changelog to `899.added.md` to get the check to pass. And then I would recommend running the pre-commit hooks for the ruff failure.

randommm:
hopefully it goes now

Issue #905, opened 2026-04-29
Add ensemble_batch_size for single-device inference


randommm:
### Describe the workflow you want to enable

Add ensemble_batch_size for single-device inference

### Describe your proposed solution

On devices with large amounts of RAM like Strix Halo, this can greatly speed up results

### Impact

None

ggprior:
Hey @randommm , thanks for the issue and your direct contribution! While we closed #906 , we internally aligned this would be a great functionality to have, but we believe we can push this even further and combine it with #1045 that does something similar on the batch axis.
We'll tag you as coauthor once the final form of this feature is fully decided. Will keep this issue open and update you soon.

PR #906, opened 2026-04-29
Add ensemble_batch_size for single-device inference

randommm:
## Issue
#905

## Motivation and Context

On devices with large amounts of RAM like Strix Halo, this can greatly speed up results

## Public API Changes

-   [ ] No Public API changes

## How Has This Been Tested?

local testing

## Checklist

### DIFF
--- a/src/tabpfn/base.py
+++ b/src/tabpfn/base.py
@@ -286,6 +286,7 @@ def create_inference_engine(  # noqa: PLR0913
     forced_inference_dtype_: torch.dtype | None,
     memory_saving_mode: MemorySavingMode,
     use_autocast_: bool,
+    ensemble_batch_size: int | None = None,
     inference_mode: bool = True,
 ) -> InferenceEngine:
     """Create the appropriate TabPFN inference engine based on `fit_mode`.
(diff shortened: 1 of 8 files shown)


src/tabpfn/inference.py
```diff
@@ -726,6 +753,123 @@ def _transform_X_test(
...
+                    break
+
+                prepare_start = time.perf_counter()
+                prepared = _prepare_member(ensemble_member)
+                forward_time += time.perf_counter() - prepare_start
+                if not batch or _is_compatible_with_batch(batch[0], prepared):
+                    batch.append(prepared)
+                else:
+                    pending_prepared = prepared
+                    break
```
adrian-prior: Doesn't this assume that compatible inference configs will always appear in a row. Let's say we have two kind of inference configs A, B. Then [A, B, A, B, A, B] would be broken up into six batches?

src/tabpfn/base.py
```diff
@@ -332,6 +338,7 @@ def create_inference_engine(  # noqa: PLR0913
             force_inference_dtype=forced_inference_dtype_,
             save_peak_mem=memory_saving_mode,
             inference_mode=inference_mode,
+            ensemble_batch_size=ensemble_batch_size,
```
adrian-prior, line 342: Is there any reason why this only supports one of the four inference engines?

adrian-prior (commented):
Hey @randommm,

Thanks for setting this up, and apologies for taking so long to get back on this! I do have some open questions, which I left on the PR, and there are also a bunch of comments left by Gemini, of which I think many are valid. Would you mind going through them?

randommm:
@adrian-prior i think it should be all fixed now.

adrian-prior:
Hey @randommm,
Thanks for the continued effort! I chatted with a colleague today who is more knowledgeable in this part of the codebase, and I think we agree on two things: 
- (a) If this works well, we would want to have this feature part of the codebase
- (b) The current implementation blows up the complexity of the code base so much that we couldn't merge it in it's current form, even if (a) is fulfilled. In particular, we believe that enabling this will probably require a large-scale refactor of the inference codebase (which might be overdue on our side anyway).

Regarding (a): Do you have any runtime benchmarking results on a single GPU with different dataset shapes, showing what speed-ups we could expect, if this were implemented? Based on this data, I would love to take the discussion back to the [issue](https://github.com/PriorLabs/TabPFN/issues/905). There, we could decide collectively if (a) is fulfilled and then how the changes would possibly look (into which my colleagues or I could also look).

noahho:
Closing due to inactivity

PR #908, opened 2026-05-04
Importance-based feature subsampling

bejaeger:
- Adds importance-based feature subsampling
- Optional lightgbm to avoid a reasonably heavy dependency per default
- "auto" mode to activate it automatically for large datasets with >200 features.


src/tabpfn/preprocessing/ensemble.py
```diff
@@ -638,6 +712,247 @@ def _subsample_features_constant_and_balanced(
...
+    When the dataset is larger, ``n_subsamples = min(n_estimators,
+    n_samples // max_samples + 1)`` independent subsamples of size ``max_samples``
+    are drawn, each producing a different ordering.  The resulting orderings are
+    then cycled to fill the full list of length ``n_estimators``.
+    """
+    model_cls = _get_extra_trees_model_cls(task_type)
+    n_samples = len(X)
+
+    def _fit_ordering(X_fit: np.ndarray, y_fit: np.ndarray) -> np.ndarray:
+        seed = int(rng.integers(0, 2**31))
```
LennartPurucker: This max seed can crash on certain os. I think you need to use something like np.info(float).max (or however the call works).

LennartPurucker (approved):
LGTM

PR #909, opened 2026-05-04
Add V3 interface and model routing mechansim

bejaeger:


src/tabpfn/base.py
```diff
@@ -115,6 +123,8 @@ def initialize_tabpfn_model(
 
         which: Which TabPFN model to load.
         fit_mode: Determines caching behavior.
+        auto_version_override: When set, overrides the default model version for
```
priorphil: Can you expand that comment? What is should this be set to and what is the default case for "auto" routing?

src/tabpfn/constants.py
```diff
@@ -34,8 +34,12 @@ class ModelVersion(str, Enum):
     V2 = "v2"
     V2_5 = "v2.5"
     V2_6 = "v2.6"
+    V3 = "v3"
 
 
+MODEL_ROUTING_V3_MAX_FEATURES = 250  # TODO: Finalize this value
```
priorphil: Cam you add a comment what each does? I.e., which one takes priority?

priorphil (approved):

src/tabpfn/base.py
klemens-floege: I think for today lets maybe not activate routing by default

klemens-floege: I think for today lets maybe not activate routing by default

klemens-floege (approved):
LGTM!

bejaeger:
Thanks for the review, will disable routing by using a large `MODEL_ROUTING_V3_MAX_FEATURES` for now.

src/tabpfn/constants.py
bejaeger: removed routing entirely

PR #911, opened 2026-05-05
Move fitted cache to device to fix multi-gpu kv cache

bejaeger:


src/tabpfn/preprocessing/torch/pipeline_interface.py
```diff
@@ -262,6 +266,12 @@ def _validate_metadata(
             )
 
 
+def _move_cache_to_device(
+    cache: dict[str, torch.Tensor], device: torch.device
```
priorphil, line 275: Are we converting the TabPFNV3Cache class to a dict? Sorry, missing some context here

priorphil (approved):

src/tabpfn/preprocessing/torch/pipeline_interface.py
bejaeger: Nope, this is the cache for the GPU preprocessors (which we store as dict), so very small. I changed the PR a bit and keep all of these caches on CPU and move them to the GPU on-the-fly.

PR #912, opened 2026-05-05
fix: add TABPFN_TOKEN to secrets

ggprior:


priorphil (approved):
Thanks!

src/tabpfn/model_loading.py
```diff
@@ -384,12 +384,14 @@ def download_all_models(to: Path) -> None:
...
                 )
                 continue
-            download_model(
+            result = download_model(
                 to=path,
                 version=model_version,
                 which=cast("Literal['classifier', 'regressor']", model_type),
                 model_name=ckpt_name,
             )
+            if result != "ok":
```
oscarkey, line 393: sorry for the driveby comment, but should this also raise an error so this stage fails in the ci and it's more obvious where the problem is?

brendan-priorlabs, line 393: This is frustrating, but I'm not sure we want this. Previously we had something like this and produced massive stack traces that confused users. Very helpful for debugging, but not a good experience for the average user that forgot to set a token. Curious what others think here. cc: @oscarkey

brendan-priorlabs (changes requested):

src/tabpfn/model_loading.py
priorphil: How about adding a flag "raise_on_error", which we'd default to False but set to True for CI to detect failures more easily?

ggprior: I'd love to raise here but would currently fail due to V3 license plumbing not being in server yet

Image

ggprior: See comment above, we can't raise currently but I can follow-up with this later today once we have the V3 license plumbing in the server

brendan-priorlabs: I like that, @priorphil. Thanks both!

brendan-priorlabs (approved):
LGTM

src/tabpfn/model_loading.py
oscarkey: I'm in favour of always raising an exception, as this function isn't on the standard path so only people who check out the repository and call `download_all_models.py` (as suggested for offline usage in the readme), or go digging for this function, will call it. And they're probably advanced enough to understand the exception. But also happy with the flag :)

PR #921, opened 2026-05-06
Document platform and CPU/GPU install notes

### LINKED ISSUE #158: [docs] Add supported operating systems, CPU vs GPU install
In the installation instructions, it would be nice to detail the supported operating systems and any gotcha's that would complicate usage on any operating systems.

For example, in [AutoGluon's installation documentation](https://auto.gluon.ai/stable/install.html), there are separate tabs for each OS with OS specific details in case there are special requirements.

Additionally, it might be good to have separate install instructions for CPU and GPU machines. For CPU machines, you can get away with installing the CPU PyTorch version, which is a _much_ smaller download (example in AutoGluon docs).

Bortlesboat:
## Summary
- add platform notes for supported Python versions and OS coverage
- clarify CPU-only vs GPU install expectations
- point users to the official PyTorch selector for CPU/CUDA wheel selection

Fixes #158.

## Verification
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,13 @@ cd TabPFN
 uv sync
 ```
 
+#### Platform and hardware notes
+
+- **Python**: TabPFN supports Python 3.9 through 3.13.
+- **Operating systems**: Linux, macOS, and Windows are covered by the test matrix. GPU usage depends on your local PyTorch, CUDA, and driver setup.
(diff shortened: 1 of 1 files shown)


oscarkey:
hey, thanks for your contribution! I'm afraid we only just noticed this PR as it never exited the draft state. There's now a second PR convering the same content (https://github.com/PriorLabs/TabPFN/pull/1077), so I will close this one in favour of that one.

PR #924, opened 2026-05-07
Introduce embedding_dim API and rename ninp to emsize

bejaeger:
## Summary

- Renames `self.ninp` → `self.emsize` across all architecture classes (`PerFeatureTransformer`, `TabPFNV2p5`, `TabPFNV2p6`, `TabPFNV3`) for consistency with config field names
- Adds `embedding_dim` property to the `Architecture` base interface, backed by `self.emsize`
- `TabPFNV3` overrides `embedding_dim` to return `self.icl_emsize` (the ICL output embedding size, which is `embed_dim * feat_agg_num_cls_tokens`) since its output embedding dimension differs from the base `embed_dim`
- Updates tests and docstrings to use the new API

## Test plan

- [x] `tests/test_inference.py` — `_TestModel.emsize` and `_TestModelLegacy.emsize` renamed from `ninp`
- [x] `tests/test_regressor_interface.py` — `test_get_embeddings` uses `model_instance.embedding_dim`
- [x] `tests/test_classifier_interface.py` — `test_get_embeddings` uses `model.models_[0].embedding_dim`
- [x] All pre-commit hooks (mypy, ruff) pass

### DIFF
--- a/src/tabpfn/architectures/interface.py
+++ b/src/tabpfn/architectures/interface.py
@@ -249,9 +249,9 @@ class Architecture(nn.Module, ABC):
         )
 
     @property
+    @abstractmethod
     def embedding_dim(self) -> int:
         """Returns the dimension of the embeddings.
 
         This is the size of the embeddings that can be used for further processing.
         """
-        return self.emsize
--- a/src/tabpfn/architectures/interface.py
+++ b/src/tabpfn/architectures/interface.py
@@ -251,7 +251,9 @@ class Architecture(nn.Module, ABC):
     @property
     @abstractmethod
     def embedding_dim(self) -> int:
-        """Returns the dimension of the embeddings.
+        """The row-level embedding dimension produced by this architecture.
 
-        This is the size of the embeddings that can be used for further processing.
+        This is the size of each per-row latent representation, which can be
+        used for downstream tasks such as clustering, feature analysis, search,
+        or meta-learning.
         """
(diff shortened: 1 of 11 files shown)


src/tabpfn/architectures/interface.py
```diff
@@ -247,3 +247,11 @@ def get_default_performance_options(self) -> PerformanceOptions:
             force_recompute_layer=False,
             use_chunkwise_inference=False,
         )
+
+    @property
+    @abstractmethod
+    def embedding_dim(self) -> int:
+        """Returns the dimension of the embeddings.
+
+        This is the size of the embeddings that can be used for further processing.
```
priorphil: What is this actually used for? Tbh, I'm not really sure we should put this on the Architecture as we can have different embedding sizes within an architecture and their meaning differs (i.e., is this per cell, per row, per block..,?).

priorphil (commented):
See comment - what do we use it for? Can we instead just remove it completely?

src/tabpfn/architectures/interface.py
bejaeger: This is the size of the embeddings per row that can be used for downstream tasks (accessible with `wrapper.get_embeddings()`). I think exactly because of what you mention, that the model can have different "embeddings", this is useful to expose. It's not strictly needed as of right now (because it's also not hooked up with the model wrapper), but the user could still get the expected size of the embeddings using `classifier.model_.embedding_dim`.

I clarified in the the docstring what this is. WDYT?

priorphil (approved):

PR #928, opened 2026-05-07
Stratified row subsampling

bejaeger:
When subsampling rows, use stratified subsampling for classification per default; use round-robin subsampling for regression.


src/tabpfn/preprocessing/ensemble.py
```diff
@@ -346,11 +347,102 @@ def _subsample_rows_balanced(
...
+
+    Returns:
+        List of row-index arrays (one per estimator), or ``None`` entries when no
+        subsampling is needed.
+    """
+    n_rows = len(y)
+    if subsample_size >= n_rows:
+        return None
+
+    classes, inverse = np.unique(y, return_inverse=True)
```
LeoGrin, line 423: nit: I think y here is a torch tensor? and it can be on gpu/mps if differentiable_input=True, in which case it would fail?

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -346,11 +347,102 @@ def _subsample_rows_balanced(
...
+    n_classes = len(classes)
+    class_sizes = np.bincount(inverse, minlength=n_classes)
+
+    class_indices: list[np.ndarray] = [
+        np.where(inverse == c)[0] for c in range(n_classes)
+    ]
+
+    target_counts = _compute_stratified_class_counts(
```
LeoGrin, line 431: nit:

> src/tabpfn/preprocessing/ensemble.py:416 computes one fixed class quota and src/tabpfn/preprocessing/ensemble.py:428 skips classes whose quota rounds to zero. For imbalanced data like class sizes [999, 1] with subsample_size=100, the quota is [100, 0], so the minority class is excluded from every estimator. That can make a classifier fit without ever seeing one of classes_. The quota needs to rotate/stochastically
    allocate fractional rare-class mass across estimators, or guarantee coverage when feasible`

I think that's true?

LeoGrin (approved):

src/tabpfn/preprocessing/ensemble.py
bejaeger: Yep! am adding it rn.

bejaeger: Done!

bejaeger:
Thanks for the quick review! Every class is included in every estimator now.

src/tabpfn/preprocessing/ensemble.py
bejaeger: I removed passing y at all to the TabPFNPreprocessor (for diff input) to keep the interface clean. If someone uses differentiable_input=True they likely don't want to subsample anyway?

PR #929, opened 2026-05-07
Add tests that classifier/regressor pass an accuracy threshold.

oscarkey:
Just a quick smoke test to try and prevent regressions like https://linear.app/priorlabs/issue/PRI-277/v26-has-low-accuracy-on-mps

We'll need a different strategy to detect more subtle regressions (see https://linear.app/priorlabs/issue/RES-1530/tabpfn-continuous-benchmarking)

Fixes RES-1528

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,25 +31,11 @@ def set_global_seed() -> None:
     random.seed(seed)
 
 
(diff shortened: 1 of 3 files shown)


eliott-kalfon (approved):
Looks good to me!

PR #933, opened 2026-05-08
Free unused KV heads from ICL cache backing storage

LeoGrin:
## Summary
In `ICLAttention.forward`, when `num_kv_heads_test` is set, the cached K/V is a slice (view) of the full-projection tensor. `.detach()` does not copy, so the cache silently retains storage for all KV heads even though only the first head(s) are used for test rows. Adding `.contiguous()` materialises just the kept slice and lets the rest be freed.

Effect: KV cache GPU memory at `fit_with_cache` shrinks to the actual MQA size (no functional change).

## Test plan
- [ ] Existing unit tests (`tests/test_consistency.py`, `tests/test_inference.py`) still pass
- [ ] Verify cache `key`/`value` storage size matches the kept-head shape

### DIFF
--- a/changelog/933.fixed.md
+++ b/changelog/933.fixed.md
@@ -1,0 +1,1 @@
+Reduce KV cache GPU memory in `fit_with_cache` by materialising only the kept KV head(s) at cache-build time. Output is unchanged.
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -926,8 +926,11 @@ class ICLAttention(nn.Module):
             k_cache, v_cache = k, v
             if self.num_kv_heads_test is not None:
                 nh_test_heads = self.num_kv_heads_test
-                k_cache = k_cache[:, :, :nh_test_heads]
-                v_cache = v_cache[:, :, :nh_test_heads]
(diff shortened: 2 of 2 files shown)


jmetzen-pl (approved):
Looks good - but the Gemini comment sounds reasonable (would check before merging).

LeoGrin:
Thanks!

 Agree with Claude here on the gemini comment:
  
  ▎ Re .detach() before .contiguous(): the cache-build path runs inside torch.inference_mode() (set up in InferenceEngineExplicitKVCache._build_cache via with ... torch.inference_mode():), so neither k nor v requires grad and .contiguous() doesn't create autograd nodes regardless of order. Happy to reorder if you'd prefer it for explicitness, but there's no measurable difference in practice.                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                    
  ▎ Re train_embeddings view retention: the cache-build path passes only train rows to the model — see InferenceEngineExplicitKVCache._build_cache, which calls model(X, ...) with X = X_train.unsqueeze(1) (no test rows concatenated). So at this site x_BRiD has shape (B, num_train, D), and x_BRiD[:, :num_train] covers the entire tensor — no view-retention waste. It's worth a defensive .contiguous() for symmetry if forward is ever called with mixed train+test rows and return_kv_cache=True, but that's not a reachable path today.

PR #935, opened 2026-05-08
Add FlashAttention-3 backend selector to v3

jmetzen-pl:
## Summary

Adds an opt-in FlashAttention-3 attention path on v3, dispatched via a new `PerformanceOptions.attention_backend` field (`"auto"` | `"sdpa"` | `"fa3"`, default `"auto"`). The dispatcher routes to FA3 only when the call is eligible (Hopper sm90+, fp16/bf16, `head_dim ∈ {64, 96, 128, 192, 256}`, `flash_attn_interface` importable); otherwise it falls back to PyTorch SDPA unchanged.

In `"auto"` mode the dispatcher additionally requires `max(seq_q, seq_kv) >= 10_000` — below that, FA3's per-call dispatch overhead exceeds its kernel-throughput win and SDPA is faster end-to-end. `"fa3"` (forced) bypasses the seqlen threshold; capability gates still apply and a clear error is raised when ineligible. `"sdpa"` always uses SDPA.

The backend selector is threaded explicitly through the ICL stage only — non-ICL call sites (`dist_embedder`, decoder, cross-attention) have FA3-ineligible head_dims and would always fall back to SDPA anyway.

FA3 is shipped only when Hopper kernels are present, since they are sm_90-specific (WGMMA, TMA). Blackwell will need FlashAttention-4 when upstream lands it.

## Implementation

- `src/tabpfn/architectures/shared/fa3_backend.py`: lazy FA3 import + capability/eligibility/preference gates.
- `tabpfn_v3._batched_scaled_dot_product_attention`: `attention_backend` kwarg threaded through `ICLAttention` → `ICLTransformerBlock` → the `icl_blocks` loop.
- `tests/test_architectures/test_attention_backends.py`: dispatch-logic tests (eligibility, fall-back, error path) run anywhere; numerical-equivalence tests gated by `@pytest.mark.hopper` and skip on non-Hopper hosts.
- `src/tabpfn/architectures/shared/fa3_setup.md`: build-from-source instructions (in-place install on Hopper, plus a cross-compile recipe for CPU-only build nodes).

## Public API changes

- New optional field `PerformanceOptions.attention_backend`. Default `"auto"` preserves existing SDPA-only behaviour everywhere FA3 isn't eligible.

## Test plan

- [x] `pytest tests/test_architectures/test_attention_backends.py tests/test_architectures/test_tabpfn_v3.py` — 17 passed locally on CPU (8 hopper-marked tests skip)
- [x] ruff + mypy + pre-commit hooks pass
- [x] Numerical equivalence on H100 (`atol=rtol=5e-3` over MHA / MQA / GQA shapes) — verified manually on internal H100 against an earlier revision of this code
- [x] CI

### DIFF
--- a/src/tabpfn/architectures/shared/fa3_setup.md
+++ b/src/tabpfn/architectures/shared/fa3_setup.md
@@ -1,0 +1,135 @@
+# FlashAttention-3 (Hopper) backend
+
+TabPFN v3 can dispatch attention to FlashAttention-3 instead of PyTorch's
+SDPA. FA3 is expected to be meaningfully faster on Hopper-class GPUs
+(H100, H200): an externally contributed micro-benchmark on v2.5 attention
+shapes reported ~1.5–1.7× over SDPA (we plan to re-run this internally
+before citing in the v3 model report). v3 numbers will land alongside the
+v3 report itself.
+
+This is **opt-in**. Default is `attention_backend="auto"`, which uses FA3
+when eligible and SDPA otherwise.
+
+## When FA3 is used
+
+The dispatcher routes a call to FA3 only when **all** of the following hold:
+
+- The `flash_attn_interface` Python package is importable.
+- The attention is on a CUDA tensor whose device is Hopper-class
+  (compute capability 9.0+).
+- The dtype is `torch.float16` or `torch.bfloat16`.
+- The head dimension is one of `{64, 96, 128, 192, 256}`.
+
+Calls that don't satisfy these (e.g. v3's `dist_embedder` with head_dim=16)
+fall back to SDPA in `auto` mode, or raise in `fa3` mode.
+
+### `auto` also applies a sequence-length threshold
+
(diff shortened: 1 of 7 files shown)


src/tabpfn/architectures/shared/fa3_setup.md
```diff
@@ -0,0 +1,135 @@
+# FlashAttention-3 (Hopper) backend
+
+TabPFN v3 can dispatch attention to FlashAttention-3 instead of PyTorch's
+SDPA. FA3 is expected to be meaningfully faster on Hopper-class GPUs
+(H100, H200): an externally contributed micro-benchmark on v2.5 attention
+shapes reported ~1.5–1.7× over SDPA (we plan to re-run this internally
+before citing in the v3 model report). v3 numbers will land alongside the
+v3 report itself.
```
jmetzen-pl: change the paragraph above

priorphil (approved):
Nice, thanks!

PR #937, opened 2026-05-08
Scale n_estimators to ensure feature coverage

bejaeger:


bejaeger:

- #937 ◀

src/tabpfn/preprocessing/ensemble.py
```diff
@@ -1171,3 +1172,38 @@ def _resolve_feature_subsampling_method(
...
+    """
+    if not preprocessor_configs:
+        return n_estimators
+    min_max_features = min(c.max_features_per_estimator for c in preprocessor_configs)
+    if min_max_features <= 0:
+        return n_estimators
+    min_required = math.ceil(n_total_features / min_max_features)
+    if n_estimators >= min_required:
+        return n_estimators
+    warnings.warn(
```
klemens-floege, line 1200: nit: Maybe we can also distinguish here the cases between a user setting n_estimators and overwriting that, or the user did not set  it and then we more of an FYI

klemens-floege (approved):
Thanks for adding! Just to sanity check: This should not be triggered on anything in TabArena right, since we have a max of 1.6k feats. With the Feature Importance sampling they do not all get different ones so with 8 estimators you do get a different prediction right?

src/tabpfn/preprocessing/ensemble.py
bejaeger: Hm...we cannot really detect whether the 8 estimators are user provided or the default. The warning might be a bit noisy but I would think it's okay for now? IF a user sees it, they can adjust the n_estimators themselves and silence it? I would say it's fine for now :)

bejaeger:
Thanks for the quick review!

src/tabpfn/preprocessing/ensemble.py
klemens-floege: Yeah sounds good!

PR #938, opened 2026-05-09
Make CPU squashing scaler faster & introduce torch squashing scaler

bejaeger:


LeoGrin (approved):
LGTM thanks!

I think we can probably subsample to speed up np.nanpercentile but we can leave that for later.

PR #941, opened 2026-05-10
Fix torch svd on multi-gpu setup

bejaeger:
Brings down preprocessing seconds to roughly <10-15% of total runtime again for 100k rows.

### TODO
- [x] run benchmark to check for identical performance


LeoGrin (approved):
LGTM!!!

PR #944, opened 2026-05-10
Activate torch squashing scaler

bejaeger:
Brings down preprocessing overhead to <3%!

## TODO
- [x] benchmarks show same results with active squashing scaler


LeoGrin (approved):
LGTM!

PR #949, opened 2026-05-11
Pull SDPA out into shared + Add support for FlashAttn on MPS.

priorphil:
On MPS, we never got flash attention because pytorch SDPA only triggers the efficient path for very specific shapes.
I've tried a couple of implementations, the safest bet seems to be MLX for our use case.
Without this, a dataset with 36000 train rows + 9000 test rows OOMs on my 24 GB machine, with this, everything runs smoothly.


eliott-kalfon:
Hi @priorphil , rerunning the script `memory_v3.py`, I get this error:

This is issue is usually caused by one of the following two reasons:

```
1) Large test set — split into batches:

    predictions = []
    for i in range(0, len(X_test), 100):
        pred = model.predict_proba(X_test[i:i + 100])
        predictions.append(pred)
    predictions = np.vstack(predictions)

2) Large training set — batching won't help.
   You need subsampling or ensembling, see:
   https://github.com/PriorLabs/tabpfn-extensions/blob/main/examples/large_datasets/large_datasets_example.py

Your sizes: 36,168 train / 9,043 test samples, 16 features.
Not sure which? If model.predict_proba(X_test[:1]) also fails, it's (2).

On MPS, consider installing MLX (pip install mlx). This automatically enables flash attention, which will drastically reduce memory usage for datasets with more than 1500 samples.

Original error: MPS backend out of memory (MPS allocated: 1.24 GiB, other allocations: 24.47 GiB, max allowed: 26.21 GiB). Tried to allocate 5.39 GiB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

This is because I do not have mlx installed. We may want to include `mlx` in our dependencies (not only in `dev`).

After installing `mlx`, I also get these warnings:

```
/Users/ekal/TabPFN/src/tabpfn/preprocessing/torch/steps.py:403: UserWarning: SVD operators ('aten::linalg_svd', 'aten::linalg_qr') are not currently supported on the MPS backend and will fall back to run on the CPU. This may have performance implications.
  svd_cache = self._svd.fit(x_scaled)
```

With inference being slow

priorphil:
Hm, I'm always careful with installing optional dependencies, didn't include them here because MLX only gets activated for larger datasets. Though at least MLX seems to be tiny (<1 MB), unlike torch, so maybe fine :) Let me ask in the channel how others feel.

Also @bejaeger: Should we remove this pre-processing warning? From running the example above (~40k rows), the main bulk is still spent in the model call when using MPS. I think moving the pre-processing to GPU has a much larger effect on H100 machines, where the GPU compute is much more powerful compared to CPU.

priorphil:
Made them default now, I guess the out of the box experience is more important :)

tests/test_architectures/test_mlx_backend.py
```diff
@@ -0,0 +1,321 @@
...
+        flash_attention_mlx(q, q, q)
+
+
+# ---------------------------------------------------------------------------
+# flash_attention_mlx — numerical equivalence
+# ---------------------------------------------------------------------------
+
+
+@_skip_unless_mlx_and_mps
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32])
```
jmetzen-pl, line 181: The FP32 cases fail for me (observed errors are ~2e-3). Can it be that MLX's fused SDPA kernel runs in reduced precision anyway?

src/tabpfn/architectures/shared/mlx_backend.py
```diff
@@ -0,0 +1,88 @@
...
+        and v.dtype in (torch.float16, torch.bfloat16, torch.float32)
+        and head_dim <= 128
+    )
+
+
+def is_mlx_preferred(
+    q_BHSD: torch.Tensor, k_BJSD: torch.Tensor, v_BJSD: torch.Tensor
+) -> bool:
+    """True iff MLX is preferred for this attention call."""
+    return is_eligible_for_mlx(q_BHSD, k_BJSD, v_BJSD) and k_BJSD.shape[2] >= 1536
```
jmetzen-pl: where is 1536 coming from?

src/tabpfn/errors.py
```diff
@@ -125,6 +125,12 @@ def __init__(
             f"Not sure which? If model.{predict_method}(X_test[:1]) "
             f"also fails, it's (2)."
         )
+        if self.device_name == "MPS":
+            message += (
+                "\n\nOn MPS, consider installing MLX (pip install mlx). This "
```
jmetzen-pl: Do we need this message when MLX is now a runtime dependency which should be installed anyway?

src/tabpfn/architectures/shared/mlx_backend.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _torch_to_mlx(torch_tensor: torch.Tensor) -> mx.array:
+    """Convert a PyTorch tensor to an MLX array, handling bfloat16 via int16."""
+    if torch_tensor.dtype == torch.bfloat16:
+        return mx.array(torch_tensor.view(torch.int16).cpu().numpy()).view(mx.bfloat16)
```
jmetzen-pl, line 51: do we need this GPU->CPU->GPU roundtrip?

src/tabpfn/architectures/shared/mlx_backend.py
```diff
@@ -0,0 +1,88 @@
...
+def _mlx_to_torch(
+    mlx_array: mx.array, device: torch.device, dtype: torch.dtype
+) -> torch.Tensor:
+    if dtype == torch.bfloat16:
+        return (
+            torch.from_numpy(np.array(mlx_array.view(mx.int16)))
+            .view(torch.bfloat16)
+            .to(device)
+        )
+    return torch.from_numpy(np.array(mlx_array)).to(device)
```
jmetzen-pl, line 64: Claude is suggesting https://docs.pytorch.org/docs/2.11/generated/torch.from_dlpack.html here to avoid GPU->CPU->GPU here; have you tried this? (not so familiar with DLPack, so unsure if it would really work, but sharing the proposal)

changelog/949.added.md
```diff
@@ -0,0 +1 @@
+Add flash attention support for MPS to reduce memory usage.
```
jmetzen-pl: might add that the attention_backend is removed again

src/tabpfn/architectures/shared/scaled_dot_product_attention.py
```diff
@@ -0,0 +1,103 @@
...
+    CUDA_MAX_GRID = 65536
+    num_iterations = (num_parallel_calls + CUDA_MAX_GRID - 1) // CUDA_MAX_GRID
+    sub_batch = (q_BHSD.shape[0] + num_iterations - 1) // num_iterations
+
+    with sdpa_kernel(backends=backends):
+        outputs = []
+        for i in range(num_iterations):
+            outputs.append(
+                torch.nn.functional.scaled_dot_product_attention(
+                    q_BHSD[i * sub_batch : (i + 1) * sub_batch].contiguous(),
```
jmetzen-pl, line 140: do these have to call contiguous() now?

tests/test_architectures/test_mlx_backend.py
```diff
@@ -0,0 +1,321 @@
...
+
+def test__eligible_false_for_cpu_tensors() -> None:
+    """CPU tensors are never eligible (not MPS)."""
+    q = torch.zeros(1, 2, 8, 64, dtype=torch.float16)
+    assert not is_eligible_for_mlx(q, q, q)
+
+
+def test__eligible_false_for_head_dim_over_128() -> None:
+    """head_dim=129 is rejected regardless of other properties."""
+    q = torch.zeros(1, 2, 8, 129, dtype=torch.float16)
```
jmetzen-pl: we need to move this tensor to mps for the test to make sense, right?

src/tabpfn/architectures/shared/mlx_backend.py
```diff
@@ -0,0 +1,88 @@
...
+
+def flash_attention_mlx(
+    q_BHSD: torch.Tensor, k_BHJD: torch.Tensor, v_BHJD: torch.Tensor
+) -> torch.Tensor:
+    """Scaled dot product attention using MLX on MPS devices, with padding for head_dim.
+
+    Pytorch SDPA has only very limited support for flash attention on MPS, which is
+    incompatible with shapes encountered in our model. To reduce memory usage, we route
+    to MLX during inference.
+    """
```
jmetzen-pl, line 76: Could add a one-line comment why padding does not change results here

src/tabpfn/architectures/shared/mlx_backend.py
priorphil: That was from local benchmarks - MLX only got faster after this threshold. Set it to 128 now, as I ran into OOM errors with 100k rows, which I don't if I lower :)

src/tabpfn/architectures/shared/mlx_backend.py
priorphil: We do the round trip either way, didn't find a reliable way to not do this.

src/tabpfn/architectures/shared/mlx_backend.py
priorphil: I was also looking into this - tbh I'm a bit reluctant to add this, since silent memory corruptions are really hard to detect. Since it looks like there's a pytorch native solution soon, I'll update the README to say that for best performance, upgrading torch is the way to go. WDYT?

src/tabpfn/architectures/shared/scaled_dot_product_attention.py
priorphil: Gemini suggested it's safer since some pytorch backends require it ¯\\\_(ツ)\_/¯

tests/test_architectures/test_mlx_backend.py
priorphil: Urgh, thanks

tests/test_architectures/test_mlx_backend.py
priorphil: Hm weird - which torch version are you running with?
Passes locally for me with
```
mlx                        0.31.2
torch                      2.11.0
```

src/tabpfn/architectures/shared/mlx_backend.py
jmetzen-pl: yes, sounds good

src/tabpfn/architectures/shared/scaled_dot_product_attention.py
jmetzen-pl: ok, is a bit unfortunate, but doing this dependent on the backend here would also be error-prone/complicated

tests/test_architectures/test_mlx_backend.py
jmetzen-pl: Let CC try different mlx versions and it found this:

```
  mlx version: 0.29.3, 0.29.4
  fp32 tests: all 8 pass
  max abs error (fp32): ~8 × 10⁻⁷ (true fp32 precision)
  ────────────────────────────────────────
  mlx version: 0.30.0, 0.30.6, 0.31.0, 0.31.2 (current)
  fp32 tests: 4 fp32 fail
  max abs error (fp32): ~2.6 × 10⁻³ (~3000× worse)

  MLX changed something in mx.fast.scaled_dot_product_attention between 0.29.4 and
  0.30.0 — almost certainly switched the fp32 fast path to use reduced-precision
  internal accumulation (fp16/bf16 register math), which is what flash-attention
  kernels typically do for throughput. fp16/bf16 paths and the GQA/dispatch tests are
  unaffected.
```

However, if that is true, it should also fail for you with 0.31.2?

bejaeger:
RE preprocessing warning: will be removed [here](https://github.com/PriorLabs/TabPFN/pull/957)

tests/test_architectures/test_mlx_backend.py
jmetzen-pl: are you using an M5 chip?

priorphil: Nope, M4

priorphil:
Was finally able to run benchmarks up to 100k; If torch nightly is available this is a huge win.

```
MLX
train  test cols  total_per_call_seconds
1000   500  100  1.94
10000  500  100  23.61
100000 500  100  414.71

TORCH NIGHTLY
train  test cols total_per_call_seconds
1000   500  100  0.90  
10000  500  100  8.54  
100000 500  100  288.29
```

tests/test_architectures/test_mlx_backend.py
priorphil: Which torch version are you running? Maybe torch changed the default as well?

jmetzen-pl: No, I am on the same torch version. Could you run this?
```
python -c "
import mlx.core as mx, torch, numpy as np, platform, subprocess
chip = subprocess.run(['sysctl', '-n', 'machdep.cpu.brand_string'],
capture_output=True, text=True).stdout.strip()
torch.manual_seed(0)
a = torch.randn(256, 256, dtype=torch.float32)
b = torch.randn(256, 256, dtype=torch.float32)
ref = a @ b
out = np.array(mx.array(a.numpy()) @ mx.array(b.numpy()))
diff = (torch.from_numpy(out) - ref).abs()
print(f'chip: {chip}')
print(f'macOS: {platform.mac_ver()[0]}')
print(f'torch version: {torch.__version__}')
print(f'mlx version: {mx.__version__}')
print(f'device: {mx.default_device()}')
print(f'max abs diff vs torch CPU: {diff.max().item():.3e}')
print(f'mean abs diff: {diff.mean().item():.3e}')
"
```

This gives for me:
  chip: Apple M5
  macOS: 26.3.1
  torch version: 2.11.0
  mlx version: 0.31.2
  device: Device(gpu, 0)
  max abs diff vs torch CPU: 5.196e-02
  mean abs diff: 9.812e-03

priorphil: Ran two different torch versions:
```
chip: Apple M4
macOS: 26.3.1
torch version: 2.13.0.dev20260512
mlx version: 0.31.2
device: Device(gpu, 0)
max abs diff vs torch CPU: 3.815e-05
mean abs diff: 2.968e-06

chip: Apple M4
macOS: 26.3.1
torch version: 2.11.0
mlx version: 0.31.2
device: Device(gpu, 0)
max abs diff vs torch CPU: 3.815e-05
mean abs diff: 2.968e-06
```
So it does seem to be the chip :O

jmetzen-pl: I guess I need to downgrade to an M4 then ;-)
What do we do with the test? Just skip the float32 variants?

priorphil: I guess that doesn't solve the underlying issue - we need some way of detecting this on the users machine. Wonder how bad it would be to run a precision check on the first model call....

jmetzen-pl: but is there anything we can do about it? seems like the float32 results just differ between mlx and torch.cpu on an M5. We could display a warning in that particular case.

README.md
```diff
@@ -23,6 +23,11 @@
 pip install tabpfn
 ```
 
+Note: For best performance on Apple Silicon/MPS, consider installing a pytorch version after
+the nightly "2.13.0.dev20260510". This enables flash attention without relying on MLX
+(the latter requires a GPU-CPU-GPU roundtrip).
```
jmetzen-pl, line 28: we would still keep mlx as a dependency, even though users with a recent nightly don't strictly need it, right?

tests/test_architectures/test_mlx_backend.py
jmetzen-pl: Test pass now for me with the added `os.environ['MLX_ENABLE_TF32']='0'`

README.md
priorphil: Yep - I didn't find a nice way to do this. We could make it optional again but since this requires nightly pytorch I think we should keep it non-optional for now. Once the official pytorch version is released, we could think about enforcing the recent pytorch version for macOS but keep a looser bound for other platforms.

jmetzen-pl (approved):
Also re-run the flash-attention 3 tests on Hopper and they passed.

PR #951, opened 2026-05-12
Revise TabPFN Extensions description in README

noahho:
Updated the description of the TabPFN Extensions repository to clarify its purpose and features.

## Issue
Please link the corresponding GitHub issue. If an issue does not already exist,
please open one to describe the bug or feature request before creating a pull request.

This allows us to discuss the proposal and helps avoid unnecessary work.

## Motivation and Context

---

## Public API Changes

-   [ ] No Public API changes
-   [ ] Yes, Public API changes (Details below)

---

## How Has This Been Tested?

---

## Checklist

-   [ ] The changes have been tested locally.
-   [ ] Documentation has been updated (if the public API or usage changes).
-   [ ] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.
-   [ ] The code follows the project's style guidelines.
-   [ ] I have considered the impact of these changes on the public API.

---

PR #945, opened 2026-05-11
docs: clean up references to removed/deprecated tabpfn-extensions

adrian-prior:
## Summary

Align README and inline references with the tabpfn-extensions cleanup. The \`rf_pfn\` module has been removed, \`post_hoc_ensembles\` and \`hpo\` are deprecated and scheduled for removal, and the old \`examples/large_datasets\` example is gone. Several spots in this repo still pointed at those things.

Companion to the docs cleanup in [priorlabs-docs#52](https://github.com/PriorLabs/docs/pull/52).

### Changes

**README.md**
- Drop \`rf_pfn\`, \`hpo\`, and \`post_hoc_ensembles\` from the Extensions extras list — only the supported extras (\`interpretability\`, \`unsupervised\`, \`embeddings\`, \`many_class\`) remain.
- Remove the broken "Large datasets guide" link from Usage Tips; point readers at the [Models page](https://docs.priorlabs.ai/models) for per-checkpoint size limits.
- Drop the "TabPFN Workflow at a Glance" mermaid diagram. It was wired to the removed/deprecated extensions (\`subsample\`/large-datasets, \`hpo\`, \`post_hoc\`) and was already drifting from current guidance; remove it rather than maintain a stale picture.
- Drop the AutoTabPFNClassifier recommendation from the "How can I improve performance" FAQ and point readers at the Improving Performance docs.
- Drop the AutoTabPFNClassifier parenthetical from the checkpoint FAQ.

**src/tabpfn/errors.py**
- The OOM error message pointed users at the deleted \`examples/large_datasets/large_datasets_example.py\` file in tabpfn-extensions. Replace with a short subsample-or-use-a-bigger-checkpoint hint that links to the Models page.

**examples/notebooks/TabPFN_Demo_Local.ipynb**
- Remove the \`AutoTabPFNClassifier\` import from the necessary-imports cell. The symbol was imported but never used elsewhere in the notebook, so the rest of the notebook is unchanged.

## Test plan

- [x] Render the README locally and confirm no broken anchors or stray references.
- [x] \`grep -rn "rf_pfn\|AutoTabPFN\|TunedTabPFN\|RandomForestTabPFN\|/examples/large_datasets\|tabpfn-extensions\[hpo\]\|tabpfn-extensions\[post_hoc_ensembles\]\|tabpfn-extensions\[rf_pfn\]"\` returns no hits.
- [x] Notebook still imports cleanly (run the imports cell on Colab).
[...]

PR #952, opened 2026-05-12
docs(readme): keep only the pip install path

adrian-prior:
## Summary

Per Noah's request: the README's Installation section listed three paths (pip, install-from-source, local-dev with `uv`). Most readers just want to start using TabPFN — the source-install and dev-install variants belong in CONTRIBUTING-style docs, not at the top of the README. Trim to just `pip install tabpfn`.

Supersedes the equivalent change in #951 (which also bundles other v3 / extras-list edits that overlap with #945).

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,19 +24,8 @@
 > No GPU? Use our free hosted inference via [TabPFN Client](https://github.com/PriorLabs/tabpfn-client).
 
 ### Installation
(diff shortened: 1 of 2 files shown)


adrian-prior:
Folded into #945 directly to keep all README changes in one PR. Closing here; please review on #945.

PR #983, opened 2026-05-28
Quantize KV Cache of TabPFN-3 run with fit_mode="fit_with_cache"

bejaeger:


bejaeger:

- #983 ◀

src/tabpfn/architectures/kv_cache.py
```diff
@@ -5,16 +5,51 @@
...
+import torch
+from torch import Tensor
+
+# High, low, max value for each dtype
+_QUANTIZATION_RANGES: dict[torch.dtype, tuple[int, int, int]] = {
+    torch.int8: (-128, 127, 127),
+}
+
+
+def _quantize_tensor(
```
priorphil, line 30: Wow, so we don't even need blockwise quant?

tests/test_classifier_interface.py
```diff
@@ -562,6 +578,37 @@ def test__fit_preprocessors_and_with_cache_produce_equal_results(
     np.testing.assert_array_equal(preds, tabpfn.predict(X))
 
 
+# Disable MPS as it doesn't support float64.
```
priorphil: Can we just use float32 instead so these also run on MPS?

tests/test_regressor_interface.py
```diff
@@ -315,6 +331,35 @@ def test__fit_preprocessors_and_with_cache_produce_equal_results(
     np.testing.assert_array_almost_equal(preds, tabpfn.predict(X), decimal=2)
 
 
+# Disable MPS as it doesn't support float64.
```
priorphil: Same here, can we use float32?

priorphil (approved):
Nice!

priorphil (commented):
Just to double check, there's no native pytorch quantized tensor that would de-quantize on the fly?

PR #984, opened 2026-05-28
test(nightly): enforce latest tabpfn-extensions in notebook install check

adrian-prior:
## Summary
- Extend the nightly notebook install check so that, in addition to `tabpfn`, any notebook that installs `tabpfn-extensions` also pins it to PyPI latest. This addresses [PRI-272](https://linear.app/priorlabs/issue/PRI-272/improve-tabpfn-extensions-test-launched-from-tabpfn) — the previous check only verified `tabpfn`, which is why an outdated `tabpfn-extensions` in `TabPFN_Demo_Local.ipynb` could regress without catching (cf. PRI-268 / PRI-269).
- Refactor the PyPI version lookup into a generic `_latest_pypi_version(package)` (cached per-package) and add `_installed_version(package, env)` returning `None` when the package isn't in the venv, so the check can stay silent for notebooks that don't install extensions.

## Test plan
- [x] Local positive run: `RUN_NOTEBOOK_INSTALL_CHECK=1 pytest --noconftest tests/test_notebook_installs.py -v` — both notebooks pass; `TabPFN_Demo_Local.ipynb` now also verifies `tabpfn-extensions`.
- [x] Local negative-control run with `_latest_pypi_version` monkey-patched to return a fake version: both notebooks fail on `tabpfn`, and `TabPFN_Demo_Local.ipynb` additionally fails on `tabpfn-extensions=={installed} (PyPI latest is …)`, confirming the new optional check actually fires.
- [ ] Trigger the nightly workflow via `workflow_dispatch` after merge to confirm CI behavior on all matrix Python versions.

### DIFF
--- a/changelog/984.changed.md
+++ b/changelog/984.changed.md
@@ -1,0 +1,1 @@
+Nightly notebook install check now also asserts `tabpfn-extensions` resolves to PyPI latest for every example notebook that installs it, catching the case where a stale upper pin or transitive constraint would silently install an outdated `tabpfn-extensions`.
--- a/changelog/984.changed.md
+++ b/changelog/984.changed.md
@@ -1,1 +1,0 @@
-Nightly notebook install check now also asserts `tabpfn-extensions` resolves to PyPI latest for every example notebook that installs it, catching the case where a stale upper pin or transitive constraint would silently install an outdated `tabpfn-extensions`.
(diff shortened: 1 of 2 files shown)


changelog/984.changed.md
```diff
@@ -0,0 +1 @@
+Nightly notebook install check now also asserts `tabpfn-extensions` resolves to PyPI latest for every example notebook that installs it, catching the case where a stale upper pin or transitive constraint would silently install an outdated `tabpfn-extensions`.
```
LeoGrin: don't think we need a changelog here as it's not relevant to external users

LeoGrin (approved):
LGTM!

PR #985, opened 2026-05-28
Ensure .devices_ attribute is always defined.

oscarkey:
Before it was undefined until `.fit()`. Now it's `None` until then.

Fixes PRI-175


oscarkey:
it turns out that sklearn requires no attribute is set during __init__ except the args to the constructor. So our current behaviour is in fact the correct one.

PR #987, opened 2026-05-29
Add a visualisation utility to plot the predicted distribution (regression)

eliott-kalfon:
## Issue
The goal of this PR is to enable users to visualise the predicted distribution for a given observation in a regression setting.

To the review, some design choices:

- No change to the `TabPFNRegressor` API
- Creation of a `tabpfn.visualization` module
- matplotlib was not added to core dependencies but to the `viz` subproject

The objective is to give the user this functionality without adding clutter or dependencies.  

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
Local run and tests
---

## Checklist

---

### DIFF
--- a/examples/plot_regression_distribution.py
+++ b/examples/plot_regression_distribution.py
@@ -14,9 +14,7 @@ from tabpfn import TabPFNRegressor
 from tabpfn.visualisation import plot_regression_distribution
 
 X, y = load_diabetes(return_X_y=True)
-X_train, X_test, y_train, y_test = train_test_split(
-    X, y, test_size=0.2, random_state=0
-)
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=0)
 
 reg = TabPFNRegressor(n_estimators=4)
 reg.fit(X_train, y_train)
--- a/examples/plot_regression_distribution.py
+++ b/examples/plot_regression_distribution.py
@@ -14,7 +14,9 @@ from tabpfn import TabPFNRegressor
 from tabpfn.visualisation import plot_regression_distribution
 
 
-def main(output_path: str = "regression_distribution.png", *, show: bool = True) -> None:
+def main(
+    output_path: str = "regression_distribution.png", *, show: bool = True
+) -> None:
     """Fit a regressor and plot the predicted distribution for three test points."""
     X, y = load_diabetes(return_X_y=True)
     X_train, X_test, y_train, y_test = train_test_split(
@@ -44,7 +46,9 @@ def main(output_path: str = "regression_distribution.png", *, show: bool = True)
             true_val, color="purple", ls="-.", lw=1.4, label=f"true = {true_val:.0f}"
         )
         leg = ax.get_legend()
-        handles = getattr(leg, "legend_handles", None) or getattr(leg, "legendHandles", [])
+        handles = getattr(leg, "legend_handles", None) or getattr(
+            leg, "legendHandles", []
+        )
         ax.legend(handles=[*handles, true_line], fontsize=9)
         ax.set_title(title)
 
--- a/examples/plot_regression_distribution.py
+++ b/examples/plot_regression_distribution.py
@@ -13,35 +13,47 @@ from sklearn.model_selection import train_test_split
 from tabpfn import TabPFNRegressor
 from tabpfn.visualisation import plot_regression_distribution
 
-X, y = load_diabetes(return_X_y=True)
-X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=0)
 
-reg = TabPFNRegressor(n_estimators=4)
-reg.fit(X_train, y_train)
-
-# Pick three test points with different predicted values for comparison.
-preds = reg.predict(X_test)
-order = np.argsort(preds)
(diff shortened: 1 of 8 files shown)


src/tabpfn/visualisation/regression_distribution.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    # Local import because matplotlib is an optional dependency.
+    try:
+        import matplotlib.pyplot as plt  # noqa: PLC0415
+        from matplotlib.patches import Patch  # noqa: PLC0415
+    except ModuleNotFoundError as err:
+        raise ModuleNotFoundError(
+            "matplotlib is required for plotting. "
+            'Install it with `pip install "tabpfn[viz]"`'
+        ) from err
```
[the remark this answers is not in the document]
eliott-kalfon: Not sure about this, as it may bloat the codebase. I am leaving the comment open, to be considered by the reviewer. Thank you!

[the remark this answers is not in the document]
bejaeger: I think these checks would be fair to add, but not essential. Also serve as documentation.

src/tabpfn/visualization/regression_distribution.py
```diff
@@ -0,0 +1,130 @@
...
+    if arr.ndim == 1:
+        arr = arr.reshape(1, -1)
+    if arr.shape[0] != 1:
+        raise ValueError(f"Expected a single row, got shape {arr.shape}.")
+    return arr
+
+
+def plot_regression_distribution(
+    regressor: TabPFNRegressor,
+    x: object,
```
bejaeger: Not sure about the interface of this one. Let's say the user does not only want to visualize it but also want's to do something else with the output of the prediction.
What about not passing the full regressor and `x` and predict inside this plotting function, and rather just passing the output of `regressor.predict(..., output_type="full")`?
Might need a small input validation as well.

examples/plot_regression_distribution.py
```diff
@@ -0,0 +1,49 @@
...
+mid_idx = int(np.argsort(preds)[len(preds) // 2])
+high_idx = int(np.argmax(preds))
+
+fig, axes = plt.subplots(1, 3, figsize=(16, 4.5))
+fig.suptitle("TabPFN predicted distributions — diabetes dataset", fontsize=13)
+
+for ax, idx, label in zip(
+    axes,
+    [low_idx, mid_idx, high_idx],
+    ["low prediction", "median prediction", "high prediction"],
```
bejaeger: could you explain in the plot what low, median, and high predictions mean?

examples/plot_regression_distribution.py
```diff
@@ -0,0 +1,49 @@
...
+):
+    plot_regression_distribution(reg, X_test[idx], ax=ax)
```
bejaeger: if you change the interface (see other comment) you can also predict three points in a batch and then just pass the results here, so it's a bit more flexible.

src/tabpfn/visualisation/regression_distribution.py
```diff
@@ -0,0 +1,130 @@
...
+        band = (centers >= lo) & (centers <= hi)
+        pct = round((quantile_interval[1] - quantile_interval[0]) * 100)
+        ax.fill_between(centers[band], density[band], color=color, alpha=0.3, lw=0)
+        legend_handles.append(
+            Patch(facecolor=color, alpha=0.5, lw=0, label=f"{pct}% interval")
+        )
+
+    for name in statistics:
+        value = float(np.atleast_1d(out[name])[0])
+        c, ls = _STAT_STYLES[name]
```
bejaeger, line 131: would throw a cryptic error if name is not in STAT_STYLES. Maybe raise a nicer error or add types for the modes.

bejaeger (changes requested):
Thanks @eliott-kalfon , looks great!
Just a suggestion for the API and a few nits.
Also, let's move to british spelling -> "visualisation" everywhere.

src/tabpfn/visualisation/regression_distribution.py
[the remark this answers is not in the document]
eliott-kalfon: Ok, then I will add this

src/tabpfn/visualization/regression_distribution.py
eliott-kalfon: Sure, that could work too - let me make this change

examples/plot_regression_distribution.py
eliott-kalfon: Fair point

examples/plot_regression_distribution.py
eliott-kalfon: Sure, the goal was to show different prediction values and distributions (see [here](https://github.com/PriorLabs/TabPFN/blob/a18fc4ed28f8a7e5f667f4925986dabc1f42253f/examples/plot_regression_distribution.py#L24-L28)), I can remove the qualifier and name them Row A, B and C. What do you think?

bejaeger (approved):
Thanks for the changes @eliott-kalfon , looks great!
I approved but have one more thing I forgot earlier:
Can you also add a test for the example? You would have to add a main() method again but then you can run it in a test and simply check there is no error and a plot is saved?

PR #993, opened 2026-06-03
Single-file TabPFN v2: add disk embeddings, KV cache, and base-checkpoint loading

priorphil:
## Why

We previously moved to single-file model implementations (replacing the multi-file `architectures/base/` setup), but the initial single-file v2 was missing a few public features, so we couldn't finish the transition and delete the old code. This is the **first of three stacked PRs** that completes the single-file v2.

## What

Copies the reference single-file v2 implementation in as `architectures/tabpfn_v2_sf.py` (registered alongside the others) and adds a comparison test that verifies it produces **identical** outputs to the multi-file `base` architecture. Then adds the three missing public features, each with tests:

1. **Load positional embeddings from disk** — for the production seed (42) and a 48-d embedding subspace, the first 2000 columns use `architectures/shared/tabpfn_col_embedding.pt` instead of device-dependent random values, mirroring the base architecture. Adds a configurable `seed`.
2. **KV cache** — a v3-style *explicit* cache: `forward()` gains `kv_cache` / `return_kv_cache` / `x_is_test_only` and returns `(output, TabPFNV2Cache)`. The between-cells attention caches the first (multi-query) train K/V head per block; the cache also carries the encoder/y-encoder fitted normalization state and the embedded all-NaN target column so test-only inference needs no training rows. Routes through `InferenceEngineExplicitKVCache`.
3. **Load base-architecture checkpoints** — a `load_state_dict` override translates the old `transformer_encoder.*` / `decoder_dict.*` key names to the single-file layout (splitting the fused QKV weight, transposing the out-projection). Encoder/y-encoder keys are shared and pass through unchanged.

## Follow-ups (stacked)

2. Move the `get_encoder` / `get_y_encoder` preprocessing into the single file.
3. Delete `architectures/base/`, rename `tabpfn_v2_sf` → `tabpfn_v2`, and drop the comparison test.

## Testing

`tests/test_architectures/test_tabpfn_v2_sf.py` — 12 tests (output/gradient equivalence vs base incl. the disk-embedding path, save-peak-memory & checkpointing equivalence, disk-embedding unit tests, KV-cache build/use/quantize, and base-checkpoint loading). Full `tests/test_architectures/` suite green (92 passed, 8 skipped).

### DIFF
--- a/src/tabpfn/architectures/tabpfn_v2_sf.py
+++ b/src/tabpfn/architectures/tabpfn_v2_sf.py
@@ -6,8 +6,9 @@ disk (for the production seed 42, see :meth:`TabPFNV2.add_column_embeddings`) an
 supports a KV cache for fast inference (an explicit cache passed through
 :meth:`TabPFNV2.forward`, see :class:`TabPFNV2Cache`).
 
-Note that this version is not compatible with the original checkpoints as layers
-have been renamed/restructured.
+Although the transformer layers have been renamed/restructured relative to the base
+architecture, checkpoints trained with the base architecture can still be loaded: see
+:meth:`TabPFNV2.load_state_dict`, which translates the old key names on the fly.
 
 Copyright (c) Prior Labs GmbH 2025.
 """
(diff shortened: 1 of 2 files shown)


priorphil:

- #996
    - #993 ◀
        - #995

priorphil:
Split into three stacked PRs for easier review: #997 (load base checkpoints), #998 (disk positional embeddings), and #999 (KV cache, now compatible with chunked evaluation). Closing in favour of those.

PR #995, opened 2026-06-03
[part of PRI-245] Single-file TabPFN v2: inline preprocessing (get_encoder/get_y_encoder)

priorphil:
## Why

Stacked on #993. The single-file `tabpfn_v2_sf` still imported `get_encoder` / `get_y_encoder` from the multi-file `base` architecture — its last dependency on `base`. To be able to delete `architectures/base/` (the intended final step of this effort), the preprocessing needs to live in the single file, like the other single-file architectures.

## What

Inlines `get_encoder`, `get_y_encoder`, and the `_legacy_normalize_features_no_op` helper into `architectures/tabpfn_v2_sf.py`, importing the encoder steps directly from `architectures.encoders` (a shared package, unaffected by any `base` deletion). `tabpfn_v2_sf` now has **no dependency on `base`**.

No behaviour change: the inlined functions are copied verbatim, and the comparison test against the base architecture (from #993) still passes, confirming identical outputs.

## Testing

`tests/test_architectures/test_tabpfn_v2_sf.py` — all 12 tests green.

## Follow-up: deleting `architectures/base/` (NOT done here)

The original plan had a third step — delete `base/`, rename `tabpfn_v2_sf` → `tabpfn_v2`, drop the comparison test. Investigation shows this is **much larger and riskier than a delete+rename**, so it is intentionally left out of this stack:

- **Checkpoint loading**: released v2 checkpoints have no `architecture_name` and default to `ARCHITECTURES["base"]` (`model_loading.py:963`). Deleting `base` requires redirecting that default to the renamed single file (and verifying every released model still loads identically).
- **Public shim**: `tabpfn.model` is a deprecated-but-supported public module that re-exports `base` submodules (`attention`, `bar_distribution`, `config`, `layer`, `memory`, `mlp`, `transformer`).
- **Breadth**: ~14 `src/` modules (classifier, regressor, model_loading, inference, finetuning, model/*) and ~10 test files import `base` directly.

A safer incremental path would be: rename `tabpfn_v2_sf` → `tabpfn_v2`, register it, redirect legacy-v2 checkpoint loading to it, and keep `base/` until the remaining consumers/tests are migrated — then delete `base/` in a final dedicated PR.

### DIFF
--- a/changelog/995.added.md
+++ b/changelog/995.added.md
@@ -1,0 +1,1 @@
+Add a single file implementation of TabPFNv2. Not activated by default yet.
--- a/src/tabpfn/architectures/tabpfn_v2_sf.py
+++ b/src/tabpfn/architectures/tabpfn_v2_sf.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import dataclasses
 from abc import ABC
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Any, Literal, cast
+from typing import Any, Literal, cast
 from typing_extensions import override
 
 import pydantic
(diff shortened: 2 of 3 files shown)


priorphil:

- #996
    - #997
        - #998
            - #999
                - #995 ◀

bejaeger (approved):
LGTM!

PR #997, opened 2026-06-03
[part of PRI-245] Single-file TabPFN v2: load base-architecture checkpoints

priorphil:
## Why

Stacked on #996 (the verbatim copy + comparison test). First of three feature PRs that complete the single-file v2 (the others: disk embeddings, KV cache).

## What

- Adds a `load_state_dict` override that detects base-architecture key names (`transformer_encoder.*` / `decoder_dict.*`) and translates them to the single-file layout: split the fused QKV weight into q/k/v, transpose the out-projection, and rename the blocks and decoder. Encoder/y-encoder keys are shared and pass through unchanged.
- The comparison tests now rely on this conversion (the bespoke `_convert_state_dict_base_to_v2` test helper is removed), plus dedicated strict-load and native-round-trip tests.
- Addresses review: `TabPFNV2Config` now uses `@pydantic.dataclasses.dataclass` (like `TabPFNV3Config`/`TabPFNV2p5Config`) so it inherits the runtime validation documented in `parse_config`.

## Testing

`tests/test_architectures/test_tabpfn_v2_sf.py` — 6 tests green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #999, opened 2026-06-03
[part of PRI-245] Single-file TabPFN v2: add explicit KV cache

priorphil:
## Why

Stacked on #998. Third of the three feature PRs.

## What

Adds a v3-style **explicit** KV cache: `forward()` gains `kv_cache` / `return_kv_cache` / `x_is_test_only` and returns `(output, TabPFNV2Cache)`. The between-cells attention caches the first (multi-query) train K/V head per block; the cache also carries the encoder/y-encoder fitted normalization state and the embedded all-NaN target column, so test-only inference needs no training rows. Routes through `InferenceEngineExplicitKVCache`; `get_architecture` no longer raises for caching.

The cache build/use paths compute the between-cells attention in one go (bypassing the batch chunking used for memory saving), matching the rest of the cache design.

## Testing

`tests/test_architectures/test_tabpfn_v2_sf.py` — 12 tests green, including `test__kv_cache__matches_standard_forward` (build/use matches the standard forward), the embeddings dict, and an int8-quantized cache.

### DIFF
--- a/src/tabpfn/architectures/tabpfn_v2_sf.py
+++ b/src/tabpfn/architectures/tabpfn_v2_sf.py
@@ -500,11 +500,8 @@ class TabPFNBlock(nn.Module):
         x_BCRE = x_BRCE.transpose(1, 2).contiguous()
         kv_entry: KVCacheEntry | None = None
         if return_kv or cached_kv is not None:
-            # Build / cache paths: bypass chunking and compute the between-cells
-            # attention in one go. We either need to capture the K/V cache entry, or the
-            # cached K/V spans the full (batch * columns) dimension. The attention folds
-            # (batch, columns) into one batch dimension, so flatten and unflatten around
-            # the call.
+            # Build / cache paths: bypass chunking. This is consistent with the
+            # original behaviour.
             B, C = x_BCRE.shape[:2]
             attn_out, kv_entry = self.per_column_attention_between_cells(
                 x_BCRE.flatten(0, 1),
--- a/src/tabpfn/architectures/tabpfn_v2_sf.py
+++ b/src/tabpfn/architectures/tabpfn_v2_sf.py
@@ -1,12 +1,10 @@
-"""The TabPFN v2 model.
+"""The TabPFN v2 model, implemented as a single file.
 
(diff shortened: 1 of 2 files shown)


priorphil:

- #996
    - #997
        - #998
            - #999 ◀
                - #995

src/tabpfn/architectures/tabpfn_v2_sf.py
```diff
@@ -615,7 +805,87 @@ def add_column_embeddings(self, x_BRCX: torch.Tensor) -> torch.Tensor:
         embs = self.feature_positional_embedding_embeddings(embs)
         return x_BRCX + embs[None, None]
 
-    def forward(  # noqa: C901
+    @staticmethod
```
priorphil, line 731: Note: This is a temporary hack - will clean this up when the preprocessing / encoders are moved into this file as well.

src/tabpfn/architectures/tabpfn_v2_sf.py
```diff
@@ -477,13 +543,80 @@ def forward(
...
+            y_encoder_state=self._state_to(self.y_encoder_state, device),
+            test_y_embedding=(
+                None
+                if self.test_y_embedding is None
+                else self.test_y_embedding.to(device)
+            ),
+            train_shape=self.train_shape,
+        )
+
+    def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV2Cache:
```
bejaeger: this would trigger quantization per default which is not tested for v2. I'd vote for removing this so quantization is not used until we test it.

bejaeger (approved):
Look mostly good to me, just one comment.
Also, did you run a benchmark for validating results?

src/tabpfn/architectures/tabpfn_v2_sf.py
priorphil: Removed, great point!

priorphil:
I haven't run a benchmark, can do that before hooking it up. There's tests that ensure that (1) output is the same as the base version and (2) that using kv cache gives the same output as not using it :)

PR #1000, opened 2026-06-03
Add auto_scale_n_estimators toggle to disable feature-coverage scaling

klemens-floege:
## What

Adds a public `auto_scale_n_estimators: bool = True` constructor argument to `TabPFNClassifier` and `TabPFNRegressor`.

When a dataset has more features than a single estimator can see (more than `max_features_per_estimator` features per estimator), `scale_n_estimators_for_feature_coverage` raises `n_estimators` to the smallest value that lets every feature appear in at least one ensemble member, emitting a `UserWarning`. Until now there was no clean way to opt out — users had to reach into the preprocessor configs and raise `max_features_per_estimator`.

Setting `auto_scale_n_estimators=False` keeps `n_estimators` exactly as provided (some features may then never be sampled).

## Changes

- `scale_n_estimators_for_feature_coverage` (`preprocessing/ensemble.py`) gains an `auto_scale_n_estimators: bool = True` parameter; returns `n_estimators` unchanged when `False`.
- The auto-scaling `UserWarning` now points users to `auto_scale_n_estimators=False` to disable it.
- New `auto_scale_n_estimators` constructor arg on both estimators (placed right after `n_estimators`), stored as `self.auto_scale_n_estimators`, documented, and threaded into all three call sites.

## Verification

End-to-end fit on synthetic data (1200 features):

| `auto_scale_n_estimators` | `n_estimators` in | `n_estimators_` used | Warned? |
|---|---|---|---|
| `True` (default) | 2 | 6 (scaled) | yes |
| `False` | 2 | 2 (unchanged) | no |

`ruff check` passes on all changed files. Behavior is unchanged by default.

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -210,6 +210,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
         self,
         *,
         n_estimators: int = 8,
+        auto_scale_n_estimators: bool = True,
         categorical_features_indices: Sequence[int] | None = None,
         softmax_temperature: float = 0.9,
         balance_probabilities: bool = False,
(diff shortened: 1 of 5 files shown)


priorphil (approved):
Thanks!

src/tabpfn/classifier.py
```diff
@@ -253,6 +254,16 @@ def __init__(  # noqa: PLR0913
                  forward pass has (slightly) different input data. Think of this as an
                  ensemble of `n_estimators`-many "prompts" of the input data.
 
+            auto_scale_n_estimators:
```
LeoGrin, line 257: the docstring should mention the cap here and for the regressor

klemens-floege: done

PR #1001, opened 2026-06-03
Store Hugging Face cache in GitHub actions cache.

oscarkey:
To avoid hitting rate limits.

This is actually only used in tabpfn-private, but making the change here to avoid conflicts.

Testing: easiest option is to wait for the merge into private and then see I think. But we have the same workflow elsewhere and it works.


LeoGrin (approved):
LGTM, thanks!

PR #1002, opened 2026-06-03
Fix crash for user-specified categoricals that are all-Nan during fit

bejaeger:
Fixed a `could not convert string to float` crash when a feature declared categorical via `categorical_features_indices` is all-missing during fit but has real string values at predict. Such columns are now kept categorical instead of being demoted to a constant numeric column, so they route through the ordinal encoder consistently between fit and predict.


bejaeger:

- #1002 ◀

LennartPurucker (approved):
LGTM!

PR #1004, opened 2026-06-03
Copy PR 667 "free runner disk..." from private to public.

oscarkey:
https://github.com/PriorLabs/TabPFN-private/pull/667


bejaeger (approved):
Thanks!

PR #1009, opened 2026-06-05
Expose keep_cache_on_device on TabPFNClassifier/Regressor

LeoGrin:
## What

Adds a public `keep_cache_on_device: bool = True` constructor argument to `TabPFNClassifier` and `TabPFNRegressor`, plumbed through `create_inference_engine()` to `InferenceEngineExplicitKVCache`.

The engine already had this flag, but it was **hardcoded to its `True` default** because `create_inference_engine()` never forwarded it — so there was no public way to set it. This wires it up end-to-end. Default behavior is unchanged.

## Why

With `fit_mode="fit_with_cache"`, the transformer KV cache is built and kept resident on the inference device for every ensemble member. On large data / many estimators that can exhaust GPU memory. Setting `keep_cache_on_device=False` offloads each per-estimator cache to CPU as soon as it is built (and moves it back on demand at predict time), so resident device memory is roughly **one estimator's cache at a time** instead of all members summed.

```python
clf = TabPFNClassifier(fit_mode="fit_with_cache", keep_cache_on_device=False)
clf.fit(X_train, y_train)   # caches offloaded to CPU per-estimator
clf.predict(X_test)          # cache moved to device per call, then released
```

## Testing

- New CPU-runnable test `test__explicit_kv_cache__offload_to_cpu` asserts the flag is honored and caches land on CPU.
- Existing `test__explicit_kv_cache__keep_cache_on_device` (CUDA-gated) and `test__explicit_kv_cache__produces_outputs` still pass.
- Verified sklearn `get_params`/`clone` round-trip the new param.

---

> [!NOTE]
> **Low Risk**
> Additive API with default `True` preserves existing behavior; only affects memory/latency tradeoffs when users opt into `fit_with_cache` with `keep_cache_on_device=False`.
> 
> **Overview**
> Adds a public **`keep_cache_on_device`** parameter (default `True`) to **`TabPFNClassifier`** and **`TabPFNRegressor`**, and threads it through **`create_inference_engine()`** into **`InferenceEngineExplicitKVCache`** when **`fit_mode="fit_with_cache"`**.
> 
> With **`keep_cache_on_device=False`**, per-estimator KV caches are offloaded to CPU as they are built and moved back to the inference device on demand during predict, trading **lower resident GPU memory** for **per-call transfer overhead**. Default behavior is unchanged.
> 
> Docs/changelog note the option; tests add an end-to-end public API placement check (classifier/regressor, True/False) and broaden the engine-level “reuse on-device tensors” test across devices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3c79e5f468e092461f90679fe8f40571fe43ea2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -12,6 +12,7 @@ import torch
 from numpy.random import default_rng
 from torch import Tensor
 
+from tabpfn import TabPFNClassifier, TabPFNRegressor
 from tabpfn.architectures.interface import Architecture, PerformanceOptions
 from tabpfn.architectures.kv_cache import KVCache, KVCacheEntry
 from tabpfn.architectures.tabpfn_v3 import TabPFNV3Cache
@@ -28,6 +29,8 @@ from tabpfn.preprocessing import (
 )
 from tabpfn.preprocessing.ensemble import TabPFNEnsemblePreprocessor
 from tabpfn.preprocessing.torch import FeatureSchema
+
+from .utils import get_pytest_devices, get_pytest_devices_with_mps_marked_slow
 
 
 class _TestModel(Architecture):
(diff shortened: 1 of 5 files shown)


tests/test_inference.py
```diff
@@ -599,3 +599,52 @@ def test__explicit_kv_cache__keep_cache_on_device() -> None:
...
+
+    Exercises the full public path with the real model: constructor ->
+    ``fit`` -> ``create_inference_engine`` -> ``InferenceEngineExplicitKVCache``.
+    With ``keep_cache_on_device=True`` each cache stays on the build device;
+    with ``False`` each cache is offloaded to CPU. On the GPU CI job
+    (``TABPFN_EXCLUDE_DEVICES`` drops cpu/mps) ``device="cuda"`` so the True/False
+    placement is genuinely distinguished; on CPU jobs it is an end-to-end smoke
+    test. ``.key`` is the (int8) tensor on both the quantized and plain cache
+    entries, so the placement check holds either way.
+    """
```
bejaeger, line 620: the doc string is rather verbose. Should we make it more concise? 
Also, should we only run this on GPU? This doesn't seem to be testing anything that's not tested in other tests run on CPU?

bejaeger (approved):
LGTM, thanks! Just one comment on the tests.

priorphil (approved):

tests/test_inference.py
LeoGrin: Shortened the docstring! For running on GPU, would vote to keep it like this for simplicity given that it takes 2s and does check it doesn't crash when turning the flag on on cpu (though I agree it's a bit of an overkill)

PR #1012, opened 2026-06-07
feat: warn on label leakage (feature identical to or perfectly correlated with target)

### LINKED ISSUE #368: Warn if label leakage

Print a warning if the features contain the label or are perfectly correlated to it. 

Check correlation of the columns in X to y in training set and print warning

_No response_

### Additional context

_No response_

### Impact

None

mvanhorn:
## Summary

Adds a best-effort label-leakage check in `_validate_fit_inputs` (#368): on fit, if any feature column is identical to the target or perfectly correlated with it, a `UserWarning` is emitted naming the offending feature(s), since that almost always indicates leakage and misleadingly high scores.

## Behavior

- Detects two leakage forms per feature: exact equality with the target (over the non-NaN overlap) and `|pearson| > threshold` for numeric columns.
- Constant columns/targets are skipped (correlation undefined), and at least three valid finite pairs are required before the correlation path runs, because any two non-constant points are perfectly correlated by definition and would otherwise produce false positives on small/clean data.
- Detection is wrapped in a guard so a malformed-input failure never blocks a fit. The `warnings.warn` call is kept outside that guard, so when users run with warnings as errors (`python -W error` / `warnings.simplefilter("error")`) a genuine leakage warning still propagates instead of being silently swallowed.

## Testing

`pytest tests/test_classifier_interface.py -k leak` — all leakage tests pass, including:
- clean features do not warn; identical-column and perfectly-correlated columns warn (with feature name)
- near-perfect correlation (below threshold) does not warn
- detector errors (e.g. `corrcoef` raising) do not raise
- two-row data does not false-positive
- a real leakage warning propagates under `simplefilter("error")`

`ruff check src/tabpfn/validation.py tests/test_classifier_interface.py` passes.

Closes #368

AI was used for assistance.

### DIFF
--- a/src/tabpfn/validation.py
+++ b/src/tabpfn/validation.py
@@ -13,6 +13,7 @@ import warnings
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
+import numpy as np
 import pandas as pd
 import torch
 from sklearn.base import is_classifier
@@ -23,12 +24,14 @@ from tabpfn.misc._sklearn_compat import check_array, validate_data
 from tabpfn.settings import settings
 
 if TYPE_CHECKING:
-    import numpy as np
     import numpy.typing as npt
     import torch
 
     from tabpfn import TabPFNClassifier, TabPFNRegressor
     from tabpfn.constants import XType, YType
+
+
+_LABEL_LEAKAGE_CORRELATION_THRESHOLD = 0.9999
 
 
 def ensure_compatible_fit_inputs(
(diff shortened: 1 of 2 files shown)


noahho:
Hi @mvanhorn — thank you for this, and for the obvious care that went into it.

After discussing internally, we've decided not to add label-leakage detection to the model code itself. Our view is that this is really a system/harness concern rather than a model one.

So I'm going to close this PR and the issue. I'm sorry for the wasted effort — the issue should have been triaged before you invested time.

PR #1019, opened 2026-06-09
[Fixes PRI-247] Reduce memory usage of the single file models. 

priorphil:
Now they're below the original implementation and faster.

Base model peak memory: 2.2088 GB, time: 197 ms
TabPFN model peak memory: 2.0242 GB, time: 182 ms


bejaeger (approved):
Amazing!

PR #1020, opened 2026-06-09
Upgrade uv so that exclude-newer = "7 days" works

brendan-priorlabs:


oscarkey (approved):
lgtm for uv change, not sure why ci is failing though

oscarkey:
hoping I've fixed it in https://github.com/PriorLabs/TabPFN/pull/1027

oscarkey:
I think it's safe to close this PR. Looking at a recent run on main: https://github.com/PriorLabs/TabPFN/actions/runs/28363957214/job/84025180512
- The "Install uv" step installs version 0.11.25
- The "Install dependencies" step doesn't install uv
- We don't activate the venv, we use `uv run`, which should call the system uv not the uv in the venv?

But let me know if I'm missing something and we can merge ofc.

PR #1021, opened 2026-06-09
Try to make MacOS CI run on arm64 rather than x86.

oscarkey:

PR #1027, opened 2026-06-09
Add n_features_in_ to SafePowerTransformer for scikit-learn 1.9.

oscarkey:
This is causing tests to fail on main after the scikit-learn update.

In this test we set some attributes on SafePowerTransformer to pretend it's been fit. scikit-learn 1.9 requires a new attribute. It has no effect outside of the tests, where we actually call .fit() which sets the attribute correctly.


oscarkey:

- #1027 ◀
    - #1021

priorphil:
This should be fixed on main by now :) (Had to add it to get https://github.com/PriorLabs/TabPFN/pull/1019 in )

oscarkey:
oh nice thanks!!

PR #1029, opened 2026-06-09
Fix regressor ensemble members sharing one mutable target_transform

LeoGrin:
## The bug

`generate_regression_ensemble_configs` builds member configs via `product(preprocessor_configs, target_transforms)` + `_balance`, which stores the **same transformer instance by reference** in multiple `RegressorEnsembleConfig` objects (with the default `REGRESSION_Y_PREPROCESS_TRANSFORMS = (None, "safepower")` and 8 estimators, members 2, 3, 6 and 7 all share one `PowerTransformer`).

`_transform_labels_one` (`preprocessing/transform.py`) then calls `config.target_transform.fit_transform(y_train)` **in place** for each member. With the default `n_preprocessing_jobs=1`, joblib runs in-process (no pickling), so all sharing members re-fit the single shared object — after fit it holds only the *last* member's fitted state.

At predict time, `regressor._iter_borders` calls `transform_borders_one(std_borders, target_transform=config.target_transform)`, so every sharing member's output borders are inverse-mapped with the wrong fitted transform whenever per-member training targets differ — i.e. whenever row subsampling is active (`inference_config SUBSAMPLE_SAMPLES`; `FinetunedTabPFNRegressor` sets `SUBSAMPLE_SAMPLES=50_000` by default). The result is silently wrong regression predictions. With `n_preprocessing_jobs>1` the configs are pickled per worker, which masks the bug.

**Repro evidence (before the fix):** with `SUBSAMPLE_SAMPLES=0.5`, predictions differed between `n_preprocessing_jobs=1` and `n_preprocessing_jobs=2` by up to `2.3e-01` (~12.5% of the target std) on a 200-row synthetic dataset.

## The fix

Deep-copy the target transform into each member config in `generate_regression_ensemble_configs`, so every member owns the instance it fits. (`copy.deepcopy` rather than `sklearn.base.clone` so any pre-set state is preserved and clone-roundtrip quirks of custom transformers are avoided.)

This mirrors what already happens implicitly when configs are pickled for parallel preprocessing workers, making `n_preprocessing_jobs=1` and `>1` behave identically.

## Verification

- New regression test `test__generate_regression_ensemble_configs__target_transforms_not_shared` (fails on `main`).
- End-to-end: predictions for `n_preprocessing_jobs=1` vs `2` with subsampling are now bit-identical (`max diff = 0.0`, was `2.3e-01`).
- Each member's stored fitted transform now exactly matches a fresh transform fitted on that member's own (z-normalized, subsampled) training targets (`max diff = 0.0` on the inverse-transformed border grid).
- `pytest tests/test_preprocessing/` → 270 passed, 8 skipped; `pytest tests/test_regressor_interface.py` → 104 passed, 59 skipped; ruff check/format clean.

---

> [!NOTE]
> **Low Risk**
> Localized preprocessing config construction with a regression test guard; behavior change only fixes incorrect shared mutable state.
> 
> **Overview**
> Fixes a regression bug where **`generate_regression_ensemble_configs`** reused the same **`target_transform`** object across multiple **`RegressorEnsembleConfig`** entries. Per-member label preprocessing fits that transform **in place** (`_transform_labels_one`), so with **`n_preprocessing_jobs=1`** the last fitted member overwrote shared state and predictions could be wrong when members trained on different targets (e.g. row subsampling).
> 
> Each member config now gets **`copy.deepcopy(target_transform)`** so every ensemble member owns its own fitted transform. A changelog note documents the fix.
> 
> Adds **`test__generate_ensemble_configs__no_fittable_object_shared_between_members`**, which recursively checks that regression and classification configs (including nested **`Pipeline`** steps) do not share any sklearn-style fittable instance across members or with caller prototypes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 108bc03dfff55f449cd316f23a1355fca2a628e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


priorphil:
Reassigning to Ben since he has more context here :)

bejaeger (approved):
LGTM!

PR #1030, opened 2026-06-09
Fix save_fitted_tabpfn_model moving live estimator modules to CPU

LeoGrin:
## Summary

`save_fitted_tabpfn_model` snapshots fitted attributes to CPU before pickling so a model saved on a CUDA/MPS machine can be loaded on CPU. The snapshot called `.to("cpu")` on every torch object — but while `Tensor.to` returns a copy, `nn.Module.to` moves the module **in place**. Saving therefore relocated the live estimator's `FullSupportBarDistribution` modules (`znorm_space_bardist_`, `raw_space_bardist_`) to CPU, and any subsequent `predict(output_type="median"/"quantiles"/"main")` on a GPU/MPS estimator crashed with a device-mismatch `RuntimeError`.

## Fix

Deep-copy `nn.Module` values before moving them to CPU, so the snapshot is CPU-resident while the live estimator keeps its modules on the original device. Tensors keep the plain `.to("cpu")`, which already copies.

## Repro (before this fix, on MPS)

```python
model = TabPFNRegressor(device="mps").fit(X, y)
model.save_fit_state("model.tabpfn_fit")
model.predict(X, output_type="median")  # RuntimeError: device mismatch
```

## Testing

- New regression test `test__save_fit_state__does_not_move_live_estimator_to_cpu`, parametrized over available devices (MPS marked slow per the file's convention). Verified it fails on MPS without the fix and passes with it.
- `tests/test_save_load_fitted_model.py` passes fully (13 passed) locally on an MPS machine.

---

> [!NOTE]
> **Low Risk**
> Localized change to fitted-state serialization with a targeted regression test; no auth or API contract changes beyond fixing incorrect in-place device mutation.
> 
> **Overview**
> Fixes a bug where **`save_fit_state` / `save_fitted_tabpfn_model`** prepared fitted attributes for CPU serialization by calling `.to("cpu")` on every torch object. **`Tensor.to` copies**, but **`nn.Module.to` moves in place**, so saving on CUDA/MPS silently relocated the live regressor’s bar-distribution modules (`znorm_space_bardist_`, `raw_space_bardist_`) to CPU and broke later **`predict`** calls that need those modules on the model device (e.g. `output_type="median"` / `"quantiles"`).
> 
> Fitted-attribute snapshotting now **`copy.deepcopy`s `nn.Module` values before `.to("cpu")`**, leaving the in-memory estimator unchanged; tensors still use `.to("cpu")` only.
> 
> Adds a **device-parametrized regression test** that asserts bar distributions stay on the fit device after save and that median/quantile predict still work, plus a **changelog** entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fae0fd8c44628297da9aea3c3e20176b31d77ea4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bejaeger (approved):
Nice!

PR #1033, opened 2026-06-09
Align torch preprocessing steps with their CPU reference implementations

LeoGrin:
Two GPU-side preprocessing steps silently diverged from the CPU implementations they are documented to match. CPU is the reference in both fixes.

## 1. `TorchSoftClipOutliers` skipped clipping for single-row inputs

`transform()` returned the input unchanged whenever `x.shape[0] == 1`, even with valid fitted bounds in the cache. In the KV-cache predict paths (`InferenceEngineCacheKV.iter_outputs`, `InferenceEngineExplicitKVCache._call_model`) only `X_test` goes through the pipeline (`num_train_rows=0`), so **predicting exactly one sample skipped outlier clipping while predicting two or more applied it** — predictions depended on the test batch size. Observed on a v2 checkpoint with `fit_mode="fit_with_cache"` and an outlier feature value: `predict(X[i:i+1])` vs `predict(X)[i]` differed by 0.0086 in probability with a predicted-class flip. Single-row predicts are the primary use case of the KV-cache mode.

The guard was a leftover from the degenerate single-row-*fit* case, which `fit()` already handles by returning ±inf bounds (making `transform` a mathematical no-op) — the same pattern the sibling torch scalers (`TorchSquashingScaler`, `TorchSafeStandardScaler`) use: guard in fit via a neutral cache, transform unconditional. The CPU/in-model reference (`remove_outliers` in `encoders/steps/_ops.py`) clips single rows whenever bounds are provided. Fix: remove the transform-side guard.

After the fix, batch-vs-single-row predictions in `fit_with_cache` mode agree to float noise (4.8e-7, was 8.6e-3 + class flip).

## 2. `TorchAddSVDFeaturesStep` missing the CPU step's `n_features < 2` no-op guard

The CPU `AddSVDFeaturesStep` declares itself a no-op for fewer than 2 features (`TruncatedSVD` requires `n_components < n_features`). The torch step had no guard and appended one SVD column — a rescaled copy (correlation 1.000) of the lone input column. With GPU preprocessing enabled (**the default for the bundled model config**) and the default presets (`svd_quarter_components` in ensemble config 1), any single-feature dataset fed the model a 3-column matrix on the GPU path vs 2 columns on the CPU path — observed max probability diff 4.9e-2 between `ENABLE_GPU_PREPROCESSING` on/off in otherwise default usage. Fix: mirror the CPU guard (`is_no_op` marker in the fitted cache, transform adds nothing).

## Tests

- `test__transform__single_row_clipped_like_batch` — a single test row is clipped identically to the same row in a batch (fails on `main`).
- `test__transform__single_row_after_single_row_fit_is_noop` — pins that the degenerate single-row-fit case stays a no-op without the guard.
- `test__add_svd_features__single_feature_is_noop_like_cpu` — torch step adds nothing for 1 feature, asserted side-by-side against the CPU step's `is_no_op` (fails on `main`).

`tests/test_preprocessing/` + `tests/test_torch_preprocessing/` + `tests/test_consistency.py` (reference predictions): 436 passed. The original end-to-end repros confirm both divergences are gone. Ruff clean.

---

> [!NOTE]
> **Medium Risk**
> Changes inference-time preprocessing on the default GPU path (KV-cache single-sample predict and single-feature SVD), which can alter probabilities but is intentional parity with CPU.
> 
> **Overview**
> Aligns two GPU preprocessing steps with their CPU reference behavior so predictions no longer depend on test batch size or on `ENABLE_GPU_PREPROCESSING`.
> 
> **`TorchSoftClipOutliers`** no longer skips clipping when `transform` sees a single row. That guard made KV-cache inference (test-only transforms) apply outlier soft-clipping only for multi-row test batches, so single-sample `predict` could disagree with the same row in a larger batch. Degenerate single-row **fit** still yields ±inf bounds via `fit()`, so transform remains a no-op there without a transform-side check.
> 
> **`TorchAddSVDFeaturesStep`** now mirrors the CPU step’s `n_features < 2` no-op: fit stores an `is_no_op` cache marker and transform returns the input without appending an extra SVD column for one-feature datasets.
> 
> Regression tests cover single-row vs batched clipping, single-row fit no-op, and torch vs CPU SVD behavior for one feature; changelog entry documents both fixes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit addbe1e207cd19f8081a9b469eb3a7086391ffce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bejaeger:
Thanks!
FYI, I also changed the tests to more systematically test regressions for cpu vs gpu pipeline.

bejaeger (approved):

PR #1036, opened 2026-06-09
Remove the never-functional per_feature preprocessor option

LeoGrin:
## Why remove instead of fix

`PreprocessorConfig(name="per_feature")` — *"a different transformation for each feature"* — has crashed unconditionally on every dataset since the initial public release. The branch calls `rng.choice(preprocessors)` on the heterogeneous registry list; sklearn `Pipeline`s are sequence-like with different lengths, so numpy's internal `asarray` raises `ValueError: ... inhomogeneous shape` (an error since numpy 1.24). Both `np.random.Generator.choice` and the legacy `RandomState.choice` fail the same way, and the `# type: ignore` on the call dates to the original TabPFN V2 code drop — it never worked in this repository.

Even if the sampling worked, the candidate pool is unsound for the per-column position: it includes the `adaptive` ColumnTransformer (requires a DataFrame with name-pattern selectors, would crash on the ndarray input) and the multi-output `norm_and_kdi` FeatureUnion (would corrupt the one-column-per-feature schema).

No built-in preset references the option (`presets.py` / `inference_config.py` have zero hits), and since every use crashes at fit, no working user configuration can depend on it — the removal is behaviorally a no-op.

## Changes

- Drop `"per_feature"` from the `PreprocessorConfig.name` `Literal`.
- Remove the dead branch in `ReshapeFeatureDistributionsStep._create_transformers_and_new_schema` (and the now-unused `rng`).

Net: −15/+4 lines. `tests/test_preprocessing/`: 269 passed. Ruff clean.

---

> [!NOTE]
> **Low Risk**
> Removes a config value that always crashed on use; no presets referenced it, so behavior is unchanged for working setups.
> 
> **Overview**
> Removes the **`per_feature`** preprocessor mode, which was exposed on `PreprocessorConfig.name` but never worked at runtime (random per-column sampling over heterogeneous sklearn pipelines failed on fit).
> 
> **`PreprocessorConfig`** no longer allows `"per_feature"` in the `name` `Literal`. **`ReshapeFeatureDistributionsStep`** always applies the named transform from the preprocessor registry via a single `feat_transform` branch; the dead per-feature `rng.choice` path and unused RNG are gone. Docstrings are updated accordingly, and the changelog records the API removal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a638d63bcc89e2e57ac636794e983d328dd2e70a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bejaeger (approved):
LGTM!

PR #1038, opened 2026-06-10
Drop Python 3.9 support, bump minimum to 3.10, add Python 3.14

priorphil:
Note that I had to add some changes like removing `Union` as ruff rules change with the bumped python version.
I've also opted to use strict=True since this is typically safer bug-wise but might introduce crashes - let me know if 
you prefer strict=False :)

Note2: I've updated the branch protection rule as well to gate on the 3.10 tests, not 3.9.

Claude stuff below:

## Summary

Removes support for Python 3.9 (minimum is now **3.10**) and adds official support for Python **3.14**.

### Dropping 3.9 / minimum 3.10

**`pyproject.toml`**
- `requires-python = ">=3.9"` → `">=3.10"`
- Removed the `eval-type-backport>=0.2.2` dependency — it was only needed on Python 3.9 to enable `X | Y` union notation in Pydantic models (native since 3.10).
- Removed the `Programming Language :: Python :: 3.9` classifier.
- Bumped tooling targets: ruff `target-version = "py310"`, mypy `python_version = "3.10"`, pyright `pythonVersion = "3.10"`.

**`.github/workflows/ci.yml`**
- Bumped the `lowest-direct` test matrices (compatibility + GPU) from `3.9` → `3.10`.
- Updated stale comments that referenced the `>=3.9` constraint / 3.9-on-x86 limitation.

### Adding 3.14
- Added the `Programming Language :: Python :: 3.14` classifier.
- 3.14 is already exercised by the CI matrix (`highest` dependency set on Ubuntu/macOS/Windows + GPU), so no workflow changes were needed.

### Docs
- Updated the README Python-versions badge and FAQ (now `3.10`–`3.14`).
- Updated `tests/README.md` CI version note.

### Notes
- The only remaining `sys.version_info < (3, 10)` branch lives in `src/tabpfn/misc/_sklearn_compat.py`, which is vendored verbatim from scikit-learn (ruff-ignored `ALL`). Left untouched to stay faithful to upstream; it is harmless on 3.10+.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,7 +318,6 @@ jobs:
         include:
           - python-version: "3.10"
             dependency-set: lowest-direct
-            torch-override: ""
           - python-version: "3.14"
             dependency-set: highest
 
@@ -343,13 +342,6 @@ jobs:
 
       - name: Install dependencies
         run: uv sync --group ci --resolution ${{ matrix.dependency-set }}
-
-      - name: Override torch for GPU compatibility
-        if: ${{ matrix.torch-override != '' }}
-        run: |
-          uv pip install "${{ matrix.torch-override }}" \
-            --extra-index-url https://download.pytorch.org/whl/cu128 \
-            --index-strategy unsafe-best-match
 
       - *restore-hf-cache
       - *restore-model-cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,12 +321,6 @@ jobs:
             torch-override: ""
           - python-version: "3.14"
             dependency-set: highest
-            # Pin torch to 2.10.0 for now, otherwise the runner raises a
-            # `The NVIDIA driver on your system is too old` error.
-            # TODO: Update the runner
-            # We install torch separately (post-sync) so the cu128 GPU wheel can
-            # be forced without re-resolving the rest of the locked dependencies.
-            torch-override: "torch==2.10.0+cu128"
 
     runs-on: ubuntu-22.04-4core-gpu
     env:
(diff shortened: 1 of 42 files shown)


.github/workflows/ci.yml
```diff
@@ -316,17 +316,16 @@ jobs:
...
             dependency-set: highest
             # Pin torch to 2.10.0 for now, otherwise the runner raises a
             # `The NVIDIA driver on your system is too old` error.
             # TODO: Update the runner
-            # We install torch separately (post-sync) to avoid uv trying to
-            # re-resolve the lock for all requires-python versions (>=3.9),
-            # which fails because torch>=2.10.0 only supports Python>=3.10.
+            # We install torch separately (post-sync) so the cu128 GPU wheel can
+            # be forced without re-resolving the rest of the locked dependencies.
             torch-override: "torch==2.10.0+cu128"
```
brendan-priorlabs: Hmm, so do we even need this anymore? The original comment (which I think I wrote...) reads like this was only needed for 3.9

brendan-priorlabs (approved):
LGTM, but please do take a look at that one comment

brendan-priorlabs:
Thanks for this, @priorphil!

.github/workflows/ci.yml
priorphil: Will check - though pinning might still be required due the comment line above:
```
# Pin torch to 2.10.0 for now, otherwise the runner raises a
# `The NVIDIA driver on your system is too old` error.
```

priorphil: Update: Seems like dropping it works fine, so I removed the remaining torch-override parts.

PR #1039, opened 2026-06-10
[Part of PRI-245] Add KVCache for 2.5 and 2.6. Shared base class.

priorphil:


src/tabpfn/architectures/kv_cache.py
```diff
@@ -155,19 +159,34 @@ def is_populated(self) -> bool:
         """True when the cache contains valid data."""
         return any(entry.is_valid() for entry in self.kv.values())
 
+    def is_empty(self) -> bool:
+        """True when the cache has not been populated yet."""
+        return not self.is_populated()
+
     def to(self, device: torch.device | str) -> KVCache:
```
bejaeger, line 168: Would it make sense to make this abstract so we don't forget to implement it in the model-specific caches?

bejaeger (approved):
Looks great, thanks!

src/tabpfn/architectures/kv_cache.py
priorphil: Makes sense, I guess we don't ever have just the kv cache entries and use the KVCache class directly :)

PR #1042, opened 2026-06-10
Add workflow that assigns reviewers to PRs from forks+dependabot.

oscarkey:
And remove the opensource-review-team from codeowners so GitHub's automatic assignment no longer triggers.

Some important files are still covered by CODEOWNERS so will still get reviewers automatically, but the workflow should skip these.

It's unclear from the docs if the github token will have permission to assign the team as a reviewer, but I'll test it after merge.

### DIFF
--- a/.github/workflows/assign-pr-reviewer.yml
+++ b/.github/workflows/assign-pr-reviewer.yml
@@ -1,0 +1,41 @@
+# When a PR is opened from a fork or by dependabot, request a review from the
+# opensource-review-team so a reviewer is randomly assigned.
+#
+# We use this workflow instead of GitHub's built-in auto-assignment system because we
+# want to avoid auto assigning reviewers to PRs from Prior Labs team members, and GitHub
+# does not support this.
+name: Assign reviewer to external+dependabot PRs
+
+on:
+  # pull_request_target so the workflow has write permissions for PRs from forks.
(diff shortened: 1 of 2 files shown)


.github/workflows/assign-pr-reviewer.yml
```diff
@@ -0,0 +1,41 @@
...
+# opensource-review-team so a reviewer is randomly assigned.
+#
+# We use this workflow instead of GitHub's built-in auto-assignment system because we
+# want to avoid auto assigning reviewers to PRs from Prior Labs team members, and GitHub
+# does not support this.
+name: Assign reviewer to external+dependabot PRs
+
+on:
+  # pull_request_target so the workflow has write permissions for PRs from forks.
+  # We do not check out PR code, so this is safe.
```
brendan-priorlabs: Can we put in a giant all caps warning saying something like, "Under no circumstances may actions/checkout be added to this workflow. It introduces a critical security hole. Do not remove this warning!"

brendan-priorlabs (approved):
LGTM! w/ one comment

.github/workflows/assign-pr-reviewer.yml
oscarkey: good idea, done!

PR #1045, opened 2026-06-11
`.predict_proba_batched()` interface for batched classifier predicts

ggprior:
## Issue

* We have a use-case where we have many small independent datasets to process
* This doesn't use the GPU very well, in fact CPU inference is faster
* Hence the proposal of this batched API for inference (previously this was only used for finetuning), that allows utilizing the GPU better for this use-case

---

## Public API Changes

Adding `.predict_proba_batched(..)` to classifier

Note: Adding this to the regressor interface would also be possible but not as nicely as for classifier as we can't easily reuse `.forward(..)` with the batch dimension, so I've left this out for now, but let me know if I should add regressor also for completeness.

---

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -29,7 +29,7 @@ from typing_extensions import Self, deprecated
 import numpy as np
 import torch
 from sklearn import config_context
-from sklearn.base import BaseEstimator, ClassifierMixin, check_is_fitted
+from sklearn.base import BaseEstimator, ClassifierMixin, check_is_fitted, clone
 from tqdm.auto import tqdm
 
 from tabpfn.base import (
@@ -52,6 +52,7 @@ from tabpfn.inference import (
     InferenceEngine,
     InferenceEngineBatchedNoPreprocessing,
     InferenceEngineCachePreprocessing,
+    _maybe_run_gpu_preprocessing,
 )
 from tabpfn.inference_tuning import (
     ClassifierEvalMetrics,
@@ -908,6 +909,13 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
                 self.inference_precision, self.devices_
             )
 
+        # Preprocessed labels are integer-encoded [0, ..., n-1]. Needed so batched
+        # *inference* postprocessing can shape outputs; harmless for fine-tuning,
+        # where the wrapper sets these. Only set if not already provided.
+        if not hasattr(self, "n_classes_"):
+            self.n_classes_ = max(int(t.max().item()) for t in y_preprocessed) + 1
+            self.classes_ = torch.arange(self.n_classes_)
+
         feature_schema = convert_batch_of_cat_ix_to_schema(
             batch_of_cat_indices=cat_ix,
             num_features=X_preprocessed[0].shape[1],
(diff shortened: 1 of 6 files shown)


tests/test_finetuning_classifier.py
```diff
@@ -1261,3 +1261,144 @@ def get_column_order_by_mean(x: torch.Tensor) -> list[int]:
             "Column order is not different for any batch! Fixed preprocessing seed is "
             "not working."
         )
+
+
+@pytest.mark.parametrize("device", devices)
+def test__batched_inference__matches_single_dataset(device: str) -> None:
```
LeoGrin, line 1307: nit: should this maybe be in a different file if we're making that more general than finetuning?

src/tabpfn/classifier.py
```diff
@@ -925,6 +933,132 @@ def fit_from_preprocessed(
 
         return self
 
+    def predict_proba_batched(
```
LeoGrin, line 941: nit: might be a big refactor so feel free to ignore, but I feel like a nicer API would be to have fit_batched and predict_batched. (or just fit and predict supporting batched inputs too, might be even better)

src/tabpfn/classifier.py
```diff
@@ -925,6 +933,132 @@ def fit_from_preprocessed(
...
+        self,
+        X_list: list[XType],
```
LeoGrin: X_train_list, y_train_list, X_test_list?

tests/test_finetuning_classifier.py
```diff
@@ -1261,3 +1261,144 @@ def get_column_order_by_mean(x: torch.Tensor) -> list[int]:
...
+
+    The transformer batch dimension is independent, so stacking datasets along it
+    yields independent in-context predictions in one forward. This verifies both
+    the output contract ((n_query, batch_size, n_classes)) and numerical
+    equivalence with running each dataset alone from the same preprocessed inputs.
+    """
+    if device == "cuda" and not torch.cuda.is_available():
+        pytest.skip("CUDA device requested but not available.")
+
+    def mkds(seed: int, n: int = 60, f: int = 5) -> tuple[np.ndarray, np.ndarray]:
```
LeoGrin, line 1319: I think the datasets we test here are not diverse enough (no categorical etc, on which they would fail IIUC)

LeoGrin: should also test dataframe inputs?

LeoGrin: and ragged datasets if we want to support?

LeoGrin (changes requested):
Thanks! A few comments.

##  + a few issues (Lightly edited and checked claude dump):

 1. X_test skips predict-time input validation — equivalence claim breaks for non-numeric input.
  The standard path (_raw_predict, classifier.py:1467-1486) runs ensure_compatible_predict_input_sklearn, fix_dtypes, and process_text_na_dataframe(ord_encoder=...) on X_test before
  the member preprocessors. predict_proba_batched feeds raw X_test straight into member.transform_X_test() (classifier.py:1016), which is just feature-subsetting +
  cpu_preprocessor.transform (ensemble.py:68-74). For DataFrames with categoricals, text (or numpy with object dtype I think), or NaNs this will diverge from predict_proba or crash, while the docstring promises
  "preprocessed exactly as in fit() + predict_proba()". Apply the same validation per X_test, or explicitly restrict to numeric arrays.

  2. Post-processing parity gaps vs predict_proba.
  _predict_proba applies things the new method skips or mis-applies:
  - _maybe_reweight_probas (tuned classification thresholds) is never applied.
  - With balance_probabilities=True, _apply_balancing uses class_counts_ from the last dataset's fit for all datasets.
  - If inference tuning calibrated softmax_temperature_, the last dataset's temperature is applied batch-wide.
  - The USE_SKLEARN_16_DECIMAL_PRECISION rounding + renormalization is skipped (minor).

We could just raise in predict_proba_batched when tuned thresholds / balance_probabilities / calibrated temperature are in play if we want to merge this quickly?

  3. Ragged datasets are silently wrong, but the docstring advertises them.
  The docstring says differing shapes "are padded by the collator", but padding with 0.0 inserts fake context rows (X=0, y=class 0) and constant-zero feature columns — the model
  attends to them, so results are not equivalent to per-dataset prediction. Nothing masks padded positions (the new collator comment admits "masking … is left to the
  model/preprocessing", but nothing downstream does it). Padded query rows also come back as garbage probabilities in the output array, untrimmed. Tests only cover same-shaped datasets. 

  4. The "no side effect" claim is wrong — the estimator is left in a broken state.
  "Only fit_mode is restored; every other fitted attribute is overwritten — the estimator
  ends up with the last dataset's full fit state (classes, encoders, schema, tuning state, ensemble configs) except executor_, which is the fused batched engine holding all datasets' tensors. Any prior fit is destroyed, subsequent predict/predict_proba/get_embeddings/save_fit_state are broken or misleading, and the batch stays pinned in memory.

  5. Collator: stale docstring, and the relaxation also affects fine-tuning.
  meta_dataset_collator's Note (data_util.py:621-623) still says batch_size=1 is "enforced by an internal assertion" — the assertion is gone. Removing it also silently enables batch_size>1 in the existing fine-tuning DataLoader path, where padded query rows would now contribute to the loss. If that's intended, it deserves its own test; otherwise keep a guard against ragged shapes there.

## Minor

  - zip(X_list, y_list, X_test_list, strict=False) (classifier.py:1000) — lengths were just validated, so use strict=True; strict=False reads as deliberately tolerating mismatch.
  - n_test = None (classifier.py:992) is dead — it's always reassigned inside the loop before use.

### Performance note (not sure about this one)

  Each loop iteration calls full self.fit(X, y), which unconditionally calls _initialize_model_variables() (classifier.py:602) — on small datasets the model rebuild is ~170–210 ms and dominates fit time. For the PR's stated use case (many small datasets) this sequential CPU cost may eat much of the fused-GPU win. A natural follow-up is reusing models_ across the loop (the no_refit pattern fit_from_preprocessed already uses).

LeoGrin:
> Note: Adding this to the regressor interface would also be possible but not as nicely as for classifier as we can't easily reuse .forward(..) with the batch dimension, so I've left this out for now, but let me know if I should add regressor also for completeness.

I think it's fine to leave it for now if we want to merge quickly. But we could perhaps make a function in the regressor raising NotImplementedError and telling people to open an issue if they're interested? Not sure.

tests/test_finetuning_classifier.py
ggprior: moved to `test_classifier_interface.py`

src/tabpfn/classifier.py
ggprior: 👍

tests/test_finetuning_classifier.py
ggprior: Added `test__predict_proba_batched__matches_per_dataset_dataframe` with categorical, more X_test validation

ggprior:
> Thanks! A few comments.
> 
> ## + a few issues (Lightly edited and checked claude dump):
> 1. X_test skips predict-time input validation — equivalence claim breaks for non-numeric input.
>    The standard path (_raw_predict, classifier.py:1467-1486) runs ensure_compatible_predict_input_sklearn, fix_dtypes, and process_text_na_dataframe(ord_encoder=...) on X_test before
>    the member preprocessors. predict_proba_batched feeds raw X_test straight into member.transform_X_test() (classifier.py:1016), which is just feature-subsetting +
>    cpu_preprocessor.transform (ensemble.py:68-74). For DataFrames with categoricals, text (or numpy with object dtype I think), or NaNs this will diverge from predict_proba or crash, while the docstring promises
>    "preprocessed exactly as in fit() + predict_proba()". Apply the same validation per X_test, or explicitly restrict to numeric arrays.

Good point, added validation to X_test


> 2. Post-processing parity gaps vs predict_proba.
>    _predict_proba applies things the new method skips or mis-applies:
> 
> * _maybe_reweight_probas (tuned classification thresholds) is never applied.
> * With balance_probabilities=True, _apply_balancing uses class_counts_ from the last dataset's fit for all datasets.
> * If inference tuning calibrated softmax_temperature_, the last dataset's temperature is applied batch-wide.
> * The USE_SKLEARN_16_DECIMAL_PRECISION rounding + renormalization is skipped (minor).
> 
> We could just raise in predict_proba_batched when tuned thresholds / balance_probabilities / calibrated temperature are in play if we want to merge this quickly?
> 

Raising for now, but I'll also see how it would look like if we fold this into `.fit()/.predict()`.

> 3. Ragged datasets are silently wrong, but the docstring advertises them.
>    The docstring says differing shapes "are padded by the collator", but padding with 0.0 inserts fake context rows (X=0, y=class 0) and constant-zero feature columns — the model
>    attends to them, so results are not equivalent to per-dataset prediction. Nothing masks padded positions (the new collator comment admits "masking … is left to the
>    model/preprocessing", but nothing downstream does it). Padded query rows also come back as garbage probabilities in the output array, untrimmed. Tests only cover same-shaped datasets.

Now raising `ValueError` instead

> 4. The "no side effect" claim is wrong — the estimator is left in a broken state.
>    "Only fit_mode is restored; every other fitted attribute is overwritten — the estimator
>    ends up with the last dataset's full fit state (classes, encoders, schema, tuning state, ensemble configs) except executor_, which is the fused batched engine holding all datasets' tensors. Any prior fit is destroyed, subsequent predict/predict_proba/get_embeddings/save_fit_state are broken or misleading, and the batch stays pinned in memory.

docstring changed

> 5. Collator: stale docstring, and the relaxation also affects fine-tuning.
>    meta_dataset_collator's Note (data_util.py:621-623) still says batch_size=1 is "enforced by an internal assertion" — the assertion is gone. Removing it also silently enables batch_size>1 in the existing fine-tuning DataLoader path, where padded query rows would now contribute to the loss. If that's intended, it deserves its own test; otherwise keep a guard against ragged shapes there.

comment updated

> 
> ## Minor
> * zip(X_list, y_list, X_test_list, strict=False) (classifier.py:1000) — lengths were just validated, so use strict=True; strict=False reads as deliberately tolerating mismatch.
> * n_test = None (classifier.py:992) is dead — it's always reassigned inside the loop before use.

👍 , changed

> ### Performance note (not sure about this one)
> Each loop iteration calls full self.fit(X, y), which unconditionally calls _initialize_model_variables() (classifier.py:602) — on small datasets the model rebuild is ~170–210 ms and dominates fit time. For the PR's stated use case (many small datasets) this sequential CPU cost may eat much of the fused-GPU win. A natural follow-up is reusing models_ across the loop (the no_refit pattern fit_from_preprocessed already uses).

Addressed. Now an additional ~11% faster as well 👍

LeoGrin:
> The "no side effect" claim is wrong — the estimator is left in a broken state.
"Only fit_mode is restored; every other fitted attribute is overwritten — the estimator
ends up with the last dataset's full fit state (classes, encoders, schema, tuning state, ensemble configs) except executor_, which is the fused batched engine holding all datasets' tensors. Any prior fit is destroyed, subsequent predict/predict_proba/get_embeddings/save_fit_state are broken or misleading, and the batch stays pinned in memory.

> docstring changed

I think this should rather be fixed? Right now it can lead to quite confusing issues for the user, e.g after predict_proba_batched:

  - executor_ is still InferenceEngineBatchedNoPreprocessing (was InferenceEngineCachePreprocessing), prior fit destroyed — exactly as before.
  - predict() → bare AssertionError with an internal "Bad combination of inference mode…" message. The docstring says to refit, but a user who doesn't read it gets a cryptic assert.
  - get_embeddings() → TypeError: ... unexpected keyword argument 'only_return_standard_out' — an even more confusing failure from the wrong engine.
  - save_fit_state() → succeeds silently. A user who saves after a batched predict gets a misleading artifact that won't behave like a normal saved model.
  
I think ideally we make it have no side effect, or otherwise make the error messages clearer?

Otherwise LGTM!

ggprior:
> > The "no side effect" claim is wrong — the estimator is left in a broken state.
> > "Only fit_mode is restored; every other fitted attribute is overwritten — the estimator
> > ends up with the last dataset's full fit state (classes, encoders, schema, tuning state, ensemble configs) except executor_, which is the fused batched engine holding all datasets' tensors. Any prior fit is destroyed, subsequent predict/predict_proba/get_embeddings/save_fit_state are broken or misleading, and the batch stays pinned in memory.
> 
> > docstring changed
> 
> I think this should rather be fixed? Right now it can lead to quite confusing issues for the user, e.g after predict_proba_batched:
> 
> * executor_ is still InferenceEngineBatchedNoPreprocessing (was InferenceEngineCachePreprocessing), prior fit destroyed — exactly as before.
> * predict() → bare AssertionError with an internal "Bad combination of inference mode…" message. The docstring says to refit, but a user who doesn't read it gets a cryptic assert.
> * get_embeddings() → TypeError: ... unexpected keyword argument 'only_return_standard_out' — an even more confusing failure from the wrong engine.
> * save_fit_state() → succeeds silently. A user who saves after a batched predict gets a misleading artifact that won't behave like a normal saved model.
> 
> I think ideally we make it have no side effect, or otherwise make the error messages clearer?
> 
> Otherwise LGTM!

addressed in [0040774](https://github.com/PriorLabs/TabPFN/pull/1045/commits/0040774cd1ece1a48676a0a8826cf04c4a201f95) (by running on a clone)

LeoGrin (approved):
Thanks, LGTM!

One last nit: the test test__predict_proba_batched__matches_per_dataset_dataframe now tests with a DataFrame with categorical, but could be made more robust (object, strings, NaNs etc).

PR #1046, opened 2026-06-12
Fix regression checkpoint loading.

priorphil:


bejaeger (approved):
Oh boy! Thanks for the fix.

PR #1048, opened 2026-06-13
Fix fitted model archive path handling

fallintoplace:
## Summary
- Derive the `.tabpfn_fit` archive base from the file suffix instead of using a global string replacement.
- Explicitly create the destination parent directory before moving the generated archive.
- Add regression coverage for save paths whose parent directory also contains `.tabpfn_fit`.

## Why
The previous string replacement could strip `.tabpfn_fit` from parent directory names while building the temporary zip path. That could create or touch a sibling directory unrelated to the requested save path before moving the final archive.

## Tests
- `uv run --no-project --with ruff==0.15.12 ruff check src/tabpfn/model_loading.py tests/test_save_load_fitted_model.py`
- `uv run --no-project --with ruff==0.15.12 ruff format --check src/tabpfn/model_loading.py tests/test_save_load_fitted_model.py`
- `uv run pytest tests/test_save_load_fitted_model.py::test__save_fit_state__keeps_tabpfn_fit_parent_name`

### DIFF
--- a/tests/test_save_load_fitted_model.py
+++ b/tests/test_save_load_fitted_model.py
@@ -18,7 +18,7 @@ from tabpfn import TabPFNClassifier, TabPFNRegressor
 from tabpfn.architectures.interface import ArchitectureConfig
 from tabpfn.base import RegressorModelSpecs, initialize_tabpfn_model
 from tabpfn.inference_tuning import ClassifierEvalMetrics
-from tabpfn.model_loading import save_fitted_tabpfn_model, save_tabpfn_model
+from tabpfn.model_loading import save_tabpfn_model
 
 from .utils import get_pytest_devices, get_pytest_devices_with_mps_marked_slow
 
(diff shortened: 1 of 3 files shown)


tests/test_save_load_fitted_model.py
```diff
@@ -36,6 +37,21 @@ def _make_classification_data_with_categoricals() -> tuple[np.ndarray, np.ndarra
     return X_cat, y
 
 
+class _DummyExecutor:
```
adrian-prior: Do we really need these dummy classes here? Can't we just use actual estimators with `_make_regression_data()`?

adrian-prior (approved):
@fallintoplace Thanks a lot for setting this up! Overall, it looks good. One minor nit in the tests. Can you also set up a changelog for the PR (this is required -> see the failing test). I am currently looking into why the other tests are failing (this is probably not caused by you).

adrian-prior:
@fallintoplace Most of the test are green now. For [Assign reviewer to external+dependabot PRs / assign-reviewer (pull_request_target)](https://github.com/PriorLabs/TabPFN/actions/runs/27471342437/job/81393238532?pr=1048), you probably need to take a merge from `main` and the other one is the missing changelog.

fallintoplace:
@adrian-prior I will take a look in a few hours, thank you for your attention.

PR #1051, opened 2026-06-15
Temporarily disable new PR assigning workflow and revert to CODEOWNERS.

oscarkey:
The new workflow fails because it doesn't have permission to access the team:
https://github.com/PriorLabs/TabPFN/actions/runs/27471342437/job/81202694148

I want to discuss whether we can fix this without adding more permissions to the workflow, as it also uses pull_request_target. In the meantime, switch back to how things were before
https://github.com/PriorLabs/TabPFN/pull/1042


adrian-prior (approved):
Stamp!

PR #1055, opened 2026-06-16
fix: passthrough infs to the model

arthur-priorlabs:
## Issue

RES-1538

## Motivation and Context

TabPFN v3 can natively handle `inf`s in its forward pass, but we explicitly raise during `fit()` due to `validate_data(ensure_all_finite="allow-nan")`.

This PR adds an opt-in `passthrough_inf` flag that lets `±inf` values flow through preprocessing to the model instead of being rejected:

- **Validation** (`validation.py`): when `passthrough_inf=True`, `ensure_all_finite=False` so `±inf` is no longer rejected on `fit()`/`predict()`.
- **Cleaning stage** (`clean.py`): the ordinal encoder's `check_array` still rejects `inf`, so `process_text_na_dataframe` records `±inf` positions in numeric columns, replaces them with `NaN` before encoding, and writes them back into the output afterwards. The flag is threaded through `clean_data` from the estimator.
- **Ensemble pipeline** (`inf_handling.py`): `InfToNanStep` / `RestoreInfStep` bracket each member's preprocessing pipeline, recording `±inf` before the (non-finite-intolerant) steps and restoring them after, so the original infinities reach the model.

When `passthrough_inf=False` (the default), behaviour is unchanged: `±inf` is rejected at validation.

---

## Public API Changes

-   [ ] No Public API changes

Adds a `passthrough_inf: bool = False` constructor argument to `TabPFNClassifier`, `TabPFNRegressor`, and the finetuned estimators. Default `False` preserves existing behaviour, so this is backwards compatible.

---

## How Has This Been Tested?

Added `tests/test_preprocessing/test_inf_handling.py` covering:

- `InfToNanStep` / `RestoreInfStep` unit behaviour (replace, record per-feature mask, no-op without infinities, round-trip restore).
- Pipeline wiring: the inf steps bracket the pipeline only when `passthrough_inf=True`, and propagation of the flag through the ensemble config generators.
- Validation accepts `±inf` when enabled and rejects it when disabled.
- Preprocessing is identical with/without the flag on finite input.
- End-to-end `fit()` + `predict()` for both `TabPFNClassifier` and `TabPFNRegressor`, with `±inf` injected into `X`, for both flag values (succeeds when enabled; raises `TabPFNValidationError` when disabled). This reproduces and guards against the original `clean_data` crash.

All run locally and pass.

---

## Checklist

---


src/tabpfn/preprocessing/clean.py
```diff
@@ -266,11 +330,26 @@ def process_text_na_dataframe(
     else:
         X_encoded = X.to_numpy()
 
-    string_cols_ix = [X.columns.get_loc(col) for col in string_cols]
+    # The encoder may reorder columns, so map input columns to their output
+    # positions instead of assuming the encoded array stays positionally aligned.
+    output_position = _encoded_column_positions(ord_encoder, X.columns)
```
bejaeger: are you sure this is needed?
We should be using a `OrderPreservingColumnTransformer` for the ordinal_encoder so this doesn't happen :)

arthur-priorlabs: Fixed in 55001a8 + documented in a345f37

src/tabpfn/preprocessing/clean.py
```diff
@@ -235,17 +243,36 @@ def process_text_na_dataframe(
...
+    inf_col_ix: list[int] = []
+    pos_inf = neg_inf = None
+    if passthrough_inf:
+        numeric_cols = X.select_dtypes(include=["number"]).columns
+        inf_col_ix = [X.columns.get_loc(c) for c in numeric_cols]
+        numeric_values = X[numeric_cols].to_numpy()
+        pos_inf = np.isposinf(numeric_values)
+        neg_inf = np.isneginf(numeric_values)
+        if pos_inf.any() or neg_inf.any():
+            X[numeric_cols] = X[numeric_cols].replace([np.inf, -np.inf], np.nan)
```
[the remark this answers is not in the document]
arthur-priorlabs: fixed type hints in  https://github.com/PriorLabs/TabPFN/commit/a345f37da2d1ee052f67fcadb7ecd1ee865c765d

src/tabpfn/preprocessing/clean.py
```diff
@@ -273,4 +300,13 @@ def process_text_na_dataframe(
...
     )
-    return typing.cast("np.ndarray", X_encoded.astype(np.float64))
+    X_encoded = X_encoded.astype(np.float64)
+
+    # Write the recorded +/-inf values back into their original numeric cells.
+    if passthrough_inf and inf_col_ix and (pos_inf.any() or neg_inf.any()):
+        sub = X_encoded[:, inf_col_ix]
+        sub[pos_inf] = np.inf
+        sub[neg_inf] = -np.inf
+        X_encoded[:, inf_col_ix] = sub
```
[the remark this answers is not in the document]
arthur-priorlabs: fixed type hints in  https://github.com/PriorLabs/TabPFN/commit/a345f37da2d1ee052f67fcadb7ecd1ee865c765d

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -76,24 +153,40 @@ def from_only_categorical_indices(
...
+                positional ``f{i}`` names are generated. Callers that reorder
+                columns are responsible for passing names already in output order.
         """
         numerical_indices = [
             i for i in range(num_columns) if i not in categorical_indices
         ]
         if not numerical_indices and not categorical_indices:
             return cls(features=[])
 
+        chosen_names = [f"f{i}" for i in range(num_columns)] if names is None else names
```
bejaeger: nit: can we make this list comprehension a helper and also use in `build_input_feature_names`. Then it's guaranteed to be the same

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -102,6 +195,11 @@ def feature_names(self) -> list[str | None]:
         """Get list of feature names (derived from features list)."""
         return [f.name for f in self.features]
 
+    @property
+    def feature_name_set(self) -> set[str]:
+        """Set of non-``None`` feature names currently in use."""
+        return {f.name for f in self.features if f.name is not None}
```
bejaeger: `f.name is not None` is always true now it seems?<br>what is this used for? could also just be `set(feature_names)`

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -157,23 +255,37 @@ def append_columns(
...
-        chosen_names = [None] * num_new if names is None else names
-        if len(chosen_names) != num_new:
-            raise ValueError(f"Expected {num_new} names, got {len(chosen_names)}")
+        if names is None:
+            prefix = name_prefix or "added"
+            names = [f"{prefix}_{i}" for i in range(num_new)]
+        if len(names) != num_new:
+            raise ValueError(f"Expected {num_new} names, got {len(names)}")
 
+        unique_names = make_names_unique(names, existing=self.feature_name_set)
```
bejaeger: I'm wondering, could we make `make_names_unique` a part of a `FeatureSchema` `__init__` ? Then we don't need to manually do the overlap removal but feature schema takes care of it?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -213,11 +325,7 @@ def update_from_preprocessing_step_result(
         new_features = list(self.features)
         for step_idx, original_idx in enumerate(original_indices):
             step_feature = new_schema.features[step_idx]
-            new_features[original_idx] = Feature(
-                name=step_feature.name,
-                modality=step_feature.modality,
-                scheduled_gpu_transform=step_feature.scheduled_gpu_transform,
-            )
+            new_features[original_idx] = copy(step_feature)
```
bejaeger: is a `copy` enough if `Feature` evolves? should we do `deepcopy`?

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -429,6 +514,7 @@ def _maybe_append_added_columns(
         X: np.ndarray | torch.Tensor,
         feature_schema: FeatureSchema,
         result: PreprocessingStepResult,
+        step: PreprocessingStep,
```
bejaeger: nit: could just pass `step.added_feature_prefix()` instead of the whole object

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -20,6 +26,54 @@
...
+
+
+def _input_names(feature_schema: FeatureSchema) -> list[str]:
+    """Names of the input features.
+
+    Every feature is named by the time it reaches this step (input features are
+    named from the DataFrame columns or positionally; features added by earlier
+    steps are named from their transform). ``Feature.name`` is therefore non-None.
+    """
+    return [f.name for f in feature_schema.features]
```
bejaeger: Feature.name is always non-none according to the type hint, so this function doesn't seem to be needed, as it's only a one-liner

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -47,6 +48,102 @@
...
+    # non-None (input features named from columns/positionally; added features
+    # named from their transform).
+    input_names = [f.name for f in feature_schema.features]
+    names: list[str] = []
+    n_transformed = 0
+    for col in layout:
+        if col.is_passthrough:
+            names.append(input_names[col.source_ix])
+        else:
+            names.append(f"reshape_{n_transformed}")
```
bejaeger, line 126: should this still carry over the name of the feature that is transformed?<br>or is that the reason you introduced `ancestor`? what if there are multiple transforms?

tests/test_preprocessing/test_feature_name_uniqueness.py
```diff
@@ -0,0 +1,327 @@
...
+        f"schema has unnamed (None) features: {names}"
+    )
+    assert len(names) == len(set(names)), (
+        f"schema has duplicate feature names: "
+        f"{[n for n in set(names) if names.count(n) > 1]}"
+    )
+
+
+@contextlib.contextmanager
+def assert_every_schema_unique() -> Iterator[None]:
```
bejaeger, line 50: could we remove some of these tests if FeatureSchema handles the "making unique" part?

bejaeger (commented):
Great work!<br>Most of my comments are nits and things for you to consider. I'll have another look when you ping me again. Thanks!

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: Fixed in f0b307f.

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: Thanks, I thought I'd caught them all!
You're right, this whole property is unneeded (fixed in 08f75a9)

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: We could, but in general feature names are unique at FeatureSchema construction time (this is tested by [test_feature_name_uniqueness.py](https://github.com/PriorLabs/TabPFN/blob/arthur/res-1538-passthrough-infs-to-the-model/tests/test_preprocessing/test_feature_name_uniqueness.py), along with a check after `fit_transform`).
If we don't mind the overhead we could enforce this at construction time instead.

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: You're right, I initially thought `deepcopy` would also (needlessly) clone immutable fields, but a quick test showed that that is not the case, so little harm in using `deepcopy` (4µs vs 1 µs with copy).
Fixed in 783c464.

src/tabpfn/preprocessing/pipeline_interface.py
arthur-priorlabs: Fixed in 39a37c8.

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
arthur-priorlabs: Fixed in 7dbff94.

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
arthur-priorlabs: That's the reason why I've introduced ancestor: the idea is that `ReshapeFeatureDistribution` may either replace the column, or append to the columns, so we can't carry the name over in general if we want them to be unique.
Ancestry fixes that: if the new feature column B is derived from A, we set `B.ancestor = A`.

tests/test_preprocessing/test_feature_name_uniqueness.py
arthur-priorlabs: Yep, I just would rather assert that at implementation/pytest time rather than compute-time for every `FeatureSchema`, but your call 😄

arthur-priorlabs:
Drafting as the model itself also seems to pass `inf`s through 😓

src/tabpfn/classifier.py
```diff
@@ -238,6 +240,7 @@ def __init__(  # noqa: PLR0913
         eval_metric: str | ClassifierEvalMetrics | None = None,
         tuning_config: dict | ClassifierTuningConfig | None = None,
         show_progress_bar: bool = False,
+        passthrough_inf: bool = False,
```
ggprior: Random Q, or should this rather be the default behavior, and this flag should be opt-out? Technically a breaking change, but very minor (i.e., a dataset that previously raised an error would now be processed correctly)

arthur-priorlabs: Very good point!
I've kept it as a feature flag right now so I can test for regressions; it would make sense to remove it before merging

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
bejaeger: Okay got it! thanks

tests/test_preprocessing/test_feature_name_uniqueness.py
bejaeger: hm....okay. Happy to leave as is.

src/tabpfn/classifier.py
bejaeger: Hm...yeah good point. I wonder though if there is value in being strict here. This allows the user to decide what to do with the Inf's and potentially remove corrupted samples/columns.

arthur-priorlabs: Maybe a middle ground would be suggest to enable `passthrough_inf` in the exception we raise?

bejaeger:
Will have a look at it tomorrow, promise!

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -273,6 +273,7 @@ def __init__(  # noqa: PLR0913
         save_checkpoint_interval: int | None = 10,
         use_fixed_preprocessing_seed: bool = True,
         experiment_logger: FinetuningLogger | None = None,
+        passthrough_inf: bool = False,
```
bejaeger: doc string seems missing, same for `finetuned_classifier` and `finetuned_regressor`

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -47,6 +48,102 @@
...
+    n_transformed: int,
+    append_to_original: bool,
+    apply_to_categorical: bool,
+) -> list[_ReshapeColumn]:
+    """Describe the reshape step's output columns, in output order.
+
+    Single source of truth for the output column layout, mirroring the
+    ``ColumnTransformer`` assembled in ``_create_transformers_and_new_schema``;
+    names (:func:`_build_reshape_output_names`) and ancestors
+    (:func:`_build_reshape_ancestors`) are both derived from it so they can't
```
bejaeger: btw, we don't want to use the `:func:`, .. etc. markers as we don't intend to generate automatic docs from this. could you try to remove them.

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
```diff
@@ -47,6 +48,102 @@
...
+    """Map each output column back to the input feature it derives from.
+
+    An entry is the *name* of the source input feature for distribution-
+    transformed columns, or ``None`` for passthrough columns, which already
+    carry their source name directly.
+    """
+    input_names = [f.name for f in feature_schema.features]
+    return [
+        None if col.is_passthrough else input_names[col.source_ix] for col in layout
+    ]
```
bejaeger: is this extra fn needed or should we make it part of `set_ancestors`?

src/tabpfn/preprocessing/clean.py
```diff
@@ -229,23 +255,152 @@ def _align_columns_to_fitted_dtypes(
     return X
 
 
+def inf_masks_pandas_only(
```
bejaeger: could this be private? same for `inf_masks_numpy_numeric_` ?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -5,13 +5,88 @@
...
+        suffix = 1
+        while candidate in seen:
+            candidate = f"{name}_{suffix}"
+            suffix += 1
+        seen.add(candidate)
+        out.append(candidate)
+    return out
+
+
+def default_input_feature_names(num_features: int) -> list[str]:
```
bejaeger: could be private?

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -42,18 +117,32 @@ class GPUTransformType(str, Enum):
 class Feature:
     """A single feature with its name and modality.
 
+    Warning: features are computed/updated at `fit()`-time only.
+    Existing/transformed columns carry over their features, using
+    `dataclasses.replace()` if needed instead of implicitly copying with
+    `Feature()`.
```
bejaeger: is this comment necessary here?

src/tabpfn/preprocessing/pipeline_interface.py
```diff
@@ -16,6 +16,96 @@
...
+        mask = inf_masks.get(feat.name)
+        if mask is None:
+            continue
+        if bool((mask != mask[0]).any()):
+            new_features[idx] = dataclasses.replace(feat, non_constant_inf=True)
+            changed = True
+
+    if not changed:
+        return feature_schema
+    return dataclasses.replace(feature_schema, features=new_features)
```
bejaeger, line 95: what about using the `FeatureModality` enum to carry that information?
IMO it would make sense and would be a bit cleaner than introducing a special attribute in `Feature` for this case.

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -42,18 +117,32 @@ class GPUTransformType(str, Enum):
...
+            value (so they are genuinely non-constant once restored). Honoured by
+            :class:`RemoveConstantFeaturesStep`.
     """
 
-    name: str | None
+    name: str
     modality: FeatureModality
     scheduled_gpu_transform: GPUTransformType | None = None
+    ancestor: str | None = None
+    non_constant_inf: bool = False
```
bejaeger: I think I'd use the `modality` instead of introducing a new, specialized attribute `non_constant_inf`. (see other comment as well)

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1861,13 +1870,15 @@ def _preprocess_raw(
...
-        if scaler_cache is not None:
-            x_RiBC = self.standard_scaler.transform(x_RiBC, fitted_cache=scaler_cache)
-        else:
-            x_RiBC = self.standard_scaler(x=x_RiBC, num_train_rows=num_train)
+        if scaler_cache is None:
+            # Fit on the imputed train rows (matching ``__call__``'s split), so the
+            # returned statistics are exactly those applied here and can be reused.
+            fit_data = x_RiBC[:num_train] if num_train > 0 else x_RiBC
+            scaler_cache = self.standard_scaler.fit(fit_data)
+        x_RiBC = self.standard_scaler.transform(x_RiBC, fitted_cache=scaler_cache)
```
bejaeger, line 1878: Could you add a test that fails without this change and is fixed with it?
Also, seems like we'll have to do the same changes for all other archs.

bejaeger (changes requested):
Thanks a lot for all the changes and work on this!
Gave another set of comments, we're close!

src/tabpfn/finetuning/finetuned_base.py
arthur-priorlabs: Fixed in 977a3ee0fafc5c48f2c74e3eea8c9581a59553cc

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
arthur-priorlabs: Fixed in a7592b4a2266cb71a95851236fcb298350f30daa

src/tabpfn/preprocessing/steps/reshape_feature_distribution_step.py
arthur-priorlabs: Fixed in bd082f5dfceeaa7703f921c8e42a74f94b936c72

src/tabpfn/preprocessing/clean.py
arthur-priorlabs: Fixed in c26a42805da3ee9b05b8786d44a9f0b256360e4f

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: Done in 9b886e452c9a234e97b88bfbdd77bc3469d2d43d

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: The warning definitely is, took me quite a while to figure out that you couldn't expect to be able to modify features in-place at `transform()` time and have them be consistent with everything else 😓

For the `dataclasses.replace` bit, there were so many `Feature(feature.name, feature.modality, ...)` instances that I felt the need to put this. Maybe now that we're using `replace` everywhere people (and Claude) will pick up on it.
Removed in c26a42805da3ee9b05b8786d44a9f0b256360e4f.

src/tabpfn/preprocessing/pipeline_interface.py
arthur-priorlabs: Any modality can carry `inf`s, so it's really orthogonal to the modality.

src/tabpfn/preprocessing/datamodel.py
arthur-priorlabs: See above.

src/tabpfn/architectures/tabpfn_v3.py
arthur-priorlabs: For the v3 test, see: https://github.com/PriorLabs/TabPFN/blob/e2beb6725da8ece2ceebca5df792eef864d06b5d/tests/test_preprocessing/test_inf_handling.py#L613-L661

I've added the same test for all 2.x archs in 02c6757; 2.6 works when removing "fit_with_cache" guard (done in f34bf62, will submit a separate PR).

src/tabpfn/preprocessing/datamodel.py
bejaeger: okay makes sense, happy to keep it!

src/tabpfn/preprocessing/pipeline_interface.py
bejaeger: hm...that's right. 
nitty but could we rename to `non_constant_with_inf` in that case? might make it clearer.
This would also allow us to decide how to treat infs for different modalities. E.g. for categoricals they should be treated as MCAR, for numericals more like MNAR.

src/tabpfn/validation.py
```diff
@@ -31,6 +31,23 @@
...
     from tabpfn.constants import XType, YType
 
+    T = TypeVar("T")
+
+
+def _resolve_early_config(
+    estimator: TabPFNRegressor | TabPFNClassifier, key: str, default: T
+) -> T:
+    """Resolves a config key, potentially before ``_initialize_model_variables()``
+    and config validation.
```
bejaeger: Why is this needed?
Doesn't that mean changing the flag doesn't have an effect because we evaluate it earlier here?

Could you point to the case where the `ensure_compatible_fit_inputs...` run before `_initialize_model_variables` and this is needed?

pyproject.toml
```diff
@@ -83,7 +83,7 @@ dev = [
 # The minimum subset of the dev dependencies required to run the tests on the CI.
 # The idea is to be as close to the deployment environment as possible.
 ci = [
-  "licensecheck>=2025.1.0",
+  "licensecheck>=2025.1.0,!=2026.0.2,!=2026.0.3",
```
bejaeger: why this change? Could you add a comment?

src/tabpfn/preprocessing/pipeline_interface.py
arthur-priorlabs: Done in da9c351

src/tabpfn/validation.py
arthur-priorlabs: Sorry, I'd added the info in the commit message (be84bff).

This happens in `FinetunedTabPFNBase._fit()`:

https://github.com/PriorLabs/TabPFN/blob/da9c351d0c00f4eaa6b7ff931ec5998b1ccfaa01/src/tabpfn/finetuning/finetuned_base.py#L643-L672

pyproject.toml
arthur-priorlabs: Versions 2026.0.2 and 2026.0.3 are just borked and crash the CI

src/tabpfn/validation.py
bejaeger: Okay thanks! could you move `self.finetuned_estimator_._initialize_model_variables()` in the snippet you posted before the call to `ensure_compatible_fit_inputs` ?<br>Then we don't need `_resolve_early_config`it seems?

arthur-priorlabs: That's right, the tradeoff is that we'll need to wait for the (much slower) `self.finetuned_estimator_._initialize_model_variables()` to complete in case `ensure_compatible_fit_inputs` would have failed early

pyproject.toml
arthur-priorlabs: Documented in 3b0011e

src/tabpfn/validation.py
arthur-priorlabs: Fixed in fe14384

bejaeger: Yeah, but that's totally fine for the finetuning use case.

bejaeger (approved):
Thanks for all the work and changes! Excited to get this in.

PR #1058, opened 2026-06-16
Update + fix Claude code review workflow.

oscarkey:
This changes updates it with the changes we've found to work well on our internal version:
https://github.com/PriorLabs/fomo-fitting/blob/main/.github/workflows/claude-code-review.yml

- Dropped synchronize trigger to reduce code
- pull-requests: write: needs to be able to write comment back to PR
- Skip Dependabot PRs
- Top-level permissions: {}
- `--comment` + only comment if needed: actually post a comment back, but only if something to say
- enable mcp__github_inline_comment__create_inline_comment: otherwise it can't post inline comments
- show_full_output gated on runner.debug == '1': so we can obtain model output for debugging
- Bumped action SHA to d44caa1 (although I see a new one was released yesterday, dependabot will pick that up)

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,8 +1,13 @@
 name: Claude Code Review
 
 on:
+  # `synchronize` intentionally excluded: the auto-review is for PR
+  # terminal states (first open / marked-ready / reopened), not every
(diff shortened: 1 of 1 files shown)


noahho (approved):

noahho:
@oscarkey ready to merge?

PR #1068, opened 2026-06-23
Fixing fine-tuning script issues

### LINKED ISSUE #1064: finetune leads to worse performance
I run the origin finetune code in the repository(on higgs), but it seems tabpfn perform worse after finetuning. I try other datasets on openml such as covertype and also try my own data, but also behave worse. Does someone meet the same problem ?

alanprior:
### What & why
Users reported that `FinetunedTabPFN{Classifier,Regressor}` made performance *worse*, not better. Root cause was two bugs that made fine-tuning look broken regardless of the actual training, plus example hyperparameters that no longer suit the current default model:

  1. **Wrong model generation.** The fine-tuner was hard-coded to TabPFN **V2.5**, while a default `TabPFNClassifier()`/`TabPFNRegressor()` (the natural baseline, and the examples' "Default"
  print) loads **V3**. So users were comparing a fine-tuned *older* model against a *newer* base — worse by construction.
  2. **Degraded-weights fallback.** With early stopping, if no epoch beat the default model on validation, the wrapper returned the **last (degraded) epoch's** weights instead of the original
  — i.e. it could hand back a model strictly worse than the base it started from.

 After the fixes, fine-tuning is never silently worse than base, and the shipped examples now *improve* over base:

* Higgs: ROC `0.8211 → 0.8279`, log-loss `0.5157 → 0.5057`
* California: MSE `0.1388 → 0.1363`

## Issue

https://github.com/PriorLabs/TabPFN/issues/1064

## Motivation and Context

---

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?

The script now produces improved results, apart from some internal evaluation.
---

## Checklist

-   [X] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.

---


src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -210,7 +211,7 @@ def _metric_name(self) -> str:
     def _create_estimator(self, config: dict[str, Any]) -> TabPFNClassifier:
         """Create the TabPFNClassifier with the given config."""
         return TabPFNClassifier.create_default_for_version(
-            version=ModelVersion.V2_5,
+            version=self.finetune_model_version,
```
psinger-prior, line 214: I am wondering if we could just not set the version so it will always use the latest one

psinger-prior (approved):
lgtm!

src/tabpfn/finetuning/finetuned_classifier.py
alanprior: Thanks!
It is possible but I preferred to explicitly specify it.

Issue #1066, opened 2026-06-22
Single-file implementation of TabPFN-v2.5 exhibits unnecessary memory overhead during inference


deepdoubledescent:
### Describe the bug

`TabPFNBlock.forward` contains calls to `.contiguous()` to allow for downstream in-place modification of Tensors in `chunked_evaluate_maybe_inplace`. `torch.Tensor.contiguous()` creates a copy of the original Tensor, and currently, the original Tensor stays in memory after, such that the whole `x_BRCD` (which corresponds to effectively the whole dataset) is duplicated in memory throughout a whole `TabPFNBlock.forward` pass.

Commit [[a1cb358](https://github.com/PriorLabs/TabPFN/commit/a1cb35894f6813dae29230c33d10230a12f926ce)](https://github.com/PriorLabs/TabPFN/commit/a1cb35894f6813dae29230c33d10230a12f926ce) tries to address this issue by deleting any local references to `x_BCRE` within `TabPFNBlock.forward` after calling `.contiguous()`. Unfortunately, this does not work, since these references are only locally confined to `TabPFNBlock.forward`.

`TabPFNBlock.forward` gets called in a loop:

```
for block in self.blocks:
    if force_recompute_layer:
        ...
    else:
        x_BRCD = block(
            x_BRCD, num_train_and_thinking_rows, save_peak_memory_factor
        )
```

Apparently, due to the way how Pythons garbage collection works, the line `x_BRCD = block(...)` will keep the reference to `x_BRCD` alive _until_ `block(...)` _returns_. Thus, deleting the local reference to `x_BCRE` within `TabPFNBlock.forward` will have no effect, since there will still be this external reference to it. Calling `x_BCRE.contiguous()` will create a new copy of `x_BCRE`, but the old one cannot be deleted since there is still the external `x_BRCD` holding a reference to it.

**How to replicate**:

See "Steps/Code to Reproduce"-section

**Suggested fixes**:

1) Remove `.contiguous()` in `TabPFNBlock.forward`. From my observation, `.contiguous()` has no effect on performance or memory management. On the contrary, even when applying the fix below, `.contiguous()` will create a whole copy of the input data and cause a short memory spike.
2) If `.contiguous()` is required, my LLM of choice alternatively suggests to change the signature of `TabPFNBlock.forward` to contain only an indirect reference to `x_BCRD`:

```
for block in self.blocks:
    
    # wrap x_BRCD and delete the direct reference
[...]

PR #1070, opened 2026-06-24
Free tensor to reduce peak mem. Enable Flash attn on v2_6.

priorphil:
## Issue
Reduces memory as mentioned in https://github.com/PriorLabs/TabPFN/issues/1066.
Note that the contiguous was not the issue here (benefit of removing it was a measurement issue) but the input tensor staying alive during the whole block was real. This PR now frees the input tensor.

On a dataset of size (10000, 50)

 v2_5 = OlD 564 MB vs NEW 425 → ~139 MB saved
 v2     = OLD 978 MB vs 918 → ~60 MB saved

v2_6 didn't use flash attention on MPS before. After fixing this, activation memory is the same as v2_5.

### DIFF
--- a/changelog/1070.changed.md
+++ b/changelog/1070.changed.md
@@ -1,0 +1,1 @@
+Reduce memory usage for v2.x architectures. Enable flash attention on MPS for v2_6.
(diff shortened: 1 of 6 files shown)


oscarkey (approved):
ooooh nice! Much clearer comments than mine too.

priorphil:
@oscarkey would you like to take another look? I had to change the approach slightly, somehow set_() crashes on the CI mac runners (but passes locally)

oscarkey (approved):
still lgtm!

PR #1071, opened 2026-06-25
Add signal to the consistency data to reduce variance in softmax due to different BLAS backends on different platforms.

priorphil:
Before this, even when running the exact same package versions as on CI, some tests would fail locally. This is likely due to different BLAS backends being used on different Apple hardware generations. Moving data points away from decision boundaries hopefully reduces variance from softmax/matmul computations.


tests/test_consistency.py
```diff
@@ -248,10 +262,13 @@ def test__fit_predict__predictions_match_reference(
     reference_predictions = _load_reference_predictions_or_fail(test_case_name)
     predictions = _predict(test_case)
 
+    # We allow large relative tolerance because predictions can be close to 0.
```
bejaeger, line 265: why not increase atol for that reason?

bejaeger (approved):
Thanks!

tests/test_consistency.py
priorphil: because atol is fine for small values. The issue here is for values like `expected = 0.0001; actual = 0.0009` absolute diff is < 0.001 while rtol is 0.1.

PR #1072, opened 2026-06-25
CI: parallelise CPU test runs with pytest-xdist (~2.3x faster)

LeoGrin:
## What

Run the CPU test jobs under `pytest-xdist` instead of serially.

- Add `-n auto --dist worksteal` to the pytest steps in `test_compatibility` and `test_ubuntu_latest_314`.
- Pin `OMP_NUM_THREADS=MKL_NUM_THREADS=OPENBLAS_NUM_THREADS=1` in those two jobs.
- Move `pytest-xdist` from the `dev` group into the `ci` group (CI installs `--group ci`).

The GPU job (`test_gpu`) and the first-party-packages job are intentionally left as-is.

## Why

The suite is ~1100+ independent tests and currently runs serially. It parallelises cleanly.

Measured on a 4-core node matched to the CI runner (threads pinned, model cache on local disk), on the `-m "not slow"` suite, same node back-to-back:

| Config | time | speedup |
|---|---|---|
| serial | 408s | 1.0x |
| `-n2 --dist worksteal` (OMP=2) | 198s | 2.1x |
| **`-n auto`=`-n4 --dist worksteal` (OMP=1)** | **180s** | **2.3x** |

All **1178 tests pass** in every configuration (0 failures under xdist).

## Why thread pinning is mandatory

Without `*_NUM_THREADS=1`, each xdist worker spawns its own multi-threaded torch; on a 4-core runner the 4 workers then fight over the cores and the run is *slower* than serial. One thread per worker + one worker per core is the fast configuration.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI workflow and dev/ci dependency grouping; application runtime and GPU jobs are untouched.
> 
> **Overview**
> Speeds up CPU CI by running pytest under **pytest-xdist** (`-n auto --dist worksteal`) instead of a single serial process.
> 
> In `test_compatibility`, Linux/Windows (and lowest-direct MPS-disabled matrix legs) use xdist with `OMP_NUM_THREADS`, `MKL_NUM_THREADS`, and `OPENBLAS_NUM_THREADS` set to **1** so each worker does not oversubscribe cores. **macOS stays serial** because parallel MPS workers exhaust shared GPU memory. The `test_ubuntu_latest_314` job always parallelises with the same thread env at job level. **GPU** and **first-party** test jobs are unchanged.
> 
> `pytest-xdist` moves from the `dev` dependency group into **`ci`** so `uv sync --group ci` installs it. A towncrier **changed** note documents faster parallel CPU test suites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6467f9c3d112e69f4faacb867c531ba7aead4ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,13 +63,6 @@ jobs:
     env:
       HF_TOKEN: ${{ secrets.HF_TOKEN }}
       TABPFN_TOKEN: ${{ secrets.TABPFN_TOKEN }}
-      # The test runs use pytest-xdist (`-n auto`), so each worker gets one core.
-      # Pin BLAS/torch to a single thread per process; otherwise each worker spawns
-      # its own multi-threaded torch and they oversubscribe the cores, which is
-      # dramatically *slower* than serial.
-      OMP_NUM_THREADS: 1
-      MKL_NUM_THREADS: 1
-      OPENBLAS_NUM_THREADS: 1
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4.3.1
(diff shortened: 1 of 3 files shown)


LeoGrin:
Closing for now. CPU-test parallelisation works on Linux/Windows (~2.3x), but macOS can't be parallelised without dropping MPS coverage (the CI MPS GPU is virtualised and can't host concurrent xdist contexts), and macOS-serial gates the matrix wall-clock — so the net win here is runner-cost, not wall-clock. Parking until we decide whether to drop CI MPS coverage. Branch `ci/parallel-cpu-tests-pr` is kept.

PR #1073, opened 2026-06-26
Pin macos-15 runner, as tests fail on macos-26.

oscarkey:
GitHub is gradually switching macos-latest to macos-26, which makes the tests flake. Just pin to macos-15 for now, while we investigate.

Also: Don't request x64 Python, as we want to test on arm without Rosetta (this is what most people will have). It also doesn't seem to be having any affect in recent runs anyway, e.g. see "Install uv" here: https://github.com/PriorLabs/TabPFN/actions/runs/28105021796/job/83216827365

Part of PRI-331

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,16 +44,17 @@ jobs:
           - os: ubuntu-latest
             python-version: "3.10"
             dependency-set: lowest-direct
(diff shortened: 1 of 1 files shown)


arthur-priorlabs (approved):
Thanks!

PR #1076, opened 2026-06-26
fix: guard TorchTruncatedSVD against non-finite input and SVD non-convergence

### LINKED ISSUE #1044: [Bug] `torch.linalg.svd` fails to converge in `TorchTruncatedSVD` on degenerate data
## Description
When running TabPFN's preprocessing on datasets with degenerate row-subsamples (e.g., near-constant columns at small row caps), the PyTorch SVD computation can fail to converge. This results in a `RuntimeError` (typically `svd_compute: the updating process of SBDSDC did not converge`) originating from `TorchTruncatedSVD.fit()`, which crashes the pipeline. 

## Where it occurs
The issue is located in `src/tabpfn/preprocessing/torch/torch_svd.py` within `TorchTruncatedSVD.fit()`. Currently, the calls to `torch.linalg.svd` and `torch.svd_lowrank` are executed without any fallback mechanisms for numerical instability:

A robust internal fix would be to wrap the SVD computation in a try-except block and implement a fallback mechanism. Standard PyTorch workarounds for this include:

Fallback to driver="gesvd": The gesvd driver is more numerically stable than the default gesdd for degenerate matrices.

Fallback to float64: Upcasting the matrix to torch.float64 for the SVD computation and downcasting the results.

Graceful degradation: If both fail, return a zeroed projection or disable the SVD features for that specific batch.

_No response_

### Additional context

_No response_

### Impact

None

devangpratap:
## Issue

Fixes PriorLabs/TabPFN#1044

## Motivation and Context

`TorchTruncatedSVD.fit` replaced only NaN values with 0 before computing the SVD, leaving any +/-inf in place. Those infinities flow straight into `torch.linalg.svd` / `torch.svd_lowrank`, which then either raise a non-convergence `LinAlgError` (crashing the preprocessing pipeline) or silently return non-finite components, depending on the LAPACK backend. This shows up on datasets with degenerate row-subsamples (for example near-constant columns at small row caps), as reported in PriorLabs/TabPFN#1044.

This masks all non-finite values (NaN and +/-inf) to 0 before the decomposition, matching how `TorchSafeStandardScaler` already treats inf. It also adds a numerical-stability fallback: if the single-precision SVD fails to converge, it retries the exact decomposition in float64; the randomized (lowrank) path falls back to the exact decomposition on the same error.

---

## Public API Changes

- [X] No Public API changes

---

## How Has This Been Tested?

Added two regression tests:

* `test__fit__non_finite_input_does_not_crash` feeds +inf, -inf, and NaN into `fit` and asserts the returned components and singular values are finite.
* `test__exact_svd__retries_in_float64_on_non_convergence` forces the single-precision SVD to raise `LinAlgError` and asserts the float64 fallback runs and round-trips the dtype.

Ran the `test_torch_svd.py` suite locally (passes), plus pre-commit (ruff, ruff-format, mypy) clean.

---

## Checklist

- [X] Documentation has been updated (if the public API or usage changes). No public API or usage change.

### DIFF
--- a/changelog/1076.fixed.md
+++ b/changelog/1076.fixed.md
@@ -1,0 +1,1 @@
+Fix crash in `TorchTruncatedSVD` on non-finite input by masking NaN and +/-inf before the SVD, and retry in float64 when the decomposition fails to converge on degenerate data.
--- a/src/tabpfn/preprocessing/torch/torch_svd.py
+++ b/src/tabpfn/preprocessing/torch/torch_svd.py
@@ -23,6 +23,8 @@ def _exact_svd(
     try:
         u, s, vh = torch.linalg.svd(x, full_matrices=False)
     except torch.linalg.LinAlgError:
+        if x.dtype == torch.float64:
+            raise
         u, s, vh = torch.linalg.svd(x.double(), full_matrices=False)
         u, s, vh = u.to(x.dtype), s.to(x.dtype), vh.to(x.dtype)
     return u[:, :n_components], s[:n_components], vh[:n_components, :]
--- a/src/tabpfn/preprocessing/torch/torch_svd.py
+++ b/src/tabpfn/preprocessing/torch/torch_svd.py
@@ -7,6 +7,27 @@ from __future__ import annotations
 import torch
 
 
+def _exact_svd(
+    x: torch.Tensor,
+    n_components: int,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Exact truncated SVD, retrying in float64 if it fails to converge.
+
+    On degenerate (near rank-deficient) data the default LAPACK/cuSOLVER
+    ``gesdd`` driver can raise a non-convergence ``LinAlgError``. Recomputing
+    in double precision is more numerically stable and succeeds where the
+    single-precision decomposition does not.
+
+    Returns the components truncated to the top ``n_components``.
+    """
+    try:
+        u, s, vh = torch.linalg.svd(x, full_matrices=False)
+    except torch.linalg.LinAlgError:
+        u, s, vh = torch.linalg.svd(x.double(), full_matrices=False)
+        u, s, vh = u.to(x.dtype), s.to(x.dtype), vh.to(x.dtype)
+    return u[:, :n_components], s[:n_components], vh[:n_components, :]
+
+
 def _svd_flip_stable(
     u: torch.Tensor,
     v: torch.Tensor,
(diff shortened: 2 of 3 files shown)


alanprior:
@devangpratap thank you for your contribution! Seems useful! We will review this and get back to you!

alanprior:
Hello @devangpratap, we iterated on this internally, and we believe that the current code covers this problem already. Were you able to replicate the issue reported? Could you provide a working script that fails and motivated your fix?

Thanks!

devangpratap:
Went back through this properly. Script that shows the actual behavior on main:

```python
import torch
from tabpfn.preprocessing.torch.torch_svd import TorchTruncatedSVD

# small matrix, exact svd path: no crash, but silently returns NaN
x = torch.randn(8, 5)
x[0, 0] = float("inf")
out = TorchTruncatedSVD(n_components=4).fit(x)
print(torch.isnan(out["singular_values"]).any())  # True

# large matrix, svd_lowrank path: raises
x2 = torch.randn(5000, 300)
x2[0, 0] = float("inf")
TorchTruncatedSVD(n_components=128).fit(x2)
# _LinAlgError: input matrix is ill-conditioned or has too many repeated singular values
```

But you're right that this doesn't matter in practice. Through the actual pipeline it's dead code: TorchSafeStandardScaler already sanitizes inf before it reaches the SVD (steps.py:414-423, torch_svd.py:340-341). Checked it by running the pre-fix torch_svd.py with an inf value through TorchAddSVDFeaturesStep end to end, no crash, no NaNs. So the isfinite() guard isn't fixing anything reachable today, only relevant if TorchTruncatedSVD is called directly, which nothing in the codebase does.

The other piece, the float64 retry when SVD fails to converge, is still worth keeping. Tried reproducing the SBDSDC error from #1044 on finite degenerate data (near-constant columns, rank-1, all-zeros, extreme scale) and couldn't trigger it either, seems LAPACK/hardware dependent rather than something a script can force.

I'll cut the isfinite() masking and keep just the float64 retry, since that one's a no-op on the normal path and covers the actual failure mode in #1044.

devangpratap:
Closing this. Neither half of the fix protects against something reachable.

isfinite() guard: dead code. TorchSafeStandardScaler always sanitizes input before it reaches the SVD, verified end to end. Even tried overflowing the scaler's own mean/std to inf, output stayed finite.

float64 retry: couldn't reproduce the SBDSDC error from #1044. ~31k trials across degenerate matrix shapes, macOS and Linux, single and multi-threaded. Zero failures.

Reopen if the reporter comes back with a concrete traceback.

PR #1077, opened 2026-06-27
Document OS and CPU/GPU install guidance

kiwoongyoon:
Title: Document OS and CPU/GPU install guidance

## Summary

- Add supported environment notes to the README installation section.
- Document the Windows/WSL2 caveat for GPU workloads.
- Point users to the PyTorch installer selector when they want a smaller CPU-only PyTorch build or a CUDA-specific build before installing TabPFN.

## Test evidence

- `python - <<'PY' ... PY` passed and confirmed the README contains the expected OS and CPU/GPU installation guidance.
- Reviewed `git diff -- README.md`; the patch is documentation-only.

## Risks or notes for maintainers

- The OS wording is conservative and says TabPFN is primarily tested on Linux and macOS. Please adjust if native Windows should be documented as fully supported.
- The docs link to the PyTorch selector instead of hard-coding CPU/CUDA install URLs, so users get current commands from PyTorch.


priorphil:
Hey, I'm OOO for 2 weeks with no laptop access, could you reassign to open
source stewards to allow for auto-assign? Thanks!

README.md
```diff
@@ -25,9 +25,30 @@
 pip install tabpfn
 ```
 
-Note: For best performance on Apple Silicon/MPS, consider installing a pytorch version after
-the nightly "2.13.0.dev20260510". This enables flash attention without relying on MLX
-(the latter requires a GPU-CPU-GPU roundtrip).
+TabPFN supports Python 3.10+ and is primarily tested on Linux and macOS. Windows
+users should prefer WSL2 for GPU workloads because PyTorch CUDA support is
```
oscarkey: > Windows users should prefer WSL2

do you happen to have a page from the pytorch docs we could link to here, to give some context?

README.md
```diff
@@ -25,9 +25,30 @@
...
+version, install PyTorch first with the command from the
+[PyTorch installation selector](https://pytorch.org/get-started/locally/), then
+install TabPFN:
+
+```bash
+# Example: CPU-only PyTorch. For CUDA, use the command generated by the
+# PyTorch installation selector for your platform and CUDA version.
+pip install torch --index-url https://download.pytorch.org/whl/cpu
+# Example: Install PyTorch first (get your specific command from the selector)
+pip install torch --index-url https://download.pytorch.org/whl/cpu
```
oscarkey:
```suggestion
```
I suggest we only need the first example.

README.md
```diff
@@ -25,9 +25,30 @@
 pip install tabpfn
 ```
 
-Note: For best performance on Apple Silicon/MPS, consider installing a pytorch version after
-the nightly "2.13.0.dev20260510". This enables flash attention without relying on MLX
-(the latter requires a GPU-CPU-GPU roundtrip).
+TabPFN supports Python 3.10+ and is primarily tested on Linux and macOS. Windows
```
oscarkey: > primarily tested on Linux and macOS

We run our full CI suite on Windows, so I don't think we need to say this. Is there a particular issue you've been facing on windows?

oscarkey (changes requested):
thanks for your contribution! just a few of comments/questions.

kiwoongyoon:
Thanks for the review. I pushed a small follow-up commit that removes the Windows caveat wording, keeps the install guidance to a single simple example, and points users to the PyTorch installation selector for CPU/CUDA/platform-specific commands including Windows and WSL notes.

oscarkey (approved):
thanks!

PR #1078, opened 2026-06-27
Debug failing macos-26 runner

oscarkey:
## Summary

Throwaway debug PR to diagnose why MPS numerics break on GitHub's `macos-26-arm64` runner image (see [PRI-331](https://linear.app/priorlabs/issue/PRI-331/ci-failing-when-runner-is-macos-26)). Not for merge — delete once we have the answer.

Adds a single new workflow `.github/workflows/debug-macos-26.yml` that runs on `workflow_dispatch` and on PRs that touch the workflow file itself, so iteration here doesn't pollute other PRs' CI.

The workflow runs three jobs in parallel:
- `macos-26` baseline (reproduces the failure)
- `macos-26` with `PYTORCH_ENABLE_MPS_FALLBACK=1` (tests whether some op silently dispatches an unsupported MPS kernel)
- `macos-15` (known-good control)

Each job:
1. Dumps `sw_vers`, `uname`, full `SPHardwareDataType` and `SPDisplaysDataType` so we can see the exact OS build / chip / Metal driver. The earlier-investigated failing run was on `macos-26-arm64` image `20260623.0192`, build `25E246` — but we don't yet know the chip generation; this prints it.
2. Reports torch version + MPS availability with `PYTORCH_ENABLE_MPS_FALLBACK` echoed back.
3. Runs a graded **MPS-vs-CPU correctness ladder** (elementwise → reductions → matmul → layernorm → SDPA → dtype sweep → TabPFN-scale shapes) and prints `max_abs_diff` per op. The first op whose diff jumps from ~1e-7 to ~1e-2+ is the broken kernel. Each row is flagged automatically.
4. Reproduces two known-failing tests as ground truth: `test__batched_inference__matches_single_dataset[mps]` and `test__fit_and_predict__on_demo_dataset__accuracy_reasonable[v2]`.

## What we expect to learn

- **Hardware/OS exact**: confirms whether the runner is M1-class (older Apple Silicon vs the M4 where we couldn't reproduce), and whether OS is on 26.4 (the build we saw in logs) vs 26.5+.
- **Which kernel is wrong**: the ladder localises the failure to a specific op + dtype + shape regime, instead of inferring it from "accuracy collapsed to 0.33".
- **Whether `PYTORCH_ENABLE_MPS_FALLBACK=1` masks it**: a one-line workaround if so.
- **Whether `macos-15` is clean on the same ladder**: confirms (or refutes) that the bug is OS-version-specific.

## Test plan

- [ ] Trigger via `workflow_dispatch` (or push another commit to this PR)
- [ ] Read the diagnostic output for both `macos-26` variants and the `macos-15` control
- [ ] Decide next experiment (likely: try torch nightly, or file an upstream PyTorch issue with the ladder output)
- [ ] Close & delete branch once we have the answer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,10 @@ env:
 jobs:
   check_python_linting:
     name: Ruff Linting & Formatting
+    # Debug PR ok/debug-macos-26 only runs its own debug-macos-26.yml
+    # workflow; full CI is irrelevant there. Skipping the lint gate
+    # cascades to every downstream job via their `needs:` dependencies.
+    if: github.head_ref != 'ok/debug-macos-26'
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
--- a/.github/workflows/debug-macos-26.yml
+++ b/.github/workflows/debug-macos-26.yml
@@ -63,6 +63,11 @@ jobs:
       - name: Install dependencies (highest resolution, matches CI failure)
         run: uv sync --group ci --resolution highest
 
+      - name: Override torch with nightly (test upstream MPS fix)
+        run: |
+          uv pip install --pre --upgrade torch \
+            --index-url https://download.pytorch.org/whl/nightly/cpu
+
       - name: torch + MPS environment
         run: |
           uv run --no-sync python - <<'PY'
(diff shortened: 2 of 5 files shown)


oscarkey:
Sorry Tuana this should have been in draft from the start. Pls ignore!

PR #1083, opened 2026-07-02
Fix fp16 inference in InferenceEngineBatchedNoPreprocessing

eliott-kalfon:
The batched (no-preprocessing) inference engine cast only the inputs to `force_inference_dtype`, leaving the model in fp32. With inference_precision=torch.float16 this made an fp16 activation meet an fp32 weight in the feature-embedding Linear, raising `'mat1 and mat2 must have the same dtype, but got Half and Float'`. `predict_proba_batched` hit this on GPU; standard `predict_proba` was fine because its engines cast the model via `ModelCache.set_dtype`.

Cast the model to `force_inference_dtype` alongside the inputs in iter_outputs. Add a GPU-only regression test that runs `predict_proba_batched` at fp16 and checks it matches the fp32 result.

## Issue
https://linear.app/priorlabs/issue/RES-1979/fix-predict-proba-batched-crash-with-torchfloat16

## Motivation and Context
We would need this to work for performance optimisation
---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
Locally

---

## Checklist

---

### DIFF
--- a/tests/test_classifier_interface.py
+++ b/tests/test_classifier_interface.py
@@ -1434,6 +1434,49 @@ def test__predict_proba_batched__matches_per_dataset(device: str) -> None:
         np.testing.assert_allclose(proba[i], ref, atol=atol)
 
 
+def test__predict_proba_batched__fp16_matches_fp32() -> None:
+    """predict_proba_batched works with inference_precision=float16 (regression).
+
+    The batched engine used to cast only the inputs to the forced dtype, leaving
+    the model in fp32, so an fp16 forward raised "mat1 and mat2 must have the same
+    dtype, but got Half and Float". Standard predict_proba fp16 was unaffected.
+    fp16 matmul is only reliably supported on CUDA, so this is GPU-only.
+    """
+    if not torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


tests/test_classifier_interface.py
```diff
@@ -1434,6 +1434,49 @@ def mkds(seed: int, n: int = 60, f: int = 5) -> tuple[np.ndarray, np.ndarray]:
...
 
+def test__predict_proba_batched__fp16_matches_fp32() -> None:
+    """predict_proba_batched works with inference_precision=float16 (regression).
+
+    The batched engine used to cast only the inputs to the forced dtype, leaving
+    the model in fp32, so an fp16 forward raised "mat1 and mat2 must have the same
+    dtype, but got Half and Float". Standard predict_proba fp16 was unaffected.
+    fp16 matmul is only reliably supported on CUDA, so this is GPU-only.
+    """
+    if not torch.cuda.is_available():
```
ggprior: Is it actually needed to gate on CUDA? I seem to get the same behavior on CPU on torch 2.12

ggprior (approved):
LGTM

PR #1085, opened 2026-07-03
feat: advertise inf as MCAR indicator

arthur-priorlabs:
Document how TabPFN handles infinite values when detecting them during either `fit` or `predict`.
See RES-1537 and #1055.


LeoGrin (approved):
LGTM!

Nit: this would also trigger on y=inf and on data too large for the dtype right? in which case not sure the hint would apply? (I think we would fail on y=inf even with the inference config change?)

arthur-priorlabs:
You're right, we do fail on `y=inf` even with passthrough enabled (as we should); fixed

PR #1087, opened 2026-07-03
add cache calculator

### LINKED ISSUE #1086: add kv cache calculator

Want to allow to query for the cache size without having to actually run the model. This helps theoretical analysis and also practically to prevent OOM errors.

Add an API to estimate the size of the cache given a model.

_No response_

### Additional context

_No response_

### Impact

None

jmkuebler:
## Issue

This fixes issue PriorLabs/TabPFN#1086 <br>Fixes PriorLabs/TabPFN#1086

## Motivation and Context

When analyzing the size of the Cache it is good to have a theoretical model to query without having to really build the cache

---

## Public API Changes

- [X] No Public API changes

---

## How Has This Been Tested?

Added unit tests that compare the estimate with the actual cache size. Also for TabPFNv3 we test actual numbers we computed ourselves.

---

## Checklist

---

### DIFF
--- a/src/tabpfn/architectures/kv_cache.py
+++ b/src/tabpfn/architectures/kv_cache.py
@@ -19,9 +19,7 @@ from dataclasses import dataclass, field
 import torch
 from torch import Tensor
 
-# Default integer dtype the KV cache is quantized to. Single source of truth
-# shared by :meth:`KVCacheEntry.quantize` and any cache-size estimation, so the
-# stored dtype is never hardcoded in two places.
+# only supported KV quantization so far
 QUANTIZED_KV_DTYPE: torch.dtype = torch.int8
 
 # Low, high, max-magnitude value for each dtype.
@@ -89,7 +87,7 @@ class KVCacheEntry:
         """Quantize this entry with per-tensor symmetric scaling.
 
         Args:
-            dtype: Target integer dtype (default :data:`QUANTIZED_KV_DTYPE`).
+            dtype: Target integer dtype (default `QUANTIZED_KV_DTYPE`).
         """
         assert self.is_valid()
         k_q, k_s = _quantize_tensor(self.key, dtype)
--- a/src/tabpfn/architectures/kv_cache.py
+++ b/src/tabpfn/architectures/kv_cache.py
@@ -19,6 +19,11 @@ from dataclasses import dataclass, field
 import torch
 from torch import Tensor
 
+# Default integer dtype the KV cache is quantized to. Single source of truth
+# shared by :meth:`KVCacheEntry.quantize` and any cache-size estimation, so the
+# stored dtype is never hardcoded in two places.
+QUANTIZED_KV_DTYPE: torch.dtype = torch.int8
+
 # Low, high, max-magnitude value for each dtype.
 # int8 uses the symmetric range [-127, 127] (one code below the full int8
 # range) so that ``-max * scale`` equals ``+max * scale`` and dequantization
@@ -78,11 +83,13 @@ class KVCacheEntry:
             return KVCacheEntry()
         return KVCacheEntry(key=self.key.to(device), value=self.value.to(device))
 
-    def quantize(self, dtype: torch.dtype = torch.int8) -> QuantizedKVCacheEntry:
+    def quantize(
+        self, dtype: torch.dtype = QUANTIZED_KV_DTYPE
+    ) -> QuantizedKVCacheEntry:
         """Quantize this entry with per-tensor symmetric scaling.
 
         Args:
-            dtype: Target integer dtype (default ``torch.int8``).
+            dtype: Target integer dtype (default :data:`QUANTIZED_KV_DTYPE`).
         """
         assert self.is_valid()
         k_q, k_s = _quantize_tensor(self.key, dtype)
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -260,7 +260,7 @@ def get_cache_size(
     X_train: Any,
     *,
     model_config: TabPFNV3Config,
-    dtype: torch.dtype | Literal["autocast"],
+    base_dtype: torch.dtype | Literal["autocast"],
     quantize_kv_cache: bool = True,
 ) -> int:
     """Calculate the cached memory in bytes for a single TabPFN v3 estimator.
@@ -299,7 +299,7 @@ def get_cache_size(
             end-to-end runs preprocessing may change it (SVD features, categorical
             expansion, per-member subsampling), making those terms approximate.
         model_config: The v3 architecture config.
-        dtype: Either a ``torch.dtype`` (forced-precision path -- every term is
+        base_dtype: Either a ``torch.dtype`` (forced-precision path -- every term is
             sized at this dtype; on CPU this is usually Float32) or the string
             ``"autocast"`` (GPU autocast path -- KV and ``train_embeddings`` are
             sized at fp16 while ``inducing_hidden`` and ``scaler_cache`` stay
@@ -317,7 +317,7 @@ def get_cache_size(
     # Set the stored dtype of each cached component up front. On the forced-
     # precision path the model and inputs are cast to ``dtype``, so every
     # component is ``dtype``. Under autocast the cache is mixed precision.
-    if dtype == "autocast":
+    if base_dtype == "autocast":
         # Autocast keeps fp32 weights and casts ops to fp16 at runtime, so KV and
         # train_embeddings (matmul outputs) are fp16, while inducing_hidden and
         # the scaler (fp32 reduction/norm ops, scaler fit on the fp32 input) stay
@@ -328,11 +328,11 @@ def get_cache_size(
         inducing_dtype = torch.float32
         scaler_dtype = torch.float32
     else:
-        kv_dtype = QUANTIZED_KV_DTYPE if quantize_kv_cache else dtype
-        kv_scale_dtype = dtype
-        train_emb_dtype = dtype
-        inducing_dtype = dtype
-        scaler_dtype = dtype
+        kv_dtype = QUANTIZED_KV_DTYPE if quantize_kv_cache else base_dtype
+        kv_scale_dtype = base_dtype
+        train_emb_dtype = base_dtype
+        inducing_dtype = base_dtype
+        scaler_dtype = base_dtype
 
     # Accept either an array-like (read ``.shape``) or a (n_rows, n_features) tuple.
     if hasattr(X_train, "shape"):
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -45,7 +45,12 @@ from tabpfn.architectures.interface import (
     ArchitectureConfig,
     PerformanceOptions,
 )
-from tabpfn.architectures.kv_cache import KVCache, KVCacheEntry, QuantizedKVCacheEntry
+from tabpfn.architectures.kv_cache import (
+    QUANTIZED_KV_DTYPE,
+    KVCache,
+    KVCacheEntry,
+    QuantizedKVCacheEntry,
+)
 from tabpfn.architectures.shared.chunked_evaluate import chunked_evaluate_maybe_inplace
 from tabpfn.architectures.shared.scaled_dot_product_attention import (
     scaled_dot_product_attention,
@@ -229,14 +234,14 @@ class TabPFNV3Cache(KVCache):
             inducing_hidden=self._list_of_tensors_to(self.inducing_hidden, device),
         )
 
-    def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
+    def quantize(self, dtype: torch.dtype = QUANTIZED_KV_DTYPE) -> TabPFNV3Cache:
         """Return a new cache with quantized ICL KV entries.
 
         Only the ICL KV cache is quantized; ``train_embeddings``,
         ``scaler_cache``, and ``inducing_hidden`` stay in full precision.
 
         Args:
-            dtype: Target integer dtype (default ``torch.int8``).
+            dtype: Target integer dtype (default :data:`QUANTIZED_KV_DTYPE`).
         """
         quantized_kv = {
             idx: (entry.quantize(dtype) if isinstance(entry, KVCacheEntry) else entry)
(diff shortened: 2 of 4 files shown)


src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,90 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+       them. Controlled by ``include_decoder_activations`` (on by default).
+
+    The KV cache and the activations are stored at *different* dtypes: the
+    default engine int8-quantizes the KV cache (``quantize_kv_cache``) but leaves
+    ``train_embeddings`` at the compute dtype. Each term is therefore sized with
+    its own dtype rather than a single shared one. The ``scaler_cache`` /
+    ``inducing_hidden`` entries are small and not counted.
+
+    Args:
+        X_train: Training features; only ``X_train.shape[0]`` (the row count)
```
ggprior: expected only the shape would be relevant here, even better if just the row count -- no need to pass a tensor in that case, perhaps we'd just accept `n_rows` here or similar
(sorry my initial function definition was bad)

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,90 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+    *,
+    model_config: TabPFNV3Config,
+    ensemble_size: int,
+    dtype: torch.dtype,
+    quantize_kv_cache: bool = True,
+    include_decoder_activations: bool = True,
+) -> int:
+    """Estimate the cached memory in bytes for a TabPFN v3 inference run.
+
+    Analytical counterpart of the (removed) ``TabPFNV3Cache.cache_size_mb``:
```
ggprior: minor: this is reminiscing about a past state of the code base, which tends to get confusing over time

arthur-priorlabs (approved):
Looks good, although this doesn't seem to be used anywhere in the codebase.

Maybe add some kind of early warning for `fit_with_cache` when available memory is too low to handle it?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,90 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
         )
 
 
+def estimate_cache_size(
```
ggprior: Can we say more here than `estimate_`? What are the conditions for this to be exact? If it's an estimation, how far off can we expect this to be from the true value?
(note: Ideal for the engineering use case would be an upper bound)

ggprior: In fact we may be underselling ourselves here as the tests look like exact assertions!

ggprior (commented):
This will be a useful feature, thanks for working on it! Some comments below, refer to @bejaeger for approvals

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: That is a good call -- at least for v3. But potentially future models could depend on columns again. Also for a user its maybe more naive to compute the cache using your dataset as input whether some derivative of it.

Just some thoughts, happy to change to number of rows if you think that's better, and so far the method is anyways for v3 specifically.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: yes the kv and final embeddings is accurate. But so far we are excluding the inducing points. So it is a slight "underestimate". But the inducing points don't scale with the number of rows.

Maybe it is best to include them as well and then it is an accurate "computation" of everything that is cached.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: I updated it and the method now also needs the column count, as it includes also the inducing points now. 

I overloaded the X_train argument so we can also just pass a tuple of `rows, columns`.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: I added all the other terms even if they are very small and refactored the method to `calculate`

src/tabpfn/architectures/tabpfn_v3.py
ggprior: > But potentially future models could depend on columns again

good call, current approach of overloading also works 👍

src/tabpfn/architectures/tabpfn_v3.py
ggprior: sounds good!

src/tabpfn/architectures/kv_cache.py
```diff
@@ -19,6 +19,11 @@
 import torch
 from torch import Tensor
 
+# Default integer dtype the KV cache is quantized to. Single source of truth
+# shared by :meth:`KVCacheEntry.quantize` and any cache-size estimation, so the
+# stored dtype is never hardcoded in two places.
```
bejaeger: nit: comment seems unnecessarily long

src/tabpfn/architectures/kv_cache.py
```diff
@@ -78,11 +83,13 @@ def to(self, device: torch.device | str) -> KVCacheEntry:
...
 
-    def quantize(self, dtype: torch.dtype = torch.int8) -> QuantizedKVCacheEntry:
+    def quantize(
+        self, dtype: torch.dtype = QUANTIZED_KV_DTYPE
+    ) -> QuantizedKVCacheEntry:
         """Quantize this entry with per-tensor symmetric scaling.
 
         Args:
-            dtype: Target integer dtype (default ``torch.int8``).
+            dtype: Target integer dtype (default :data:`QUANTIZED_KV_DTYPE`).
```
bejaeger: nit: Let's remove the `:data:` or `:method:` hints since we're not planning to generate docs from them.

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+       compress the train rows into a fixed set) and instead scales with the
+       column count, so it can dominate on wide, small-``N`` datasets. Controlled
+       by ``include_inducing_points`` (on by default).
+
+    4. The fitted scaler stats ``scaler_cache`` (``mean`` + ``std``), of shape
+       ``(2, n_features)`` per estimator. Always counted (small).
+
+    Each term is sized with its own dtype: the default engine int8-quantizes the
+    KV cache (``quantize_kv_cache``) but leaves everything else at the compute
+    ``dtype``.
```
bejaeger: the doc string seems overly long for a method that can be easily be read in code. Could we shorten and potentially move some of the high-level comments to the code.

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
         )
 
 
+def calculate_cache_size(
```
bejaeger: nit

```suggestion
def get_cache_size(
```

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+         (GQA), else ``icl_num_heads`` (plain MHA).
+
+       When ``quantize_kv_cache`` is True the K/V are int8 plus one scalar scale
+       per key and per value tensor (``nlayers * 2`` scalars at ``dtype``).
+
+    2. The last-layer train activations ``train_embeddings`` of shape
+       ``(N_train, embed_dim * feat_agg_num_cls_tokens)`` per estimator -- **only
+       for classification** (binary or multiclass), where the ``ManyClassDecoder``
+       attends test queries to them as retrieval keys. Regression never uses
+       them. Controlled by ``include_decoder_activations`` (on by default).
```
bejaeger: Currently, we're actually a bit wasteful and also include these activations for the cache for regression.
Also, I don't like the API here because the user has to think "do I need to include it or not". Could we just parse the task_type instead?
For now though, I think we can simply remove it until we change the cache storage for regression.

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
 
+def calculate_cache_size(
+    X_train: Any,
+    *,
+    model_config: TabPFNV3Config,
+    ensemble_size: int,
+    dtype: torch.dtype,
+    quantize_kv_cache: bool = True,
+    include_decoder_activations: bool = True,
+    include_inducing_points: bool = True,
```
bejaeger: What use case do you intend here?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
         )
 
 
+def calculate_cache_size(
+    X_train: Any,
+    *,
+    model_config: TabPFNV3Config,
+    ensemble_size: int,
+    dtype: torch.dtype,
+    quantize_kv_cache: bool = True,
```
bejaeger, line 265: should we make this more explicit and have `kv_cache_dtype` directly? Maybe also change `dtype` to `base_dtype` in that case? 
By moving the responsibility to the function caller we don't have to change this fn when we change the quantization.

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
         )
 
 
+def calculate_cache_size(
+    X_train: Any,
+    *,
+    model_config: TabPFNV3Config,
+    ensemble_size: int,
```
bejaeger: nit: I'm not a fan of having this here, as the model itself shouldn't really care about the ensemble size. Could we remove and move the responsibility to the caller? WDYT?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+    total_bytes = kv_elements * kv_dtype.itemsize
+    if quantize_kv_cache:
+        # One scalar scale per key and per value tensor, stored at ``dtype``.
+        total_bytes += model_config.nlayers * 2 * dtype.itemsize
+
+    # Classification (binary or multiclass) runs the ManyClassDecoder, which
+    # attends to the cached train activations as retrieval keys; regression
+    # (max_num_classes == 0) does not. classification == "has any classes".
+    # These activations are not quantized with the KV cache -> sized at dtype.
+    is_classification = model_config.max_num_classes > 0
```
bejaeger: another advantage for passing the task type would be that we don't need this logic here.

In fact, in the model we have
`task_type = "multiclass" if config.max_num_classes >= 2 else "regression"`.

would be dangerous if this is different here.

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+        inducing_elements = (
+            model_config.dist_embed_num_blocks
+            * n_features
+            * model_config.dist_embed_num_inducing_points
+            * model_config.embed_dim
+        )
+        total_bytes += inducing_elements * dtype.itemsize
+
+    # Fitted scaler stats: mean + std, each (n_features,). Always cached.
+    total_bytes += 2 * n_features * dtype.itemsize
```
bejaeger: nit: super minor point but are you sure the scaler stats are stored in 16-bit?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -251,6 +256,133 @@ def quantize(self, dtype: torch.dtype = torch.int8) -> TabPFNV3Cache:
...
+            ``inducing_hidden`` term (see point 3).
+
+    Returns:
+        Cache size in bytes. Divide by ``1024 ** 2`` for MB.
+    """
+    # Accept either an array-like (read ``.shape``) or a (n_rows, n_features) tuple.
+    if hasattr(X_train, "shape"):
+        n_train, n_features = int(X_train.shape[0]), int(X_train.shape[1])
+    else:
+        n_train, n_features = (int(v) for v in X_train)
```
bejaeger: nit: should we just pass "num_train_rows" and "num_features" instead?

bejaeger (changes requested):
Thanks a lot @jmkuebler ! I left a few comments on the design.

I was wondering if we should also hook this up to the main classifier/regressor interface? Maybe via the inference engines in `inference.py`. If a wrapper doesn't support a `kv_cache` it could return None. WDYT?

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: I updated to just calcluate the full cache always, as we have no foreseeable use in mind to leave parts away. As of writing, this works for regression and classification, so the calculation does not distinguish. If in the future we reduce the cache for regression, we have to adjust the calculation accordingly.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: grew a bit organically (I first started with KV, then added the otehrs). I now changed the interface to always sum up all the terms, and removed those arguments for simplicity.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: removed now entirely, as for the time being the calculation does not seperate classiication and regression

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: had the discussion w/ Georg above. I went for the overloading such that both work. But fine to move exlusively to shapes if you prefer that.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: It is actually not such a minor thing unfortunately. The dtype of the scale seems to depend on which mode the model runs in. If it is forced 16 bit, things might be different than when it is autocast or on CPU.

I need to take a deeper look here and also test on a GPU to make sure I get this right..

Putting the PR on draft so we don't accidentally merge until that is resolved.

jmkuebler: Ok fixed, I now handle things seperately depending whether we are on CPU with forced dtype or whether it is GPU using autocast. 🥳

jmkuebler: also tested on GPU ✅

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: >Maybe also change dtype to base_dtype in that case? 

✅
 
>should we make this more explicit and have kv_cache_dtype directly? 

Let's keep it as is for now. We only have one format for KV, but that implicitly also determines that scales are per-tensor. The exact format thus is not solely determined by the KV type, and if we add more flexibility there, we have to properly implement it including the scales.

jmkuebler:
@bejaeger thanks a ton for your comments. I addressed most comments, but left two to resolve for you.
In particular I now added the exact dtype support both for CPU and GPU. The autocast creates different precision in the cache, which we now handle correctly and test for.

ready from my side now.

jmkuebler:
@bejaeger 
>I was wondering if we should also hook this up to the main classifier/regressor interface? Maybe via the inference engines in inference.py. If a wrapper doesn't support a kv_cache it could return None. WDYT?

I am currently not sure about the pros and cons of this one. We could also land this PR and I take the addition to the other interfaces as follow-up and we discuss this briefly?

src/tabpfn/architectures/tabpfn_v3.py
bejaeger: hm...this adds unnecessary complexity IMO so I'd prefer the simpler interface instead. But not a strict blocker from my side, since this was also discussed with @ggprior .

bejaeger (approved):
Looks great! thanks for all the changes.

I agree that creating a proper API can be done in a follow-up.
I think we should also add this `fn` to the other model versions and allow getting sth like `get_cache_size("v2.5", ...)` similar to `get_architecture` in `model_loading.py`

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: Actually I jsut did it. If the need to pass the full data comes up again, we can still ad it.

PR #1094, opened 2026-07-07
Cold-start DDP finetuning: concurrent dataset downloads race across ranks. 

anuragg1209:
## Issue

## Motivation and Context

When running the finetuning examples with multiple GPUs (torchrun --nproc-per-node=N examples/finetune_classifier.py / finetune_regressor.py) for the first time on a machine with a cold sklearn dataset cache, all N ranks call the sklearn dataset fetcher simultaneously, which can crash the run.

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

- Added a public` tabpfn.finetuning.main_process_first()` context manager (in `finetuned_base.py`, reusing `_init_distributed_if_needed` so the NCCL backend/timeout stay defined in one place): rank 0 runs the block (downloads and warms the cache) while other ranks wait at a barrier, then read from the warm cache. Leaves the process group initialized for `fit()` to reuse; no-op in single-process runs.
- Both finetune examples now wrap their dataset fetch in with `main_process_first()`:.

Verified with a 2-rank torchrun test asserting rank 1 only enters the block after rank 0 completes it.

## Checklist

---

### DIFF
--- a/src/tabpfn/finetuning/finetuned_base.py
+++ b/src/tabpfn/finetuning/finetuned_base.py
@@ -149,7 +149,13 @@ def main_process_first() -> Iterator[None]:
         yield
         return
 
-    _init_distributed_if_needed("cuda")
+    using_ddp, _, _ = _init_distributed_if_needed("cuda")
+    if not using_ddp:
+        # WORLD_SIZE was set by something other than torchrun (no LOCAL_RANK),
+        # so there is no process group to coordinate through.
+        yield
+        return
+
     is_main_process = dist.get_rank() == 0
     if not is_main_process:
         dist.barrier()
--- a/src/tabpfn/finetuning/finetuned_base.py
+++ b/src/tabpfn/finetuning/finetuned_base.py
@@ -16,6 +16,7 @@ import os
 import time
 import warnings
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
@@ -55,6 +56,8 @@ from tabpfn.validation import ensure_compatible_fit_inputs_sklearn
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from tabpfn.constants import ModelVersion, XType, YType
 
 # Currently, we only support a batch size of 1 for finetuning.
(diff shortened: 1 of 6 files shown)


src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -128,6 +131,35 @@ def _maybe_setup_ddp(
...
+    leaves it initialized so that a subsequent ``fit()`` reuses it. Call
+    ``torch.distributed.destroy_process_group()`` at the end of your script.
+    No-op when running with a single process.
+    """
+    if int(os.environ.get("WORLD_SIZE", "1")) <= 1:
+        yield
+        return
+
+    _init_distributed_if_needed("cuda")
+    is_main_process = dist.get_rank() == 0
```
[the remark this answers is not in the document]
anuragg1209: Fixed.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -55,6 +56,8 @@
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from tabpfn.constants import ModelVersion, XType, YType
```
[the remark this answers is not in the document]
anuragg1209: Addressed.

src/tabpfn/finetuning/finetuned_base.py
```diff
@@ -128,6 +130,41 @@ def _maybe_setup_ddp(
...
+        return
+
+    using_ddp, _, _ = _init_distributed_if_needed("cuda")
+    if not using_ddp:
+        # WORLD_SIZE was set by something other than torchrun (no LOCAL_RANK),
+        # so there is no process group to coordinate through.
+        yield
+        return
+
+    is_main_process = dist.get_rank() == 0
```
psinger-prior, line 158: would we need to check for local rank here? although finetuning is mostly single node anyways

psinger-prior (approved):
I did not test it (which Im sure you did), but code lgtm

src/tabpfn/finetuning/finetuned_base.py
anuragg1209: Good catch. For the current setup, they're equivalent. Since multi-node isn't supported here yet, I'd keep the global-rank check for now; happy to change if you'd prefer to future-proof?

anuragg1209: Actually,  since multi-node isn't supported by `_maybe_setup_ddp` anyway, this parameter is dormant until that changes. It's future-proofing plus documentation of the assumption. I prefer to move with the current setup for now and merge the branch.

psinger-prior (approved):

Issue #1095, opened 2026-07-08
Almost equal assertion test failures on CUDA


jmkuebler:
### Describe the bug

on main 20 tests fail when running them on RTX 6000 Pro.
I believe it is related to https://github.com/PriorLabs/TabPFN/issues/800#issuecomment-4903444425 but needs a bit more investigation.

## Test run on `main` (cuda)

**Summary:** 20 failed, 1503 passed, 111 skipped, 12 xfailed, 4446 warnings in 319.44s (0:05:19)

- ✅ 1503 passed  
- ❌ 20 failed  

> All failures are in `tests/test_save_load_fitted_model.py` and are cross-device (`cpu-cuda` / `cuda-cpu`) `AssertionError`s — a pre-existing over-tight numerical-equivalence check across devices, unrelated to the KV-cache chunking change.

### Steps/Code to Reproduce

run `pytest tests` on a GPU

### Expected Results

all tests should pass on all hardwares

## ❌ Failing tests

<details>
<summary><code>tests/test_save_load_fitted_model.py::test__save_and_load_twice__predictions_equal_to_before_save[regression-cpu-cuda]</code></summary>

```text
task_type = 'regression', saving_device = 'cpu', loading_device = 'cuda'
tmp_path = PosixPath('/tmp/pytest-of-jonas_priorlabs_ai/pytest-5/test__save_and_load_twice__pre1')

    @pytest.mark.parametrize(
        ("task_type", "saving_device", "loading_device"),
        [
            pytest.param(task_type, saving_device, loading_device, marks=pytest.mark.slow)
            if "mps" in (saving_device, loading_device)
            else (task_type, saving_device, loading_device)
            for task_type in ["regression", "classification"]
            for (saving_device, loading_device) in device_pairs
        ],
    )
    def test__save_and_load_twice__predictions_equal_to_before_save(
        task_type: str,
        saving_device: str,
        loading_device: str,
        tmp_path: Path,
    ) -> None:
        if task_type == "regression":
            estimator_class = TabPFNRegressor
            X, y = _make_regression_data()
        elif task_type == "classification":
            estimator_class = TabPFNClassifier
            X, y = _make_classification_data_with_categoricals()
        else:
            raise ValueError
    
        original_model = estimator_class(device=saving_device, n_estimators=4)
        original_model.fit(X, y)
    
        path_1 = tmp_path / "model_1.tabpfn_fit"
        original_model.save_fit_state(path_1)
[...]

Issue #800, opened 2026-02-27
Variations in regression "mean" output predictions with variations in test set


klemens-floege:
### Describe the bug

Using the "mean" output type on my local MacOS hardware I am seeing small deviations in predictions depending on whether the predictions are batched or not. Example script below: 


### Steps/Code to Reproduce

```
import numpy as np
import sklearn.datasets
from tabpfn import TabPFNRegressor

X, y, _ = sklearn.datasets.make_regression(
    n_samples=9, n_features=3, random_state=0, coef=True
)

model = TabPFNRegressor(n_estimators=2, random_state=42, device="cpu")
model.fit(X, y)

# ---- Unbatched prediction ----
pred_full = model.predict(X, output_type="mean")

# ---- Single-sample batched prediction ----
preds_one = np.concatenate(
    [model.predict(X[i : i + 1], output_type="mean") for i in range(len(X))]
)

print("Full:     ", pred_full)
print("Batched:  ", preds_one)
print("Max |Δ|:  ", np.max(np.abs(pred_full - preds_one)))
print("Per-sample Δ:", pred_full - preds_one)

np.testing.assert_allclose(pred_full, preds_one, atol=1e-5, rtol=1e-5)
```

### Expected Results

Identical predictions 

### Actual Results

```
Full:      [ 198.92455   100.43968    49.008053  172.73198    93.20077  -259.5672
   33.52475   140.33487    62.212387]
Batched:   [ 198.92467   100.43905    49.007904  172.73172    93.20089  -259.56866
   33.524574  140.33533    62.212276]
Max |Δ|:   0.0014648438
Per-sample Δ: [-0.00012207  0.00063324  0.00014877  0.0002594  -0.00012207  0.00146484
  0.00017548 -0.00045776  0.00011063]
```

### Versions

```shell
tabpfn:  6.4.1                                                                                                                                                                                                                                                                                                                                                                                                                        
torch:   2.10.0                                                                                                                                                                                                                                                                                                                                                                                                                       
numpy:   2.3.3                                                                                                                                                                                                                                                                                                                                                                                                                        
sklearn: 1.6.1                                                                                                                                                                                                                                                                                                                                                                                                                        
scipy:   1.16.2                                                                                                                                                                                                                                                                                                                                                                                                                       
Python:  3.13.3   


In my local experiments torch 2.8.0 is even worse in this regard.
```

klemens-floege:
@brendan-priorlabs was also able to reproduce this

[...]

PR #1096, opened 2026-07-08
add test row chunking by default [PRI-8]

### LINKED ISSUE #125: Batch predictions when test set is large
Copying a discussion from discord:
```
David Holzmüller
For me, on some (larger?) datasets (but also <= 10K train samples), I get the attached error (TabPFNClassifier), tested on torch 2.4 and torch 2.5.
 File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/classifier.py", line 533, in predict_proba
    for output, config in self.executor_.iter_outputs(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/inference.py", line 192, in iter_outputs
    output = self.model(
             ^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/transformer.py", line 413, in forward
    return self._forward(x, y, style=style, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/transformer.py", line 625, in _forward
    encoder_out = self.transformer_encoder(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/transformer.py", line 74, in forward
    x = layer(x, **kwargs)
        ^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/layer.py", line 449, in forward
    state = sublayer(state)
            ^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/layer.py", line 334, in attn_between_features
    return self.self_attn_between_features(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/multi_head_attention.py", line 355, in forward
    output: torch.Tensor = self._compute(
                           ^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/memory.py", line 100, in method_
    return x + method(self, x, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/multi_head_attention.py", line 504, in _compute
    attention_head_outputs = MultiHeadAttention.compute_attention_heads(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/storage/holzmudd/mamba_envs/envs/tabpfn_test/lib/python3.12/site-packages/tabpfn/model/multi_head_attention.py", line 710, in compute_attention_heads
    attention_head_outputs = torch.nn.functional.scaled_dot_product_attention(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: CUDA error: invalid configuration argument
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Jingang
Hi David, the CUDA error "invalid configuration argument" may occur if batch size is too large with flash attention and memory-efficient SDPA enabled.
https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html

David Holzmüller
I was hoping that TabPFN's interface would automatically batch the predict() step if there are too many samples, but apparently that is not the case
```

Seems that it makes sense to batch when the test set is too big, and maybe what too big means depend on different things including whether we're using flash attention and memory-efficient SDPA.

jmkuebler:
## Issue

Fixes #125 

## Motivation and Context

In cached inference (`fit_mode="fit_with_cache"`) currently the whole test set goes through the model in one forward pass, so peak GPU memory scales with `n_test` and large test result in OOM errors.

This PR splits the test set into chunks of at most `TABPFN_MAX_BATCHED_TEST_ROWS` rows (default **32768**) and concatenates the outputs. Test rows are independent given the KV cache, so chunking is mathematically equivalent to a single pass, but bounds peak memory (results differ by sub-tolerance float rounding, exact in float64; see [#800](https://github.com/PriorLabs/TabPFN/issues/800#issuecomment-4903444425)). 32768 matches unchunked speed in benchmarks; smaller chunks add overhead. 
Users can overwrite it with other chunk sizes or set to `0` to disable.

---

## Public API Changes

-   [ ] No Public API changes

New setting **`TABPFN_MAX_BATCHED_TEST_ROWS`** (int, default `32768`; `0`
disables). But everything is compatible with before. The default behavior changes though slightly.

---

### Functional correctness 
- New test `test__kv_cache_chunking__matches_unchunked` (classifier/regressor ×
  device): chunked vs. unchunked predictions match to `1e-10` in float64.
  - All test that are passing on main, do still pass.
  - Note that some other tests fail also on main, see https://github.com/PriorLabs/TabPFN/issues/1095

### Speed benchmark
We swiped different train and test ratios with `n_features=100` on an RTX 6000 Pro. On main we got OOM errors for `n_train=10_000, n_test=10_000_000` and for `n_train=1_000_000, n_test=200_000`.
The plots below show that very small chunk sizes result in slowdowns, but the default of `32768` is as fast as the original code, but leaves lots of margin to the OOM errors.
chunk_scaling

chunk_scaling_relative

---

## Checklist

---


src/tabpfn/errors.py
```diff
@@ -116,6 +117,10 @@ def __init__(
             f"X_test[i:i + 100])\n"
             f"        predictions.append(pred)\n"
             f"    predictions = np.vstack(predictions)\n\n"
+            f"   With fit_mode='fit_with_cache', test rows are chunked\n"
```
klemens-floege:
```suggestion
            f"   Only with fit_mode='fit_with_cache', test rows are chunked\n"
```

src/tabpfn/settings.py
```diff
@@ -61,6 +61,20 @@ class TabPFNSettings(BaseSettings):
         "Used to prevent macOS system crashes on Apple Silicon. "
         "Values > 1.0 are not recommended.",
     )
+    max_batched_test_rows: int = Field(
+        default=32768,
+        ge=0,
+        description="Maximum number of test rows fed through the model in a single "
```
klemens-floege, line 67: I think we could make this more concise

src/tabpfn/inference.py
```diff
@@ -1118,20 +1119,39 @@ def _call_model(  # noqa: PLR0913
...
+        max_rows = settings.tabpfn.max_batched_test_rows
+        n_test = X_test_tensor.shape[0]
+        if max_rows <= 0 or n_test <= max_rows:
+            return run(X_test_tensor)
+
+        outputs = [
+            run(X_test_tensor[i : i + max_rows]) for i in range(0, n_test, max_rows)
+        ]
+        if isinstance(outputs[0], dict):
+            return {k: torch.cat([o[k] for o in outputs]) for k in outputs[0]}
```
klemens-floege: CC does not like this:

```
Bug: dict outputs duplicate train_embeddings across chunks

In the cached fast path, train_emb is loaded from the KV cache (tabpfn_v3.py:1749), so when only_return_standard_out=False every chunk's dict contains the same train_embeddings tensor of shape (n_train, 1, E) (tabpfn_v3.py:1798). The new merge

return {k: torch.cat([o[k] for o in outputs]) for k in outputs[0]}

concatenates it once per chunk, yielding (n_chunks * n_train, 1, E). This path is reachable: get_embeddings(model, X, data_source="train") calls iter_outputs(..., only_return_standard_out=False) on whatever executor the model has (base.py:481-489), and its embed.ndim == 2 assert still passes — so a fit_with_cache model with n_test > 32768 silently returns duplicated train embeddings. Suggested fix: special-case that key (take it from outputs[0]), or skip chunking when only_return_standard_out=False. Either way, a test covering the embeddings path with chunking active would lock this down.
```

klemens-floege:
```suggestion
            return {
                k: outputs[0][k] if k == "train_embeddings"
                else torch.cat([o[k] for o in outputs])
                for k in outputs[0]
            }
```

tests/test_inference.py
```diff
@@ -774,3 +777,74 @@ def test__to__mps_target_device__applies_mps_linear_bug_workaround(
...
+
+    # Reference: chunking disabled -> a single forward pass over all test rows.
+    monkeypatch.setattr(settings.tabpfn, "max_batched_test_rows", 0)
+    pred_unchunked = predict(model, X_test)
+
+    # Chunked: several forward passes over test-row chunks, then concatenated.
+    monkeypatch.setattr(settings.tabpfn, "max_batched_test_rows", chunk)
+    pred_chunked = predict(model, X_test)
+
+    np.testing.assert_allclose(pred_chunked, pred_unchunked, rtol=1e-10, atol=1e-10)
```
klemens-floege, line 850:
```suggestion
    np.testing.assert_allclose(pred_chunked, pred_unchunked, rtol=1e-10, atol=1e-10)
    
    # get_embeddings exercises the dict output path
    # (only_return_standard_out=False), whose "train_embeddings" come from the
    # KV cache and are identical for every chunk — naive concatenation across
    # chunks would duplicate them, so check shape as well as values.
    for data_source in ("train", "test"):
        monkeypatch.setattr(settings.tabpfn, "max_batched_test_rows", 0)
        emb_unchunked = get_embeddings(model, X_test, data_source=data_source)

        monkeypatch.setattr(settings.tabpfn, "max_batched_test_rows", chunk)
        emb_chunked = get_embeddings(model, X_test, data_source=data_source)

        assert emb_chunked.shape == emb_unchunked.shape, (
            f"{data_source} embeddings changed shape under chunking: "
            f"{emb_chunked.shape} != {emb_unchunked.shape}"
        )
        np.testing.assert_allclose(emb_chunked, emb_unchunked, rtol=1e-10, atol=1e-10)
```

klemens-floege (commented):
Thanks a lot for looking into this! I have 2 questions: 

- How hardware dependent is this chunk size? E.g. MPS; difference CUDA devices, CPU etc.
- CC spotted a but with the embeddings functionality, added some of its suggestions maybe you can take a look :)

src/tabpfn/inference.py
jmkuebler: Thanks, that was a good catch! 
I know explicitly distinguish between keys that need to concatenate and ones that should not be replicated. Unknown keys will result in an error so we explicitly add those.

jmkuebler:
@klemens-floege Thanks for your review. I addressed the comments.
>How hardware dependent is this chunk size? E.g. MPS; difference CUDA devices, CPU etc.

I don't think a lot. Also for RTX 6000 Pro we could take anything between 16k and 500k without much impact, based on my investigation. So the room is quite large.

For CPU it really shouldn't matter as the RAM usually is very large (so we don't need to fear OOM so much and 32k is very conservatively small in that sense) and there is no parallelization to saturate (so 32k is also very large).

For other GPUs. Hardware with less memory could run into OOM earlier, but on the tested hardware the 32k default is still over 10x below the point where we would hit OOM, therefore for most hardwares it should also suffice -- if not, people can also raise an issue and we can look into specific hardware cases. Or they juts overwrite the default.

If the hardware supports much more parallelization, 32k could be slightly suboptimal to saturate, but even down to 8k on RTX we are almost optimal. So there we also have a good margin.

> CC spotted a but with the embeddings functionality, added some of its suggestions maybe you can take a look :)

Thanks! Fixed, see also my comment above.

klemens-floege (approved):
Looks great thanks for edits!

PR #1098, opened 2026-07-09
feat: add validation_frequency option to fine-tuning

### LINKED ISSUE #811: Add validation options to fine tuning

A few things that would be nice to have for finetuning:

- option to specify disabling validation
- option to specify frequency of validation
- TBC

Implement options

_No response_

### Additional context

_No response_

### Impact

None

Hemanth-1354:
## Issue
Closes #811

## Motivation and Context
Fine-tuning currently validates every epoch unconditionally. For long runs
or large validation sets this adds overhead, and there's no way to disable
validation (and the early stopping that depends on it) when you don't need it.

This adds a `validation_frequency: int | None = 1` parameter to
`FinetunedTabPFNBase` (and threads it through `FinetunedTabPFNClassifier` /
`FinetunedTabPFNRegressor`):

- `validation_frequency=1` (default): validates every epoch, identical to
  current behavior.
- `validation_frequency=N`: validates (and checks early stopping) only every
  N epochs.
- `validation_frequency=None`: skips validation for the entire run, which
  also disables early stopping. If `early_stopping=True` is passed alongside
  `validation_frequency=None`, a `UserWarning` is raised since the flag would
  otherwise silently have no effect.

Skipped epochs report `NaN` as the validation metric, reusing the existing
`not np.isnan(...)` guards already in the checkpoint-saving and
early-stopping code paths. Early stopping's restore-to-best-weights logic is
gated on a derived `effective_early_stopping` flag rather than the raw
`self.early_stopping`, so a run with `validation_frequency=None` can't
accidentally revert to the pre-training weights at the end.

## Public API Changes
- [ ] No Public API changes
- [x] Yes, Public API changes (Details below)

New optional keyword-only parameter `validation_frequency` on
`FinetunedTabPFNBase.__init__`, `FinetunedTabPFNClassifier.__init__`, and
`FinetunedTabPFNRegressor.__init__`. Defaults to `1`, which preserves
existing behavior exactly, no breaking changes.

## How Has This Been Tested?
- Added `test__finetuned_tabpfn_classifier__validation_frequency_skips_epochs`:
  mocks `_evaluate_model` and asserts it's called only on the expected
  epochs (plus the one pre-loop baseline eval) when `validation_frequency=2`.
- Added `test__finetuned_tabpfn_classifier__validation_frequency_none_disables_es`:
  asserts the `UserWarning` fires for `early_stopping=True` +
  `validation_frequency=None`, and that fine-tuned weights are *not* reverted
  to the base model despite `early_stopping=True`, proving early stopping is
  actually disabled rather than just skipped-but-still-restoring.
- `ruff check` / `ruff format --check` pass on all changed files.
- Existing finetuning test suite is unaffected (default `validation_frequency=1`
  reproduces the prior code path exactly).

## Checklist

### DIFF
--- a/changelog/811.added.md
+++ b/changelog/811.added.md
@@ -1,0 +1,1 @@
+`validation_frequency` option for `FinetunedTabPFNClassifier`/`FinetunedTabPFNRegressor` fine-tuning, allowing validation (and early stopping) to run every N epochs instead of every epoch, or be disabled entirely with `validation_frequency=None`.
--- a/src/tabpfn/finetuning/finetuned_base.py
+++ b/src/tabpfn/finetuning/finetuned_base.py
@@ -293,6 +293,12 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
             early stopping. Defaults to 8.
         min_delta: Minimum change in metric to be considered as an improvement.
             Defaults to 1e-4.
+        validation_frequency: Number of epochs between validation runs. If 1
+            (default), validation runs every epoch, preserving existing
+            behavior. If N > 1, validation (and any early-stopping check)
+            only runs every N epochs. Must be a positive integer. When
+            ``validation_frequency > 1``, ``early_stopping_patience`` counts
+            validation runs, not epochs.
         grad_clip_value: Maximum norm for gradient clipping. If None, gradient
             clipping is disabled. Gradient clipping helps stabilize training by
             preventing exploding gradients. Defaults to 1.0.
@@ -352,6 +358,7 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
         early_stopping: bool = True,
         early_stopping_patience: int = 8,
         min_delta: float = 1e-4,
+        validation_frequency: int = 1,
         grad_clip_value: float | None = 1.0,
         use_lr_scheduler: bool = True,
         lr_warmup_only: bool = False,
@@ -380,6 +387,7 @@ class FinetunedTabPFNBase(BaseEstimator, ABC):
         self.early_stopping = early_stopping
         self.early_stopping_patience = early_stopping_patience
         self.min_delta = min_delta
+        self.validation_frequency = validation_frequency
         self.grad_clip_value = grad_clip_value
         self.use_lr_scheduler = use_lr_scheduler
         self.lr_warmup_only = lr_warmup_only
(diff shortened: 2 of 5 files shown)


anuragg1209:
Hi @Hemanth-1354, thanks for the PR — well-thought-out description and tests, and `validation_frequency` is a feature we'd like to have!

The PR needs a rebase: **#1101** landed after you opened this and reworked `finetuned_base.py` (validation gating, checkpointing, early stopping), which is where the conflicts come from. It also means `main` now already supports **disabling validation entirely** via `validation_split_ratio=None` (or `0`) — trains on all data, skips per-epoch eval, and auto-disables early stopping — so the `validation_frequency=None` case here is now redundant.

After rebasing, we'd suggest narrowing the scope to the missing part — **validate every N epochs** — which also shrinks the diff:

- `validation_frequency: int = 1` (positive int only, no `None`) — drops the `None` type-checks and the `UserWarning` interaction with `early_stopping`.
- `main` now has the seams you need (`do_validation`, `early_stopping_enabled`, an else-branch logging train loss), so the integration reduces to one derived flag at the per-epoch eval site:

  ```python
  run_validation = do_validation and (epoch + 1) % self.validation_frequency == 0
  ```

- Your NaN-sentinel for skipped epochs still fits: the early-stopping block is already guarded by `not np.isnan(primary_metric)`, so patience won't tick on skipped epochs. Please keep your change moving the NaN check into `is_best` — interval checkpoints are for crash recovery/resume and don't need a validation metric, so they should still save on skipped epochs.
- On skipped epochs, the checkpoint `metrics` dict shouldn't contain the NaN eval result — `main` already has a fallback to `{"train_loss": mean_train_loss}` for the no-validation case, so just gate that branch on `run_validation` instead of `do_validation`.
- Small docs note: with `validation_frequency > 1`, `early_stopping_patience` counts validation runs, not epochs.

Your `validation_frequency=2` test should carry over as-is; the `None` test can be dropped along with that case.

Sorry for the moving target — the timing with #1101 was unlucky. Happy to help if anything is unclear during the rebase, and thanks again for contributing!

Hemanth-1354:
Thanks for the detailed review @anuragg1209! I've rebased on main and narrowed the scope as suggested:

- Dropped `validation_frequency=None` - now just `validation_frequency: int = 1`
- Simplified integration using main's `do_validation` / `early_stopping_enabled` seams
- One-line flag: `run_validation = do_validation and (epoch + 1) % self.validation_frequency == 0`
- NaN sentinel moved into `is_best` so interval checkpoints save on skipped epochs
- Checkpoint metrics gated on `run_validation` instead of `do_validation`
- Added positive integer validation
- Dropped the `None` test, kept `validation_frequency=2` test

@anuragg1209

anuragg1209:
Hi @Hemanth-1354, thanks again for proposing validation_frequency and for adapting the implementation as the fine-tuning code evolved. We retained the core scheduling approach and moved it forward on top of the current main. The main differences are that validation disabling continues to use the existing validation_split_ratio=None/0 API, validation frequency is validated at fit time as a positive integer, and we added resume/checkpoint/early-stopping safeguards and broader test coverage. 

We’ll acknowledge your contribution in the PR, so we added you as a co-author on the feature commit to ensure your contribution is recorded in the Git history.

We'll close this PR in favor of #1144, of which you are also a co-author. Thanks!

Issue #1104, opened 2026-07-10
Overflow in Yeo-Johnson inverse transform clip bound (RuntimeWarning: overflow encountered in cast)


tmacleod:
### Describe the bug

`_yeojohnson_inverse_transform()` clips the `expm1` argument to `log(finfo(dtype).max)` to keep the result within the _output_ dtype's representable range. That bound is computed at the output dtype's precision (e.g. `float32`), not `float64`. When the clipped value sits exactly at that boundary, `expm1` can round up past `finfo(dtype).max` by a few ULPs during the float64-internal computation, and casting that result back into the `float32` output array then overflows:
```
safe_power_transformer.py:157: RuntimeWarning: overflow encountered in cast
  x_inv[pos] = np.expm1(...)
```
Both branches (`x >= 0` and `x < 0`) are affected, since both clip against the same `max_arg`.

### Steps/Code to Reproduce

Minimal repro of the underlying float precision issue:
```
import numpy as np
dtype = np.float32
max_arg = np.log(np.finfo(dtype).max)
print(np.expm1(max_arg).astype(dtype))  # -> inf, RuntimeWarning: overflow encountered in cast
```
Reproduced directly against _yeojohnson_inverse_transform:
```
import warnings
import numpy as np
from tabpfn.preprocessing.steps.safe_power_transformer import _yeojohnson_inverse_transform

dtype, lmbda = np.float32, 0.5
max_arg64 = float(np.log(np.finfo(dtype).max))
x_boundary = np.expm1(max_arg64 * lmbda) / lmbda
x = np.array([x_boundary, x_boundary * 1.5, x_boundary * 3.0], dtype=dtype)

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    out = _yeojohnson_inverse_transform(x, lmbda)
    print([str(w.message) for w in caught])
print(out)
```

### Expected Results

```
[]
[3.4027206e+38 3.4027206e+38 3.4027206e+38]
```
No warning, and a finite result just under `finfo(float32).max`.

### Actual Results

```
['overflow encountered in cast']
[inf inf inf]
```

### Versions

```shell
Collecting system and dependency information...
PyTorch version: 2.12.1+cu132
CUDA used to build PyTorch: 13.2
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 11 Pro (10.0.26200 64-bit)
GCC version: Could not collect
Clang version: Could not collect
CMake version: version 3.31.6
[...]

PR #1105, opened 2026-07-10
Fix overflow in Yeo-Johnson inverse transform clip bound

tmacleod:
_yeojohnson_inverse_transform() clips the expm1 argument to log(finfo(dtype).max) to prevent overflow, but that bound is computed in the OUTPUT dtype (e.g. float32) rather than float64. Evaluating expm1() at that exact boundary can round up past finfo(dtype).max by a few ULPs, which then overflows on cast back into the output array:

  RuntimeWarning: overflow encountered in cast
    x_inv[pos] = np.expm1(...)

Fix: nudge max_arg down by a few ULPs so expm1's result reliably stays within dtype's representable range after the float64 round-trip.

Repro (minimal):
  import numpy as np
  dtype = np.float32
  max_arg = np.log(np.finfo(dtype).max)
  print(np.expm1(max_arg).astype(dtype))  # -> inf, RuntimeWarning

Verified before/after with a fuller script exercising both the positive and negative branches at the float32 boundary: before the fix, three boundary cases overflow to +-inf with a warning; after, all three are finite (~3.4e38, just under float32's true max) with zero warnings. float64 inputs are unaffected in both runs (the bound never gets close enough to overflow at that precision).

## Issue

https://github.com/PriorLabs/TabPFN/issues/1104

---

## Motivation and Context

Fix overflow in Yeo-Johnson inverse transform clip bound

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

Tested locally and using Claude.

---

## Checklist

---


klemens-floege (commented):
@tmacleod thank you for opening a PR into our repo; I think the code looks good but having some test that actually checks that this is working would be great. 


Idea: 
```
def test__yeojohnson_inverse_transform__float32_boundary__finite_no_overflow():
    """Inputs at the float32 extremes must clip safely, not overflow to inf."""
    x = np.array([3.4e38, -3.4e38], dtype=np.float32)
    for lmbda in [0.0, 0.5, 1.5, 2.0]:  # covers all four branches
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            result = _yeojohnson_inverse_transform(x.copy(), lmbda)
        overflow = [x_ for x_ in w if "expm1" in str(x_.message) or "cast" in str(x_.message)]
        assert not overflow
        assert np.all(np.isfinite(result))
```

tmacleod:
@klemens-floege Thank you for the review and the test suggestion. I added the following regression test to `tests/test_preprocessing/test_power_transformer.py`:

```python
def test__yeojohnson_inverse_transform__float32_boundary__finite_no_overflow():
    """Float32 boundary inputs must clip safely without overflow."""
    float32_max = np.finfo(np.float32).max

    cases = [
        (float32_max, 0.0),   # x >= 0, lambda == 0
        (float32_max, 0.5),   # x >= 0, lambda != 0
        (-float32_max, 1.5),  # x < 0, lambda != 2
        (-float32_max, 2.0),  # x < 0, lambda == 2
    ]

    for value, lmbda in cases:
        x = np.array([value], dtype=np.float32)

        with warnings.catch_warnings():
            warnings.simplefilter("error", RuntimeWarning)
            result = _yeojohnson_inverse_transform(x, lmbda)

        assert result.dtype == np.float32
        assert np.all(np.isfinite(result))
```

The test covers all four inverse-transform branches and treats runtime warnings as errors. I confirmed that it fails with the original `max_arg` calculation and passes with the four-ULP margin. The relevant test file and the full non-slow test suite also pass with the fix.

klemens-floege (approved):
@tmacleod thanks a lot for making a contribution!

PR #1106, opened 2026-07-12
Small cleanup in TabPFN (#1104)

bglglzd:
This is a focused change for the cited issue with minimal side effects.

Related to #1104.


klemens-floege (commented):
@bglglzd than you for opening a PR into our repo; it looks like you are just changing some formatting here. Is this intended?

Issue #223, opened 2025-03-01
Investigate and improve cross-platform prediction consistency


noahho:
## Issue Description

When running TabPFN consistency tests across different platforms (e.g., macOS vs Linux, x86 vs ARM), we've observed significant differences in model predictions. 

## Current Observations:

1. Despite using , regression predictions on diabetes dataset still show differences:
   - On macOS (ARM): 
   - On Linux CI: 
   - Difference: ~2.34 (about ~1.6% relative difference)

2. Classification predictions seem more stable but still show small variations

## Impact:
- Makes it difficult to have reproducible research/benchmarks across platforms
- Requires platform-specific consistency tests (as implemented in PR #217)
- Could affect production deployments across different infrastructures

## Potential Causes:
- Different CPU architectures (x86 vs. ARM)
- Different BLAS/LAPACK implementations
- OS-specific optimizations
- Compiler-specific floating-point optimizations

## Suggested Solutions to Investigate:
1. More aggressive precision control beyond sklearn's 16-decimal option
2. Implementation of deterministic mode that sacrifices some performance for better consistency
3. Platform detection with environment-specific reference values
4. Custom normalization/scaling approaches that are more robust to platform differences

## Related PR:
PR #217 worked around this by making consistency tests platform-specific, but we should investigate a more fundamental solution.

## Priority:
Medium - This is not breaking functionality but affects reproducibility

klemens-floege:
@priorphil @bejaeger do you think this was fixed with the "saving of random embeddings" fix? I think either way we probably still need to test this properly.

bejaeger:
I think the first thing we should do is to run the consistency tests in float64. Maybe that already does the trick?

klemens-floege:
@bejaeger Gave it a go here: https://github.com/PriorLabs/TabPFN/pull/1111; it seems that the time increased for some of the hardware types, but lmk what you think :)

klemens-floege:
@bejaeger We merge the float64 testing yesterday: https://github.com/PriorLabs/TabPFN/pull/1111; we also investigated whether we still have cross platform differences in float64, and we do. Shall we keep this issue still open for now or do we think float64 testing is already good enough? 

CC: @jmkuebler

bejaeger:
Can be closed from my side

klemens-floege:
Sounds good, will close this issue for now. Feel free to re-open :)

PR #1111, opened 2026-07-13
Run consistency tests in float64

klemens-floege:
## Motivation

The consistency tests compare model predictions against stored references and are currently limited to `darwin_arm64` because float32 inference produces platform-dependent results (different BLAS backends accumulate rounding differences that softmax can amplify).

This PR switches the test config to `inference_precision: torch.float64` to reduce that divergence, as a step toward enabling the consistency tests on more platforms.

## Changes

- `DEFAULT_CONFIG` in `tests/test_consistency.py` now uses `torch.float64` instead of `torch.float32`.
- Regenerated all 18 `darwin_arm64` reference predictions via `python -m tests.test_consistency`.

## Notes

- float64 is not a no-op vs float32: with the old references, 17/18 cases still passed within tolerance, but `classifier_tiny_dataset_v3` shifted one probability by 0.007 absolute (0.045 → 0.038), so references had to be regenerated.
- The `macos-latest` CI runner exercises these tests on a different machine than the one that generated the references, so a green CI here is a first cross-machine consistency signal for float64.
- Follow-up (separate PR): add a platform string for linux x86 to `ENABLED_PLATFORMS` and check whether float64 references generated on macOS ARM reproduce there.


klemens-floege:
Linked to Issue: #223

jmkuebler:
I think this is a good idea, thanks @klemens-floege !

jmkuebler (approved):
Thank you @klemens-floege! Maybe we can consider being a bit more specific that these consistency test are there to ensure we did not mess with the float64 model pass. This thus ensures that we did not mess up anything in the high-level architecture

But  it can still mean that the default path on some hardware has a regression.

Also: could it even be that with Float 64 even between hardwares we are consistent (worth a check once you do it for other hardware).

klemens-floege:
Good point, updating the docstring now!

klemens-floege:
> Also: could it even be that with Float 64 even between hardwares we are consistent (worth a check once you do it for other hardware).

Hmm let me get CC to check some things

klemens-floege:
@jmkuebler let's test your hypothesis here: PriorLabs/TabPFN#1117

PR #1113, opened 2026-07-13
Synchronize the single-device stream in parallel_execute so timing reflects GPU compute

### LINKED ISSUE #1112: model forward timing is not accurate on GPU
### Describe the bug

The model forward timing collects the time before the GPU is finished (which happens asynchronously).

### Steps/Code to Reproduce

N/A

### Expected Results

N/A

### Actual Results

N/A

### Versions

```shell
N/A
```

jmkuebler:
## Issue
Resolves #1112 1112

### What
In `_execute_in_current_thread` (the single-device path of `parallel_execute`),
call `torch.cuda.synchronize(device)` before yielding each result. 
This enables to correctly measure the forward time, without adding any overhead.

### Why
`InferenceEngine._speed_metrics` times the fit KV-cache build and the predict
forward by wrapping this generator in a `_TimedIterator`. On a single GPU the
prior path returned right after *launching* kernels, so the timer captured
kernel-launch time, not GPU compute — undercounting the forward badly and
increasingly with load. The multi-device path already forces completion (it
waits on a per-output `torch.cuda.Event`); this brings the single-device path
in line.

This is a timing-instrumentation fix, **not** a correctness change: single-device
consumers run on the same default stream, so result ordering was already correct
without the sync.

### Impact
Measured on one RTX PRO 6000, n_train=200k, 50 cols, fit_with_cache:

| config | phase | model_forward_s (before → after) | median_s (before → after) |
|---|---|---|---|
| n_est=1 | predict n=1    | 0.0125 → 0.0174 | 0.0210 → 0.0210 |
| n_est=1 | predict n=1000 | 0.0139 → 0.0577 | 0.0617 → 0.0617 |
| n_est=1 | fit            | 5.75 → 7.18     | 7.81 → 7.79     |
| n_est=8 | predict n=1    | 0.0972 → 0.1400 | 0.1416 → 0.1456 |
| n_est=8 | predict n=1000 | 0.1086 → 0.4607 | 0.4669 → 0.4673 |
| n_est=8 | fit            | 55.8 → 57.3     | 58.5 → 58.5     |

- Forward timings become accurate (predict was ~4× undercounted).
- End-to-end latency (`median_s`) is unchanged within noise → no measurable
  overhead, including at n_estimators=8.

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
(see above description)
---

## Checklist

---

### DIFF
--- a/src/tabpfn/parallel_execute.py
+++ b/src/tabpfn/parallel_execute.py
@@ -94,15 +94,7 @@ def _execute_in_current_thread(
     device: torch.device, functions: Iterable[ParallelFunction[R_co]]
 ) -> Generator[R_co]:
     for function in functions:
-        output = function(device=device)
-        # Force GPU completion before yielding so a _TimedIterator around this
-        # measures real compute, not just kernel-launch time.
-        # Negligible overhead
-        if device.type == "cuda":
-            torch.cuda.synchronize(device)
-        elif device.type == "mps" and hasattr(torch, "mps"):
-            torch.mps.synchronize()
-        yield output
+        yield function(device=device)
 
 
 def _execute_with_multithreading(
--- a/src/tabpfn/parallel_execute.py
+++ b/src/tabpfn/parallel_execute.py
@@ -100,6 +100,8 @@ def _execute_in_current_thread(
         # Negligible overhead
         if device.type == "cuda":
             torch.cuda.synchronize(device)
+        elif device.type == "mps" and hasattr(torch, "mps"):
+            torch.mps.synchronize()
         yield output
 
 
--- a/src/tabpfn/parallel_execute.py
+++ b/src/tabpfn/parallel_execute.py
@@ -94,7 +94,13 @@ def _execute_in_current_thread(
     device: torch.device, functions: Iterable[ParallelFunction[R_co]]
 ) -> Generator[R_co]:
     for function in functions:
-        yield function(device=device)
+        output = function(device=device)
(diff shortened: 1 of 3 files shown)


src/tabpfn/parallel_execute.py
```diff
@@ -94,7 +94,15 @@ def _execute_in_current_thread(
...
-        yield function(device=device)
+        output = function(device=device)
+        # Force GPU completion before yielding so a _TimedIterator around this
+        # measures real compute, not just kernel-launch time.
+        # Negligible overhead
+        if device.type == "cuda":
+            torch.cuda.synchronize(device)
+        elif device.type == "mps" and hasattr(torch, "mps"):
+            torch.mps.synchronize()
+        yield output
```
bejaeger: wondering, could that be part of the `_TimedIterator` itself? Or would that corrupt the multithread forward?

jmkuebler: Thanks @bejaeger . That indeed seems to be more elegant (also in case we ever need to turn that off). 

I gate though on the number of devices. For the multi-device / multi-thread case we are not adding any blocking statements on one specific GPU.

jmkuebler: I also made `InferenceEngineBatchedNoPreprocessing` use the timed iterator now, instead of the handwritten timer there. 
Or was there any particular reason for this hand-written timing there?

bejaeger (approved):
LGTM!

PR #1117, opened 2026-07-14
Run consistency tests on all CI platforms against shared float64 references

klemens-floege:
Stacked on #1111.

Follow-up to the review discussion on #1111: check whether float64 inference is consistent even across hardware.

Previously the consistency tests only ran on `darwin_arm64` and were silently skipped on the other CI runners. This PR:

- Enables the tests on `linux_x86_64` and `windows_amd64` (all CI runners: macos, ubuntu, windows).
- Collapses the per-platform reference layout into a single shared float64 reference set: the JSONs move from `tests/reference_predictions/darwin_arm64/` up to `tests/reference_predictions/`, and the platform-dir indirection is removed.

The CI matrix on this PR is the actual experiment: green on ubuntu + windows means float64 predictions are cross-hardware consistent within the existing tolerances. If a platform diverges, we reintroduce a per-platform reference directory for it (see git history of `test_consistency.py`).


klemens-floege:

- #1117 ◀

Issue #1116, opened 2026-07-14
TabPFN for Bayesian optimization


songlei00:
Hi TabPFN Team,

Thank you for your incredible work on TabPFN!

I have been studying the source code of [PFNs4BO](https://github.com/automl/PFNs4BO/tree/main) and noticed this repository for the TabPFN foundation model. Since this repo leverages a stronger version of PFN, I would love to use it to power my own Bayesian Optimization (BO) workflow.

I noticed that a differentiable version of the PFN regressor was recently released. This is extremely promising for BO, as it theoretically allows us to optimize the acquisition function via gradient descent. However, I am a bit unsure how to correctly invoke the forward pass to obtain the logits required for calculating the Expected Improvement (EI). **Since the standard predict() method adheres to the scikit-learn API, passing a PyTorch Tensor directly into it throws an error as it only supports scikit-learn compatible inputs (like NumPy arrays), which unfortunately breaks the PyTorch computation graph.**

Therefore, I need to find a way to compute the logits and criterion via a lower-level forward-pass call that natively supports Tensors and preserves gradients. Could you please provide a minimal working example on how to achieve something like the following?

``` python
import torch
# Assuming data_X requires gradient for acquisition optimization
data_X.requires_grad_() 

model = TabPFNRegressor(xxx)
model.fit_with_differentiable_input(train_X, train_y)

# We want to obtain logits and criterion while keeping the gradients intact back to data_X
# (Instead of using the sklearn-style predict() which throws errors for Tensors)
result = model.forward_like_call(data_X, output_type="full") 
logits: torch.Tensor = result["logits"]
criterion = result["criterion"]

# Compute EI and backpropagate to get gradients w.r.t data_X
ei_values = criterion.ei(logits, best_f=best_f, maximize=True)
ei_values.backward() 
# print(data_X.grad)
```

Any guidance or code snippets on how to properly set up this differentiable pipeline would be highly appreciated!

klemens-floege:
@songlei00 thank you for opening an Issue and for your kind words; the [PFNs4BO](https://github.com/automl/PFNs4BO/tree/main) is very cool work but comes from a different codebase. We are also excited about BO but do not support it yet out of the box in our package. Give us a bit of time to look into this and we will get back to you.

klemens-floege:
[...]

PR #1118, opened 2026-07-14
Add Bayesian optimization example with differentiable EI

klemens-floege:
## What

Adds `examples/bayesian_optimization.py`: a self-contained Bayesian optimization loop using `TabPFNRegressor` as the surrogate model, in the spirit of [PFNs4BO](https://github.com/automl/PFNs4BO) (Müller et al., ICML 2023) but powered by the TabPFN foundation model.

Addresses #1116, which asks how to compute Expected Improvement from the predicted distribution and backpropagate to the candidate inputs.

The benchmark objective (Hartmann-6) comes from `botorch.test_functions`, so the script contains only the BO loop and the TabPFN surrogate code (requires `pip install botorch` to run).

## How it works

Each BO iteration:
1. Fits on the evaluated points via `fit_with_differentiable_input` (tensors in, graph preserved).
2. Screens 512 random candidates with closed-form EI from `raw_space_bardist_.ei()` in a single forward pass — no Gaussian assumption, EI is integrated directly over the predicted bar distribution.
3. Refines the top 4 candidates by **gradient ascent on EI itself**: `differentiable_input=True` lets gradients flow from the acquisition value back to the candidate coordinates.
4. Evaluates the objective at the best candidate and repeats.

## Results

On Hartmann-6 (6D, a benchmark where random search does poorly), with 10 init + 25 BO evaluations, seed 0 (maximizing the negated function, optimum 3.3224):

| | best f found |
|---|---|
| TabPFN-BO | **2.51** |
| Random search (same budget) | 1.37 |

The win is consistent across the seeds tested (0, 1, 2). Runs in ~35 s on an M-series CPU.

### DIFF
--- a/examples/bayesian_optimization.py
+++ b/examples/bayesian_optimization.py
@@ -9,27 +9,25 @@ PFNs4BO approach (Mueller et al., ICML 2023, https://arxiv.org/abs/2305.17535,
 https://github.com/automl/PFNs4BO), using the TabPFN foundation model as the
 surrogate.
 
-The loop below minimizes the Hartmann-6 function (a classic 6D BO benchmark
-where random search does poorly):
+Each iteration of the loop below:
 
-1. Fit TabPFN on the points evaluated so far via
+1. Fits TabPFN on the points evaluated so far via
    ``fit_with_differentiable_input`` (tensors in, gradients preserved).
-2. Score a batch of random candidates with EI in a single forward pass.
-3. Refine the most promising candidates by gradient *ascent on EI itself* —
+2. Scores a batch of random candidates with EI in a single forward pass.
+3. Refines the most promising candidates by gradient *ascent on EI itself* —
    ``differentiable_input=True`` lets gradients flow from the acquisition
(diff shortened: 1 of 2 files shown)


examples/bayesian_optimization.py
```diff
@@ -0,0 +1,154 @@
...
+        [1312, 1696, 5569, 124, 8283, 5886],
+        [2329, 4135, 8307, 3736, 1004, 9991],
+        [2348, 1451, 3522, 2883, 3047, 6650],
+        [4047, 8828, 8732, 5743, 1091, 381],
+    ]
+)
+HARTMANN_ALPHA = torch.tensor([1.0, 1.2, 3.0, 3.2])
+
+
+def hartmann6(x: torch.Tensor) -> torch.Tensor:
```
eliott-kalfon: As a general comment, I think we could make this script even easier to understand by spending less time defining the function we try to optimise. What if we could use another space (even simpler) that we could define with 2-3 lines of code?
I am in favour of do-one-thing and optimising for reading

PR #1099, opened 2026-07-09
Disable KV cache quantization when inference dtype specified.

oscarkey:
A limitation of this approach is that you can't have the dtype automatically determined and also use this for the KV cache, but I'd guess anyone who is in deep enough to disable cache quantization is happy specifying a dtype.

Closes https://github.com/PriorLabs/TabPFN/issues/631
Fixes PRI-154

PR #1120, opened 2026-07-16
add KV cache dtype as explicit argument [PRI-347]

### LINKED ISSUE #1119: [PRI-347] Expose `kv_cache_dtype`

Currently the KV cache is quantized by default and cannot be used unquantized without hacking. 

add `kv_cache_dtype` as an argument

_No response_

### Additional context

_No response_

### Impact

Low (Nice to have)

jmkuebler:
## Issue

Fixes PriorLabs/TabPFN#1119 and PriorLabs/TabPFN#631.  <br>Also this is an alternative to PriorLabs/TabPFN#1099

## Motivation and Context

`kv_cache_dtype`is a tradeoff decision and should be possible to control explicitly

---

## Public API Changes

- [ ] No Public API changes
- [X] Yes, Public API changes (Details below)

---

## How Has This Been Tested?

I fixed the unit tests to not rely on Monkeypatching anymore.  <br>I also ran the script shared in [https://github.com/PriorLabs/TabPFN/issues/631](<https://github.com/PriorLabs/TabPFN/issues/631>)

```python
import random
from sklearn.datasets import load_diabetes
import numpy as np
import torch

X, y = load_diabetes(return_X_y=True)
    X, y, test_size=0.33, random_state=42
)


def _set_seeds() -> None:
    torch.manual_seed(0)
    np.random.seed(0)
    random.seed(0)


_set_seeds()
reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache = reg.predict(X_test)

reg = TabPFNRegressor(fit_mode="low_memory", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_no_cache_repeat = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_preprocessors", inference_precision=torch.float64)
reg.fit(X_train, y_train)
preds_cache_preproc = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_with_cache", inference_precision=torch.float64, kv_cache_dtype="auto")
reg.fit(X_train, y_train)
preds_kv_cache_auto = reg.predict(X_test)

_set_seeds()
reg = TabPFNRegressor(fit_mode="fit_with_cache", inference_precision=torch.float64, kv_cache_dtype="int8")
reg.fit(X_train, y_train)
preds_kv_cache_int8 = reg.predict(X_test)


def _max_diff(a: np.ndarray, b: np.ndarray) -> float:
    return np.max(np.abs(a - b) / np.abs(a))


print("max relative diffs")
print("no_cache vs no_cache_repeat:", _max_diff(preds_no_cache, preds_no_cache_repeat))
print("no_cache vs cache_preproc:", _max_diff(preds_no_cache, preds_cache_preproc))
print("no_cache vs kv_cache (auto):", _max_diff(preds_no_cache, preds_kv_cache_auto))
print("no_cache vs kv_cache int8:", _max_diff(preds_no_cache, preds_kv_cache_int8))
```

Using `kv_cache_dtype="auto"` fixes the issue. If `int8`is used we now throw a warning, when someone forces the inference dtype to sth else.

```
/usr/lib/python3.12/contextlib.py:81: UserWarning: inference_precision forces a dtype but kv_cache_dtype is 'int8', so the KV cache is still stored as int8 regardless of the forced dtype. Pass kv_cache_dtype='auto' to keep the KV cache in the forced dtype.
  return func(*args, **kwds)
max relative diffs
no_cache vs no_cache_repeat: 0.0
no_cache vs cache_preproc: 0.0
no_cache vs kv_cache (auto): 0.0
```

---

## Checklist

- [ ] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.

---

```
```

### DIFF
--- a/src/tabpfn/architectures/interface.py
+++ b/src/tabpfn/architectures/interface.py
@@ -255,9 +255,7 @@ class Architecture(nn.Module, ABC):
 
         ``"auto"`` (keep the computed dtype) is always supported. Architectures
         whose cache implements quantization override this to additionally list
-        e.g. ``"int8"``. Used to resolve the estimator's ``kv_cache_precision``:
-        ``None`` picks the most compact supported value, and an unsupported
-        request falls back to ``"auto"``.
+        e.g. ``"int8"``.
         """
         return ("auto",)
 
--- a/src/tabpfn/architectures/interface.py
+++ b/src/tabpfn/architectures/interface.py
@@ -250,6 +250,17 @@ class Architecture(nn.Module, ABC):
             use_chunkwise_inference=False,
         )
 
+    def get_supported_kv_cache_precisions(self) -> tuple[str, ...]:
+        """KV cache storage dtypes this architecture supports (``fit_with_cache``).
+
+        ``"auto"`` (keep the computed dtype) is always supported. Architectures
+        whose cache implements quantization override this to additionally list
+        e.g. ``"int8"``. Used to resolve the estimator's ``kv_cache_precision``:
+        ``None`` picks the most compact supported value, and an unsupported
+        request falls back to ``"auto"``.
+        """
+        return ("auto",)
+
     @property
     @abstractmethod
     def embedding_dim(self) -> int:
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -262,7 +262,7 @@ def get_cache_size(
     n_features: int,
     model_config: TabPFNV3Config,
     base_dtype: torch.dtype | Literal["autocast"],
-    quantize_kv_cache: bool = True,
+    kv_cache_dtype: Literal["auto", "int8"] = "int8",
 ) -> int:
     """Calculate the cached memory in bytes for a single TabPFN v3 estimator.
 
@@ -289,8 +289,8 @@ def get_cache_size(
       reduction/norm-lineage tensors (``inducing_hidden``, ``scaler_cache``)
       stay fp32.
 
-    The KV cache is additionally int8-quantized when ``quantize_kv_cache`` is
-    True (mirroring the engine's ``kv_cache_dtype="int8"``).
+    The KV cache is additionally int8-quantized when ``kv_cache_dtype="int8"``
+    (mirroring the engine's ``kv_cache_dtype`` option).
 
     Args:
         n_train: Number of training rows. The KV cache and train activations
@@ -306,16 +306,18 @@ def get_cache_size(
             ``"autocast"`` (GPU autocast path -- KV and ``train_embeddings`` are
             sized at fp16 while ``inducing_hidden`` and ``scaler_cache`` stay
             fp32, since autocast keeps those ops in fp32).
-        quantize_kv_cache: If True (default), the KV cache is sized at
+        kv_cache_dtype: If ``"int8"`` (default), the KV cache is sized at
             :data:`~tabpfn.architectures.kv_cache.QUANTIZED_KV_DTYPE` (mirrors
             the engine's ``kv_cache_dtype="int8"``) plus per-tensor scales at
-            the KV compute dtype; if False, the K/V are sized at the compute
-            dtype with no scales.
+            the KV compute dtype; if ``"auto"``, the K/V are sized at the
+            compute dtype with no scales.
 
     Returns:
         Per-estimator cache size in bytes. Multiply by the ensemble size for the
         total (each estimator holds its own cache); divide by ``1024 ** 2`` for MB.
     """
+    quantize_kv_cache = kv_cache_dtype == "int8"
+
     # Set the stored dtype of each cached component up front. On the forced-
     # precision path the model and inputs are cast to ``dtype``, so every
     # component is ``dtype``. Under autocast the cache is mixed precision.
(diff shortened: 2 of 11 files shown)


bejaeger (commented):
Looks great thanks a lot!

I forgot to mention that not all our models currently supported kv cache quantization. So not only is the new `kv_cache_dtype` argument only relevant when fit_mode is `fit_with_cache` but also only if the model is `v3` (which is the only one that currently supports int8). This is not ideal and might be confusing if we have "int8" as the default. Should we have another "magic" option like "auto" that is "smallest" or "optimized" which will always pick the smallest supported kv cache dtype?

src/tabpfn/regressor.py
```diff
@@ -243,6 +244,7 @@ def __init__(  # noqa: PLR0913
         ] = "fit_preprocessors",
         memory_saving_mode: MemorySavingMode = "auto",
         keep_cache_on_device: bool = True,
+        kv_cache_dtype: Literal["auto", "int8"] = "int8",
```
oscarkey: I'm slightly worried the parameter default value might cause us an issue if future models need a different value, because this default can't depend on the model version.

What if the default was `auto` and:

* cache_precision=auto + inference_precision=auto → cache dtype set to int8 for v3
* cache_precision=auto + inference_precision=specified as A → cache dtype is set to A
* cache_precision=specified as A + inference_precision=specified as B → cache dtype is set to A
  * This way we can also delete the warning, as the user is explicitly making the odd choice.

oscarkey: Will many end users change this? If not, we could put it in `inference_config.py`. It's still possible to override, but keeps the constructor simpler. Happy either way though.

oscarkey: Should the name be `cache_precision`?

* Consistent `keep_cache_on_device` and `inference_precision`
* Lots of our users might not be familiar with transformers and so not know what "kv" means.

Or maybe `cache_dtype`.

oscarkey (commented):
Just a few thoughts! I'm on holiday tomorrow and monday, so happy with whatever's decided if you want to merge :)

src/tabpfn/regressor.py
oscarkey: although this could be confused with the caching we already do of the input data in some of the fit modes, or any future cache we add, hmmm

bejaeger: Yeah we briefly discussed this actually. The problem is that there is more in the cache than just the kv cache and this is currently not quantized. So we opted for staying explicit here. I agree though that it's not ideal because very technical.
Maybe "hiding" it in the inference config is the way to go?

I'm indifferent on `precision` vs `dtype`. Harmonizing with `inference_precision` might be cleaner.

jmkuebler: kv_cache_precision it is

jmkuebler: the model version now has in the architecture what is supported and we resolve against that. Also made the argument optional. If `None` is provided we resolve to the defaults, otherwise we respect the user choice unless not supported, in whcih case we warn

src/tabpfn/architectures/interface.py
```diff
@@ -250,6 +250,17 @@ def get_default_performance_options(self) -> PerformanceOptions:
...
         )
 
+    def get_supported_kv_cache_precisions(self) -> tuple[str, ...]:
+        """KV cache storage dtypes this architecture supports (``fit_with_cache``).
+
+        ``"auto"`` (keep the computed dtype) is always supported. Architectures
+        whose cache implements quantization override this to additionally list
+        e.g. ``"int8"``. Used to resolve the estimator's ``kv_cache_precision``:
+        ``None`` picks the most compact supported value, and an unsupported
+        request falls back to ``"auto"``.
```
bejaeger: the last sentence seems like it doesn't belong here and rather should be in the estimator doc string?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1930,6 +1936,10 @@ def get_default_performance_options(self) -> PerformanceOptions:
             use_chunkwise_inference=True,
         )
 
+    @override
+    def get_supported_kv_cache_precisions(self) -> tuple[str, ...]:
+        return ("auto", "int8")
```
bejaeger, line 1941: nit: could technically get this from the `TabPFNV3Cache` class but let's keep it like that for now.

src/tabpfn/inference.py
```diff
@@ -894,7 +919,9 @@ def __init__(  # noqa: PLR0913
         )
 
         self.keep_cache_on_device = keep_cache_on_device
-        self.maybe_quantize_kv_cache = maybe_quantize_kv_cache
+        self.kv_cache_precision = resolve_kv_cache_precision(
+            kv_cache_precision, architecture=models[0]
+        )
```
bejaeger: Can we make this a list instead for each model in `models` and below in build_cache use the model_index? Could otherwise bite us in the future. I'd say we then want to adjust the warning in the resolve function to only be thrown once?

bejaeger (approved):
Very nice! thanks for all the changes, looks very clean now.<br>Approving to avoid blocking you but please consider my other comments before merging.

src/tabpfn/architectures/interface.py
jmkuebler: thanks, removed that. It was unecessary to state in the docstring for what we would use that.

src/tabpfn/inference.py
jmkuebler: I moved the resolution now within each cache build, so it is per estimator.

PR #1122, opened 2026-07-16
Enable bf16 autocast on bf16-capable CPUs for inference_precision="auto"

LeoGrin:
## What
Make `inference_precision="auto"` (the default) use **bfloat16 autocast on CPUs with native bf16 acceleration** (Intel AMX / AVX512-BF16, AMD Zen 4+) instead of always running float32 on CPU.

## Why
CPU inference was never tuned for reduced precision — `"auto"` forces float32 on CPU (the code comment noted fp16 autocast "kills inference speed" there). But `torch.autocast("cpu")` runs in **bfloat16**, which modern Intel/AMD server CPUs accelerate in hardware. That's a free ~2× at essentially no accuracy cost.

Measured on a Sapphire Rapids Xeon 8481C (n_train=5000, 20 features, 4 threads, default `n_estimators=8`):

| precision | predict | accuracy | log-loss |
|---|--:|--:|--:|
| float32 (old CPU default) | 67.3s | 0.9940 | 0.0346 |
| **auto → bf16 (this PR)** | **32.4s** | 0.9940 | 0.0347 |

→ **2.08×**, accuracy identical.

## How
- New `_cpu_supports_fast_bf16()` — detects hardware bf16 via torch's CPU capability check (`torch.cpu._is_avx512_bf16_supported()`, the same check vLLM uses; AMX CPUs also enumerate AVX512-BF16, so this single check covers AMX machines too), and additionally requires `torch.backends.mkldnn.is_available()` so torch builds without oneDNN's fast bf16 kernels (e.g. macOS wheels, where CPU bf16 is dramatically *slower* than fp32; Fedora-packaged torch) stay in float32. The capability helper is private torch API, so it is accessed via `getattr` — if a future torch removes it, TabPFN emits a `UserWarning` and stays on float32, so the lost speedup is visible and reportable rather than a silent regression or a crash. Conservative overall: returns `False` when unsure, so `"auto"` **never regresses** on CPUs without fast bf16 (Skylake/Cascade/Ice Lake, AVX2 laptops, macOS). Unlike the earlier `/proc/cpuinfo` approach, this also covers Windows x86 (Zen 4+ / Sapphire Rapids).
- `infer_fp16_inference_mode` → renamed to `infer_autocast_inference_mode` (CPU autocast runs bf16, not fp16), with a CPU branch that enables autocast only when every selected device is a CPU with fast bf16. No deprecation alias: GitHub-wide code search finds no external importer of the old name (only vendored copies of TabPFN). The `enable=True` error message now references `inference_precision="autocast"` (the actual API) instead of a nonexistent `fp16_inference` parameter.
- Uses **autocast** (not a forced dtype), so inputs and numpy preprocessing stay float32 — no interaction with e.g. the fingerprint preprocessing step.

## Notes
- GPU behavior unchanged.
- Escape hatch: pass `inference_precision=torch.float32` to force float32, or a specific `torch.dtype` to force it.
- ARM64 is a potential follow-up: Graviton 3/4 expose a `bf16` cpuinfo flag and aarch64 Linux wheels ship ACL bf16 kernels, but the hardware flag alone doesn't guarantee the installed wheel has fast kernels, so enabling it should be preceded by a Graviton benchmark.

## Tests
- Parametrized unit tests in `tests/test_utils.py`: detection (bf16 hardware / no bf16 hardware / no-oneDNN build / torch helper missing warns) and the CPU autocast branches (auto with/without fast bf16, explicit disable, explicit enable raising on unsupported CPUs).
- Validated end-to-end on-hardware: `auto` → `use_autocast_=True`, no crash with fingerprint on, 2.08×, accuracy unchanged.

### DIFF
--- a/src/tabpfn/classifier.py
+++ b/src/tabpfn/classifier.py
@@ -1725,6 +1725,10 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
             unit="estimator",
             disable=not self.show_progress_bar,
         ):
+            # Upcast from autocast's reduced precision so the post-processing
+            # (temperature scaling, softmax, estimator averaging) runs in
+            # float32, keeping predict_proba consistent with predict_logits.
+            output = output.float()  # noqa: PLW2901
             original_ndim = output.ndim
 
             # This block correctly handles both single configs and lists of configs
--- a/src/tabpfn/utils.py
+++ b/src/tabpfn/utils.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import contextlib
 import os
+import warnings
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Literal
 
@@ -227,16 +228,25 @@ def _cpu_supports_fast_bf16() -> bool:
 
     Requires a torch build with oneDNN, which provides the fast bf16 kernels
     (absent e.g. on macOS wheels, where CPU bf16 falls back to slow reference
-    kernels). Returns False when the capability checks are unavailable, so
-    autocast stays off rather than falling back to emulated, slower bf16.
+    kernels). AMX CPUs also enumerate AVX512-BF16, so this one check covers
+    both instruction sets.
     """
     if not torch.backends.mkldnn.is_available():
         return False
-    # The capability checks are private torch API with no public equivalent;
-    # guarded so a torch release removing them degrades to bf16-off, not a crash.
-    avx512_bf16 = getattr(torch.cpu, "_is_avx512_bf16_supported", lambda: False)
-    amx_tile = getattr(torch.cpu, "_is_amx_tile_supported", lambda: False)
-    return bool(avx512_bf16() or amx_tile())
+    # Private torch API with no public equivalent; if a torch release removes
+    # it, warn and stay on float32 rather than risk slow emulated bf16.
+    avx512_bf16 = getattr(torch.cpu, "_is_avx512_bf16_supported", None)
+    if avx512_bf16 is None:
+        warnings.warn(
+            "torch.cpu._is_avx512_bf16_supported() does not exist in this torch"
+            " version, so TabPFN cannot detect CPU bf16 support and disables"
+            " CPU bf16 autocast. Please report this at"
+            " https://github.com/PriorLabs/TabPFN/issues so detection can be"
+            " updated.",
+            stacklevel=2,
+        )
+        return False
+    return bool(avx512_bf16())
 
 
 def infer_autocast_inference_mode(
--- a/src/tabpfn/utils.py
+++ b/src/tabpfn/utils.py
@@ -226,26 +226,11 @@ def is_autocast_available(device_type: str) -> bool:
 
 @functools.lru_cache(maxsize=1)
 def cpu_supports_fast_bf16() -> bool:
-    """Whether the host CPU has native bfloat16 matmul acceleration.
+    """Whether the CPU accelerates bfloat16 (Intel AMX / AVX512-BF16, AMD Zen 4+).
 
-    ``torch.autocast("cpu")`` runs in bfloat16. That is a meaningful speed-up
-    (~2x on TabPFN inference) only when the CPU has hardware bf16 support
-    (Intel AMX or AVX512-BF16, AMD Zen 4+); on other CPUs bf16 is emulated and
-    can be *slower* than float32. We therefore only report support when such
-    hardware is positively detected, and otherwise stay in float32 so ``"auto"``
-    never regresses.
-
-    Returns:
-        True if the CPU is known to accelerate bfloat16, False when unsure.
+    Reads Linux CPU flags; returns False when they are unavailable (e.g. non-Linux)
+    so autocast stays off rather than falling back to emulated, slower bf16.
     """
-    # 1) PyTorch's own view of the usable CPU ISA. Portable across OSes and
-    #    catches Intel AMX (which reports as "AMX"). Probing must never hard-fail.
-    with contextlib.suppress(Exception):
-        capability = torch.backends.cpu.get_cpu_capability()
-        if isinstance(capability, str) and capability.upper() == "AMX":
-            return True
-    # 2) Linux CPU flags. Catches AVX512-BF16 (Cooper Lake, AMD Zen 4+), which
-    #    (1) only reports coarsely as "AVX512".
     with contextlib.suppress(OSError):
         flags = Path("/proc/cpuinfo").read_bytes()
         if b"amx_bf16" in flags or b"avx512_bf16" in flags:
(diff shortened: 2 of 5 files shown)


src/tabpfn/utils.py
```diff
@@ -234,17 +254,23 @@ def infer_fp16_inference_mode(
...
     """
     is_cpu = any(device.type.lower() == "cpu" for device in devices)
-    fp16_available = (
-        not is_cpu  # CPU can show enabled, yet it kills inference speed
-        and any(is_autocast_available(device.type) for device in devices)
-    )
+    if is_cpu:
+        # CPU autocast runs in bfloat16, which is only faster than float32 on CPUs
+        # with native bf16 support.
+        fp16_available = (
```
jmkuebler: misleading to call this `fp16_available` if we then go for bf16?

jmkuebler: maybe it would also make sense to refator the function name itself?

src/tabpfn/utils.py
```diff
@@ -222,10 +224,28 @@ def is_autocast_available(device_type: str) -> bool:
         )
 
 
+@functools.lru_cache(maxsize=1)
```
jmkuebler: Isn't it a bit over engineered to cache this? 
IIUC there is no need to call this function repeatedly and I think it would make more sense to store the needed information in attributes rather than recalling (if we did so).

tests/test_utils.py
```diff
@@ -269,3 +271,73 @@ def counting_unchunked(*args, **kwargs) -> torch.Tensor:
...
+    mocker.patch("tabpfn.utils.is_autocast_available", return_value=True)
+    with pytest.raises(ValueError, match="does not support it"):
+        infer_fp16_inference_mode([torch.device("cpu")], enable=True)
+
+
+def test__infer_fp16_inference_mode__cpu_explicitly_disabled__returns_false(
+    mocker: MagicMock,
+) -> None:
+    mocker.patch("tabpfn.utils.cpu_supports_fast_bf16", return_value=True)
+    assert infer_fp16_inference_mode([torch.device("cpu")], enable=False) is False
```
jmkuebler:
```suggestion
@pytest.mark.parametrize(
    ("supports_bf16", "enable", "expected"),
    [(True, None, True), (False, None, False), (True, False, False)],
    ids=["auto_with_bf16", "auto_without_bf16", "explicitly_disabled"],
)
def test__infer_fp16_inference_mode__cpu(
    mocker: MagicMock,
    supports_bf16: bool,
    enable: bool | None,
    expected: bool,
) -> None:
    mocker.patch("tabpfn.utils.cpu_supports_fast_bf16", return_value=supports_bf16)
    mocker.patch("tabpfn.utils.is_autocast_available", return_value=True)
    assert infer_fp16_inference_mode([torch.device("cpu")], enable=enable) is expected


def test__infer_fp16_inference_mode__cpu_without_fast_bf16_and_enabled__raises(
    mocker: MagicMock,
) -> None:
    mocker.patch("tabpfn.utils.cpu_supports_fast_bf16", return_value=False)
    mocker.patch("tabpfn.utils.is_autocast_available", return_value=True)
    with pytest.raises(ValueError, match="does not support it"):
        infer_fp16_inference_mode([torch.device("cpu")], enable=True)
```

src/tabpfn/utils.py
```diff
@@ -222,10 +224,28 @@ def is_autocast_available(device_type: str) -> bool:
...
+def cpu_supports_fast_bf16() -> bool:
```
jmkuebler: I was wondering whether this is a reasonable design and found that torch has private functions to check those capabilities https://github.com/pytorch/pytorch/blob/main/torch/cpu/__init__.py

For example vLLM (popular OS LLM serving package) uses those https://github.com/vllm-project/vllm/blob/main/vllm/platforms/cpu.py#L436

More a proposal to consider working with those. Not sure myself what is better here

jmkuebler (commented):
Thanks, that seems like great speedups. I left a few comments, maybe you can quickly see what you want to incorporate before I approve.

tests/test_utils.py
jmkuebler: would make sense to parametrize here?

src/tabpfn/utils.py
LeoGrin: renamed! (also the function name, which is public but not imported or documented anywhere)

src/tabpfn/utils.py
LeoGrin: nice thanks! Does seem better (for instance catch some windows settings we were missing apparently), switched to this :)

src/tabpfn/utils.py
```diff
@@ -233,10 +233,40 @@ def is_autocast_available(device_type: str) -> bool:
...
+    if avx512_bf16 is None:
+        warnings.warn(
+            "torch.cpu._is_avx512_bf16_supported() does not exist in this torch"
+            " version, so TabPFN cannot detect CPU bf16 support and disables"
+            " CPU bf16 autocast. Please report this at"
+            " https://github.com/PriorLabs/TabPFN/issues so detection can be"
+            " updated.",
+            stacklevel=2,
+        )
+        return False
```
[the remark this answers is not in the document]
LeoGrin: From Claude, LGTM:

This is a false positive: `torch.cpu._is_avx512_bf16_supported` was introduced in **torch 2.5.0** — exactly this repo's minimum (`torch>=2.5`) — not 2.11. See [`torch/cpu/__init__.py` at the v2.5.0 tag](https://github.com/pytorch/pytorch/blob/v2.5.0/torch/cpu/__init__.py#L41) (it's absent in v2.4.1, which only had the older `_is_cpu_support_*` names).

Empirical confirmation from this PR's own CI: the `lowest-direct` job installs `torch==2.5.0` and the detection tests pass there — they `patch.object` the attribute without `create=True`, so they would fail with `AttributeError` if the helper didn't exist on that version.

The torch 2.11-era addition is the separate public `torch.cpu.get_capabilities()` API; the private helper predates it by six releases. The warning path here only triggers if a *future* torch removes the private helper, which is exactly the case it's designed to surface.

LeoGrin:
Thanks for the great review @jmkuebler! Your comments should be adressed now.

src/tabpfn/classifier.py
```diff
@@ -1725,6 +1725,10 @@ def forward(  # noqa: C901, PLR0912
             unit="estimator",
             disable=not self.show_progress_bar,
         ):
+            # Upcast from autocast's reduced precision so the post-processing
+            # (temperature scaling, softmax, estimator averaging) runs in
+            # float32, keeping predict_proba consistent with predict_logits.
+            output = output.float()  # noqa: PLW2901
```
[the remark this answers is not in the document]
LeoGrin: Claude speaking, LGTM:

True in the letter but with no observable effect: the public outputs have always been float32 regardless of `inference_precision`. `_predict_proba` ends with `.float().detach().cpu().numpy()` ([classifier.py#L1471](https://github.com/PriorLabs/TabPFN/blob/4d3b388cb14667c50134c2e7c3a0aa7d4f3c5d30/src/tabpfn/classifier.py#L1471)), and `predict_logits` does the same — so with `inference_precision=torch.float64` the fp64 post-processing intermediates were already rounded to fp32 on the way out. Computing those intermediates in fp32 instead changes the final float32 numbers by at most ~1 ulp.

This also mirrors the regressor, which has applied the identical unconditional `output.float()` right after `iter_outputs` all along ([regressor.py#L1353](https://github.com/PriorLabs/TabPFN/blob/4d3b388cb14667c50134c2e7c3a0aa7d4f3c5d30/src/tabpfn/regressor.py#L1353)), including for fp64 users. Keeping the cast unconditional keeps the two estimators consistent.

jmkuebler (approved):
Thanks for making another pass. LGTM

PR #1138, opened 2026-07-24
Fix v2.x activation checkpointing regression from #1070

Nitasurin:
## Summary

PR #1070 changed the v2.x transformer block input from a `Tensor` to a
single-element mutable list. Each block consumes the activation using `pop(0)`,
allowing the caller to release its reference before the block creates another
large activation.

This optimization is incompatible with activation checkpointing during
fine-tuning, with different symptoms across the v2.x architectures.

### Failure modes

- **v2 silently loses transformer gradients.**

  After #1070, v2 passes the mutable list to `checkpoint` without explicitly
  setting `use_reentrant`. Under the currently supported PyTorch runtime, this
  selects the legacy reentrant checkpoint implementation.

  Reentrant checkpointing does not treat tensors nested inside Python
  containers as autograd inputs. Backward can therefore complete without an
  exception while parameters inside the checkpointed transformer blocks receive
  no gradients.

- **v2.5 and v2.6 fail during backward recomputation.**

  These architectures explicitly use non-reentrant checkpointing. During
  backward, checkpointing recomputes the checkpointed block. The mutable list is
  retained as a non-Tensor argument, but the original forward has already
  emptied it with `pop(0)`. Recalculation therefore raises:

  `IndexError: pop from empty list`

## Fix

Restore the transformer activation as a direct, top-level `Tensor` input for
v2, v2.5, and v2.6 blocks and pass that Tensor directly across checkpoint
boundaries.

This restores the checkpoint input contract and fixes both failure modes:

- the activation is again exposed to checkpointing and autograd as a top-level
  Tensor input;
- backward recomputation no longer depends on destructively consumed mutable
  state.

Separately, v2 now explicitly selects `use_reentrant=False`. Restoring the
top-level Tensor is the actual regression fix; selecting non-reentrant
checkpointing is not required once the Tensor input has been restored. It is an
intentional additional change that follows PyTorch's recommended checkpoint
implementation and matches v2.5 and v2.6.

## Changes

- Restore `torch.Tensor` as the transformer block input in v2, v2.5, and v2.6.
- Remove the single-element activation containers and destructive `pop(0)`
  calls.
- Pass activation tensors directly in normal, KV-cache, and checkpointed block
  calls.
- Explicitly use `use_reentrant=False` for the v2 checkpoint path.
- Extend the checkpoint tests to run `backward()`.
- Assert that a parameter in the first transformer block receives a gradient.

## User impact

Before this fix:

- v2 fine-tuning with activation checkpointing could complete backward while
  silently leaving checkpointed transformer-block parameters without gradients;
- v2.5 and v2.6 fine-tuning with activation checkpointing could fail during
  backward recomputation with `IndexError: pop from empty list`.

After this fix, all three architectures complete backward and propagate
gradients through their transformer blocks.

## Trade-off

This change removes the activation-reference release optimization introduced by
the state-container portion of #1070. Because direct Tensor inputs are restored
for every block call, this also affects non-checkpointed inference and KV-cache
paths, whose peak activation memory may return toward the behavior before
#1070.

The unrelated improvements from #1070 remain intact, including:

- contiguous-input handling in chunked evaluation;
- the non-contiguous-input regression test;
- the shared scaled-dot-product-attention implementation;
- Flash Attention and MPS-related changes.

This PR prioritizes correct fine-tuning and a stable checkpoint input contract.
A checkpoint-safe version of the inference reference-release optimization can
be considered separately.

## Validation

- `pytest tests/test_architectures/test_tabpfn_v2.py tests/test_architectures/test_tabpfn_v2_5.py tests/test_architectures/test_tabpfn_v2_6.py -q`
  - 34 passed
- Checkpointed and non-checkpointed forward outputs remain equivalent.
- Backward completes for v2, v2.5, and v2.6.
- Transformer-block gradient propagation is covered by regression tests.
- `ruff check` on all six changed Python files passed.
- `ruff format --check` on all six changed Python files passed.

### DIFF
--- a/src/tabpfn/architectures/tabpfn_v2_5.py
+++ b/src/tabpfn/architectures/tabpfn_v2_5.py
@@ -382,7 +382,7 @@ class TabPFNBlock(nn.Module):
     @override
     def forward(
         self,
-        x_BRCE_container: list[torch.Tensor],
+        x_BRCE: torch.Tensor,
         single_eval_pos: int,
         save_peak_memory_factor: int | None,
         *,
@@ -399,9 +399,9 @@ class TabPFNBlock(nn.Module):
         E: The embedding size of each cell.
 
         Args:
-            x_BRCE_container:
-                A length-1 list containing the transformer state passed as input to the
-                layer of shape (batch_size, num_items, num_feature_blocks, d_model).
+            x_BRCE:
+                The transformer state passed as input to the layer of shape
+                (batch_size, num_items, num_feature_blocks, d_model).
             single_eval_pos:
                 The position from which on everything is treated as test set.
             save_peak_memory_factor:
@@ -423,7 +423,6 @@ class TabPFNBlock(nn.Module):
         """
         # -- First Block: Attention between features.
         # The row attention has no train/test distinction and is not cached.
-        x_BRCE = x_BRCE_container.pop(0)
         x_BRCE = chunked_evaluate_maybe_inplace(
             self.per_sample_attention_between_features,
             x_BRCE,
@@ -830,13 +829,9 @@ class TabPFNV2p5(Architecture):
 
         kv_out: dict[int, KVCacheEntry] = {}
         for layer_idx, block in enumerate(self.blocks):
-            # Note: Using x_BRCD._set() instead of the container approach here leads
-            # to memory issues on some mac hardware.
-            x_BRCD_container = [x_BRCD]
-            del x_BRCD
             if return_kv_cache and not using_cache:
                 x_BRCD, kv_entry = block(
-                    x_BRCD_container,
+                    x_BRCD,
                     block_single_eval_pos,
                     save_peak_memory_factor,
                     return_kv=True,
@@ -844,7 +839,7 @@ class TabPFNV2p5(Architecture):
                 kv_out[layer_idx] = kv_entry
             elif using_cache:
                 x_BRCD, _ = block(
-                    x_BRCD_container,
+                    x_BRCD,
                     block_single_eval_pos,
                     save_peak_memory_factor,
                     cached_kv=kv_cache.kv[layer_idx],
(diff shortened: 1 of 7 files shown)


TuanaCelik:
Thank you for the contribution @Nitasurin - I'll try to find the best person to review this. @oscarkey any ideas?

src/tabpfn/architectures/tabpfn_v2_5.py
```diff
@@ -382,7 +382,7 @@ def __init__(
     @override
     def forward(
         self,
-        x_BRCE_container: list[torch.Tensor],
+        x_BRCE: torch.Tensor,
```
priorphil: To keep the best of both worlds, could we make this
`x_BRCE_container: list[torch.Tensor] | torch.Tensor`

and branch below as `x_BRCE = x_BRCE_container if isinstance(x_BRCE_container, torch.Tensor) else x_BRCE_container.pop(0)`?
 Then we'd pass the raw tensor in the checkpointing branch below and the list in the non-checkpointing branch.

priorphil (changes requested):
Thanks for the contribution and sorry for breaking this!

priorphil (approved):
Thanks!

src/tabpfn/architectures/tabpfn_v2_5.py
Nitasurin: Thanks for the suggestion — I agree that this is the better balance.

I had considered separating the checkpoint and non-checkpoint paths earlier. However, in order to restore fine-tuning correctness as quickly as possible and minimize the risk of introducing another regression, I initially chose the most conservative approach: directly reverting the container change across the affected paths.

After reconsidering the trade-off, I have updated the PR in commit `4595bfc`. The v2.x transformer blocks now accept either the existing single-element list or a Tensor, while keeping the existing parameter name for compatibility. Checkpointed calls pass the Tensor directly with `use_reentrant=False`; normal and KV-cache calls continue to use the list-based path, preserving the reference-release optimization introduced in #1070.

This change is applied consistently to v2, v2.5, and v2.6. The relevant architecture tests pass locally: 34 passed. Ruff lint and format checks also pass with the project-pinned version.

PR #1140, opened 2026-07-27
Fix fit() crashes in tuning: rare class missing from holdout, Generator random_state

LeoGrin:
## Summary

Two crashes in the `tuning_config` path of `TabPFNClassifier.fit()`, both found by auditing `main`:

**1. Temperature calibration / threshold tuning crashes when a rare class is absent from the tuning holdout.**
`find_optimal_temperature` called `log_loss(y_true, probas)` without `labels=`. `probas` always has `n_classes` columns, but the holdout labels come from only the first `tuning_n_folds` folds of the `StratifiedKFold`, so a rare class can be entirely absent and sklearn raises `ValueError: y_true and y_prob contain different number of classes`. Repro: 25,002 samples with a 2-sample class and `tuning_config={'calibrate_temperature': True}` (auto-resolves to 1 of 3 folds).

Notably the failure is **deterministic**, not seed-dependent: `StratifiedKFold`'s per-fold class allocation depends only on the class counts, `shuffle`/`random_state` only permute which samples fill each fold's quota. A scheduled retrain can start crashing because two rows of a new rare class were added.

The `log_loss` entry of `METRIC_NAME_TO_OBJECTIVE` (used by threshold tuning on binarized one-vs-rest labels) had the same missing-`labels` pattern and failed with `y_true contains only one label`; fixed with `labels=[0, 1]`.

**2. `fit()` crashes when `random_state` is a `np.random.Generator` and tuning is enabled.**
`get_tuning_splits` forwarded `random_state` raw to `StratifiedKFold`, whose `check_random_state` rejects `np.random.Generator` — even though `Generator` is a documented `random_state` type and works everywhere else in `fit()`. Now converted to a static seed via `infer_random_state`, matching what `fit()` itself does.

## Behavior impact

None where the code previously worked:
- With every class present in the holdout, sklearn infers `labels = np.unique(y_true) = np.arange(n_classes)` (y is label-encoded before the tuning call), so the explicit `labels=` is bit-for-bit identical — verified, including the chosen temperature.
- The `Generator` conversion only touches the `isinstance(random_state, np.random.Generator)` case, which previously always crashed; `int`/`RandomState`/`None` are passed through unchanged.

## Tests

Three regression tests in `tests/test_inference_tuning.py`, matching the file's unit-test style (no model needed). All three fail on `main` and pass with this change; the 24 existing tests in the file still pass.

### DIFF
--- a/src/tabpfn/inference_tuning.py
+++ b/src/tabpfn/inference_tuning.py
@@ -21,6 +21,8 @@ from sklearn.metrics import (
 )
 from sklearn.model_selection import StratifiedKFold
 
+from tabpfn.utils import infer_random_state
+
 if TYPE_CHECKING:
     import torch
 
@@ -119,7 +121,8 @@ METRIC_NAME_TO_OBJECTIVE = {
             y_pred,
         )
     ),
-    "log_loss": log_loss,
+    # y_true is binarized one-vs-rest and may contain a single label.
+    "log_loss": lambda y_true, y_pred: log_loss(y_true, y_pred, labels=[0, 1]),
 }
 
 
@@ -168,6 +171,10 @@ def get_tuning_splits(
     # more than 100 folds
     rounded_holdout_frac = round(holdout_frac, 2)
     n_folds = max(2, round(1 / rounded_holdout_frac))
+
+    if isinstance(random_state, np.random.Generator):
+        # StratifiedKFold does not accept np.random.Generator.
+        random_state, _ = infer_random_state(random_state)
 
     splitter = StratifiedKFold(
         n_splits=n_folds,
@@ -329,7 +336,12 @@ def find_optimal_temperature(
     # TODO: think about vectorizing this loop.
     for temperature in temperatures:
         probas = logits_to_probabilities_fn(raw_logits, temperature)
-        current_log_loss = log_loss(y_true=y_true, y_pred=probas)
+        # The holdout may lack a rare class, so labels cannot be inferred from y_true.
+        current_log_loss = log_loss(
+            y_true=y_true,
+            y_pred=probas,
+            labels=np.arange(probas.shape[-1]),
+        )
 
         if current_log_loss < best_log_loss:
             best_log_loss = current_log_loss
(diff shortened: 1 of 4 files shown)


src/tabpfn/inference_tuning.py
```diff
@@ -119,7 +121,8 @@ class ClassifierEvalMetrics(str, Enum):
             y_pred,
         )
     ),
-    "log_loss": log_loss,
+    # y_true is binarized one-vs-rest and may contain a single label.
+    "log_loss": lambda y_true, y_pred: log_loss(y_true, y_pred, labels=[0, 1]),
```
bejaeger, line 129: what if we deal with multiclass? should we take the labels from y_true unless it's a single label?

bejaeger (changes requested):

src/tabpfn/inference_tuning.py
LeoGrin: I think we assume in this dict that it's binary?

from Claude: 

multiclass can't reach this entry.
METRIC_NAME_TO_OBJECTIVE is only consumed by compute_metric_to_minimize, which has a single caller: find_optimal_classification_threshold_single_class (L249). 
That's inside the OvR loop, so y_true is always (y_true == i).astype(int) and y_pred is (probas >= threshold).astype(int) — both {0, 1} by construction.
The rest of the dict assumes the same thing: f1 uses average="binary" and roc_auc takes a 1-D score vector. Both raise on multiclass input, so labels=[0, 1] isn't adding an assumption,  it's making the dict's existing one explicit in the one spot where leaving it implicit crashed.

bejaeger: What about renaming `METRIC_NAME_TO_OBJECTIVE -> BINARY_METRIC_NAME_TO_OBJECTIVE` in that case?

bejaeger (dismissed):

src/tabpfn/inference_tuning.py
bejaeger: Thanks!

bejaeger (approved):
LGTM!

PR #1142, opened 2026-07-28
Expose ManyClassDecoder.attention_weights for decoder-head readout

eliott-kalfon:
## What

Adds `ManyClassDecoder.attention_weights(train_embeddings, test_embeddings)` to the v3 architecture: the canonical per-training-row attention distribution of the multiclass decoder head, shape `(B, M, N)`, averaged over heads and summing to 1 over the training axis.

## Why

The multiclass head classifies by attending each test row over the training rows and averaging their one-hot labels weighted by that attention. The prediction is therefore a weighted vote, and reading out *which* training rows drive a prediction (and by how much) is a natural interpretability primitive.

Today that read-out can only be obtained by **reimplementing the head's forward pass externally** (q/k projections, optional `softmax_scaling_layer`, `scores / sqrt(head_dim)`, softmax over train rows, mean over heads). `forward` itself computes this distribution implicitly inside a fused flash-attention kernel that never materializes it ([tabpfn_v3.py](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/architectures/tabpfn_v3.py)), so downstream code has to duplicate internal math that can silently drift out of sync.

This method makes the decoder own that computation, so tooling reads the weights from the canonical source instead of re-deriving them.

## Design

- One method on `ManyClassDecoder`; nothing on the hot inference path changes.
- Naturally optional: it exists only on multiclass models that have this decoder. Regression and other heads simply don't expose it, so callers gate on its presence.
- Mirrors `forward`'s scoring exactly. Collapsing the returned weights by training label reproduces `forward`'s pre-log class average, and hence its logits up to the head's log-clamping.

## Tests

`test__many_class_decoder_attention_weights_matches_forward` (with and without `SoftmaxScalingMLP`) asserts the weights form a proper distribution and that collapsing them by class label, then applying the head's log transform, reproduces `forward`'s logits. Verified separately on a fitted `TabPFNClassifier` that it matches the existing external reimplementation bit-for-bit.

### DIFF
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -557,6 +557,43 @@ class ManyClassDecoder(nn.Module):
         # convert to logits:
         return torch.log(torch.clamp(test_output_MBT, min=1e-5) + 3e-5)
 
+    def attention_weights(
+        self,
+        train_embeddings: torch.Tensor,  # (B, N, E)
+        test_embeddings: torch.Tensor,  # (B, M, E)
+    ) -> torch.Tensor:
+        """Per-train-row attention weights, averaged over heads: ``(B, M, N)``.
+
+        Runs the same query/key projection, optional query scaling and
+        scaled-dot-product softmax over the training rows as ``forward``, but
+        returns the attention distribution itself rather than the label-weighted
+        average. ``weights[..., n]`` is the vote mass this decoder places on
+        train row ``n`` for a test row; the weights are non-negative and sum to 1
+        over the training axis.
+
+        ``forward`` produces the same distribution implicitly inside a fused
+        attention kernel that never materializes it, so this recomputes the
+        scores explicitly. Collapsing the weights by training label recovers the
+        pre-log class average that ``forward`` turns into logits. Provided for
+        interpretability read-out of the decoder head; not used on the inference
+        path.
+        """
+        B, M, _ = test_embeddings.shape
+        N = train_embeddings.shape[1]
+        q_BMHD = self.q_projection(test_embeddings).view(
+            B, M, self.num_heads, self.head_dim
+        )
+        if train_embeddings.dtype != q_BMHD.dtype:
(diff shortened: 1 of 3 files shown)


src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -557,6 +563,32 @@ def forward(
         # convert to logits:
         return torch.log(torch.clamp(test_output_MBT, min=1e-5) + 3e-5)
 
+    def attention_weights(
+        self,
+        train_embeddings: torch.Tensor,  # (B, N, E)
+        test_embeddings: torch.Tensor,  # (B, M, E)
+    ) -> torch.Tensor:
+        """Per-train-row attention weights, averaged over heads: ``(B, M, N)``.
```
priorphil: Can we shorten this doc string?

priorphil (dismissed):

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -557,6 +563,28 @@ def forward(
...
+
+        ``weights[..., n]`` is the vote mass placed on train row ``n`` for a
+        test row; non-negative and summing to 1 over the training axis.
+        Collapsing by training label recovers the pre-log class average that
+        ``forward`` turns into logits.
+
+        ``forward`` fuses this into a single attention kernel to avoid
+        materializing an O(N*M) tensor.
```
priorphil:
```suggestion
        """Per-train-row attention weights, averaged over heads: `(B, M, N)`.

        `weights[..., n]` is the vote mass placed on train row `n` for a
        test row; non-negative and summing to 1 over the training axis.
        Collapsing by training label recovers the pre-log class average that
        `forward` turns into logits.

        `forward` fuses this into a single attention kernel to avoid
        materializing an O(N*M) tensor.
```

priorphil (dismissed):

priorphil (approved):

PR #1143, opened 2026-07-28
Re-enable macos tests with minimum dependency versions.

oscarkey:
This requires:
- Bumping lightgbm to >=4.4.0, the first version with a macos arm64 wheel, so the lowest-direct resolution works. This was [released](https://pypi.org/project/lightgbm/4.4.0/#history) in June 2024, which is newer than our usual aim of at least 3 years ago. There are no effects on the torch, pandas, numpy, or sklearn versions, which are often the main issues.
- Disabling some preprocessing consistency tests in the case of minimum dependency versions


jmkuebler (dismissed):
LGTM, thank you. Please check Codex comment before merging

oscarkey:
fixed, and tidied slightly. I'm afraid it requires approval again.

jmkuebler (approved):

PR #1151, opened 2026-07-30
Try to deflake windows.

priorphil:
We occasionally get `Windows fatal exception: Windows fatal exception: code 0xcode 0xc000001dc000001d`. Let's see if disabling AVX512 solves this.

Update from Claude after debugging:

Root cause is settled, and the fix is one environment variable.

It's oneDNN's AMX kernels. All three oneDNN caps survive; capping MKL or ATen still aborts. Since AVX512_CORE_FP16 is the highest oneDNN tier below AMX, the only thing it removes versus the default is the AMX kernel set — that isolates the fault to an AMX instruction. Which matches the hardware exactly: Emerald Rapids advertises AMX in CPUID, but Windows keeps the tile XSTATE disabled until a process opts in, so the first ldtilecfg takes a #UD → 0xc000001d. The AMD EPYC runners have no AMX and never fault.

Two details worth noting from the log. This crash was on torch 2.5.0 / oneDNN v3.5.3, while the earlier one was torch 2.13.0 / oneDNN v3.12.0 — same CPU model, same fault, seven minor oneDNN versions apart. So it won't age out by bumping torch. And ONEDNN_VERBOSE=1 produced no output at all before the abort; oneDNN's verbose goes through buffered stdio, so the buffer died with the process. The exit-code table carried the result instead.
I
 kept the CPU/dispatch logging — it's ~2s and it's what made this diagnosable — but it's easy to drop if you'd rather have the quieter log.

One caveat on the value: if a future oneDNN ever rejects AVX512_CORE_FP16, it warns and falls back to unrestricted, and the crash returns silently. AVX512_CORE also survived and is the more conservative spelling if you'd prefer belt-and-braces over the last bit of AVX-512 kernel coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,24 +125,34 @@ jobs:
           key: model-cache-${{ hashFiles('model_cache/**') }}
           enableCrossOsArchive: true
 
-      # Diagnostics for the intermittent Windows abort
-      # `Windows fatal exception: code 0xc000001d` (STATUS_ILLEGAL_INSTRUCTION),
-      # which kills roughly 1 in 10 Windows jobs. Every occurrence dies at the
-      # same test -- the suite's first fit/predict on real weights -- under both
-      # dependency sets, and a retry clears it. faulthandler cannot walk the
-      # native stack on Windows ("cannot get C stack on this system"), so the
-      # log never names the faulting DLL; record the CPU and the ISA torch
-      # picks for it instead.
+      # Works around `Windows fatal exception: code 0xc000001d`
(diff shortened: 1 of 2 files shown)


oscarkey (dismissed):
are we worried about breaking on real windows machines with this enabled?

priorphil:
@oscarkey this should be ready for another round of review. Re your question: This probably actually breaks users of TabPFN, I've updated the README with this. Will try to file a bug report with torch if I can manage to extract meaningful signal.

oscarkey (approved):

PR #1158, opened 2026-08-03
Add workflow that re-approves PRs from team members.

oscarkey:
We now require the most recent push to be approved, to avoid external PRs merging unreviewed code. This is too much overhead for internal PRs, so re-approve them automatically.

It's possible this workflow might need the `issues: read` permission to read the timeline, but that's unclear. So try without it first.

Manual testing: merge this, verify that it approves for internal PRs but not for PRs from a fork

### DIFF
--- a/.github/workflows/reapprove-internal-prs.yml
+++ b/.github/workflows/reapprove-internal-prs.yml
@@ -1,0 +1,118 @@
+# Automatically re-approve PRs from PriorLabs org members when new code is pushed.
+#
+# The branch protection rules require that the most recent push is reviewed and
+# approved, which is important for PRs from external collaborators. However, this is
+# extra overhead for PRs from Prior Labs members, where we might want to request small
+# changes during review and trust that the author will apply these. In this case,
+# approve the current version of the PR if a previous version was already approved.
+name: Re-approve internal PRs
+
+on:
+  pull_request:
+    types: [synchronize]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  reapprove:
+    name: Restore dismissed approval
+    # MEMBER means the author is a member of the organisation that owns the repository.
+    # Also check if the PR came from a fork, to avoid approving PRs from member's forks,
+    # and for defence in depth.
+    if: >
+      (
+        github.event.pull_request.author_association == 'MEMBER' ||
+        github.event.pull_request.author_association == 'OWNER'
+      ) &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-slim
+    steps:
+      - name: Restore approval dismissed by this push
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const pullNumber = context.payload.pull_request.number;
+
+            const isHuman = (user) => user?.type === "User";
+
+            // The push that triggered this run also dismisses the existing approvals,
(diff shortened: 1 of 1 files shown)


.github/workflows/reapprove-internal-prs.yml
```diff
@@ -0,0 +1,113 @@
...
+    # - The PR does not come from a fork, for defence in depth.
+    if: >
+      (
+        github.event.pull_request.author_association == 'MEMBER' ||
+        github.event.pull_request.author_association == 'OWNER'
+      ) &&
+      (
+        github.event.review.author_association == 'MEMBER' ||
+        github.event.review.author_association == 'OWNER' ||
+        github.event.review.user.login == 'github-actions[bot]'
```
brendan-priorlabs, line 45: What is this part here for?

.github/workflows/reapprove-internal-prs.yml
```diff
@@ -0,0 +1,113 @@
...
+            const reviews = await github.paginate(github.rest.pulls.listReviews, {
+              owner,
+              repo,
+              pull_number: pullNumber,
+              per_page: 100,
+            });
+
+            // The head can already be approved, either by a human or by an earlier run
+            // of this workflow.
+            if (reviews.some((r) => r.state === "APPROVED" && r.commit_id === headSha)) {
```
brendan-priorlabs: I'm a bit worried about this part... Isn't it possible to have two different PRs with the same head SHA, but then, subsequently different commits afterwards? Maybe we can use the PR id instead?

.github/workflows/reapprove-internal-prs.yml
```diff
@@ -0,0 +1,113 @@
...
+              .filter((e) => e.event === "review_dismissed")
+              .map((e) => e.dismissed_review)
+              .find((d) => d?.review_id === dismissedReview.id);
+            if (!dismissal) {
+              core.info(`Found no dismissal of review ${dismissedReview.id}.`);
+              return;
+            }
+
+            // The dismissed review may not have been an approval (e.g. if it requested
+            // changes), in which case we don't want to approve.
```
brendan-priorlabs, line 91: Lot of good cases in here worth manual testing

brendan-priorlabs (commented):
Thanks for this @oscarkey! Looking pretty good overall. Just a couple small comments

.github/workflows/reapprove-internal-prs.yml
oscarkey: This causes the bot to re-approve after a second code push. i.e.
1. Reviewer approves
2. Author pushes new code, which dismisses 1
3. Bot approves
4. Author pushes new code, which dismisses 3
5. Bot approves

Without this line, (5) will be skipped, because (3) was made by a bot, not a member or an owner.
I added a comment to clarify.

.github/workflows/reapprove-internal-prs.yml
oscarkey: I think that is possible, however `reviews` is already filtered by the PR id. This statement is using `headSha` to check if any of the reviews of the PR are for the latest commit. I tried to clarify.

.github/workflows/reapprove-internal-prs.yml
oscarkey: yes definitely. Claude claims this workflow will be live even in this PR, which would be convenient, but I can't find anything in the docs about this...

brendan-priorlabs (approved):
LGTM. Thanks for following up on those comments!

PR #1159, opened 2026-08-05
Warn at fit when X contains likely-text columns

vahidblm:
## Issue
No GitHub issue. This was driven by repeated reports of users fitting on free-text columns, which leads to silently worse results.

## Motivation

High-cardinality string columns that cannot be parsed as numbers are labelled `FeatureModality.TEXT` by modality detection, but nothing used this label: they were silently swept into the same `OrdinalEncoder` as real categoricals (which selects columns by dtype), turning near-unique text into near-unique integer codes, without any warnings.

`fit()` now emits a `UserWarning` naming the affected columns and pointing at the fixes: convert to a numeric dtype (numbers stored as strings), or use [tabpfn-client](https://github.com/PriorLabs/tabpfn-client) if it's a genuine text, or pass the column in `categorical_features_indices` if it's a real high-cardinality category (in this case, there won't be warnings).

The check runs inside `detect_feature_modalities` (the single place the TEXT labels are produced, before `clean_data` rebuilds the schema). Detection logic itself is not changed (purely diagnostic).

---

## Public API Changes

-   [x] No Public API changes (new warning only; no signatures changed)

---

## How Has This Been Tested?

- Unit tests in `tests/test_preprocessing/test_modality_detection.py`: free text warns; low-cardinality strings and fully numeric strings do not warn; a numeric column with one stray "N/A" token warns; columns declared categorical (both plain string and `category` dtype) do not warn.
- One parametrized estimator-level test (classifier and regressor) in `tests/test_preprocessing/test_modality_detection.py`: `fit` warns and names the column, `predict` stays quiet, and `categorical_features_indices` silences it.
- `stacklevel=6` so the warning points at the caller's `fit()` line (the `@config_context(...)` decorator on `fit` adds a contextlib wrapper frame, on top of the `detect_feature_modalities` frame); pinned by `warning.filename` asserts in the tests.

---

## Checklist

-   [x] Documentation has been updated (if the public API or usage changes) [no update is need, as this is just a warning].
-   [x] A changelog entry has been added (see `changelog/README.md`).


src/tabpfn/validation.py
```diff
@@ -155,6 +161,60 @@ def validate_dataset_size(
...
+        f"high-cardinality categoricals, which usually adds noise rather than "
+        f"signal: {columns}.\n"
+        "If such a column holds numbers stored as strings, convert it to a numeric "
+        "dtype. If it holds genuine text, this package has no text handling -- "
+        "consider the tabpfn-client API, which embeds text natively: "
+        "https://github.com/PriorLabs/tabpfn-client \n"
+        "To silence this for a column that is genuinely a high-cardinality category, "
+        "pass its index in `categorical_features_indices`.",
+        UserWarning,
+        stacklevel=4,
```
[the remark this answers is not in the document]
vahidblm: This is fixed now, changed to `stacklevel=5`.

changelog/1159.added.md
```diff
@@ -0,0 +1 @@
+`fit()` now warns when a column of `X` looks like free text: a string column with more than 30 distinct values that cannot be parsed as numbers. TabPFN has no text understanding, so such columns were silently encoded as near-unique integer codes, which adds noise and hurts performance. The warning names the affected columns and points to the fixes: convert the column to a numeric dtype if it holds numbers stored as strings, use the [tabpfn-client](https://github.com/PriorLabs/tabpfn-client) API if it holds genuine text, or pass the column's index in `categorical_features_indices` if it really is a high-cardinality category (which also silences the warning).
```
bejaeger: I would say ``fit()` now warns when a column of `X` looks like free text.` is enough. And I'd avoid stating current settings like "30 distinct values" in the changelog that might change over time.

src/tabpfn/validation.py
```diff
@@ -155,6 +161,64 @@ def validate_dataset_size(
...
+    text_names = [
+        feature.name.removeprefix(INPUT_FEATURE_PREFIX)
+        for index, feature in enumerate(feature_schema.features)
+        if feature.modality is FeatureModality.TEXT and index not in declared
+    ]
+    if not text_names:
+        return
+
+    shown = text_names[:_MAX_TEXT_COLUMNS_IN_WARNING]
+    columns = ", ".join(repr(name) for name in shown)
```
bejaeger: nit: consider `column_names_to_print`

src/tabpfn/classifier.py
```diff
@@ -729,6 +728,12 @@ def _initialize_dataset_preprocessing(
             max_unique_for_category=self.inference_config_.MAX_UNIQUE_FOR_CATEGORICAL_FEATURES,
             min_unique_for_numerical=self.inference_config_.MIN_UNIQUE_FOR_NUMERICAL_FEATURES,
         )
+        # Must stay before `clean_data`: the TEXT labels do not survive the first
+        # preprocessing step that rebuilds the schema.
+        warn_if_text_features(
+            feature_schema,
+            declared_categorical_indices=self.categorical_features_indices,
+        )
```
bejaeger: what about making this part of `detect_feature_modalities` itself?
Advantages: 1) a single place where we call this function. 2) we can add a single test to the `detect_feature_modalities` tests and don't need two separate tests for the classifier and regressor?

bejaeger (changes requested):
Thanks @vahidblm ! Looks mostly good to me, please see my 2 comments.

changelog/1159.added.md
vahidblm: Done!

src/tabpfn/validation.py
vahidblm: Fixed!

src/tabpfn/classifier.py
vahidblm: Good point! I added everything into `detect_feature_modalities` and simplified the tests too (so now all the tests are in `test_modality_detection`

bejaeger: Thanks!

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -75,7 +81,73 @@ def detect_feature_modalities(
...
+    # labels are produced, and they do not survive the first preprocessing step
+    # that rebuilds the schema.
+    warn_if_text_features(
+        feature_schema,
+        declared_categorical_indices=provided_categorical_indices,
+    )
+    return feature_schema
+
+
+def warn_if_text_features(
```
bejaeger: nit: can be made private

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -75,7 +81,73 @@ def detect_feature_modalities(
             big_enough_n_to_infer_cat=big_enough_n_to_infer_cat,
         )
         features.append(Feature(name=feature_name, modality=feat_modality))
-    return FeatureSchema(features=features)
+    feature_schema = FeatureSchema(features=features)
+    # Warn here rather than at each call site: this is the single place the TEXT
+    # labels are produced, and they do not survive the first preprocessing step
+    # that rebuilds the schema.
```
bejaeger: nit: I guess this comment refers to the previous state of the code base and is not really needed

bejaeger (dismissed):
Thanks for the two changes!<br>Two more nits, but otherwise LGTM! approving already

src/tabpfn/preprocessing/modality_detection.py
vahidblm: Done

src/tabpfn/preprocessing/modality_detection.py
vahidblm: Done

LeoGrin (approved):
LGTM!

PR #1162, opened 2026-08-06
Cache built models in load_model (opt-in)

ggprior:
This is a nice-to-have in server contexts where the model state dict is ideally loaded once and then should stay resident. Otherwise weight loading (in `.fit()`) dominates small inferences.
I'm flexible about the exact design, what's proposed herein is an opt-in LRU cache stored on the module level. Opt-in is gated on `TABPFN_MODEL_CACHE_SIZE`

<details>
<summary>Claude:</summary>

load_model rebuilds the architecture and re-runs load_state_dict on every call — only the raw checkpoint bytes are cached (_load_checkpoint_cached), never the built module. Any workload that reloads the same checkpoint (cross-validation, per-group models, serving) pays that rebuild each time; in a serving setup it was ~40% of warm latency.

Add an opt-in LRU of built models keyed by (path, file identity), behind TABPFN_MODEL_CACHE_SIZE (default 0 = off, behaviour unchanged). Only the non-mutating build (cache_trainset_representation=False) is cached; the caching path always rebuilds, so a shared instance never carries a prior fit's train-set state.

Tests: cache hit reuses the model, the mutating build is never cached, off by default, LRU eviction, and a signature tripwire so the cache key is revisited if load_model gains a parameter.

</details>

### DIFF
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -17,10 +17,12 @@ import tempfile
 import urllib.request
 import warnings
 import zipfile
+from collections import OrderedDict
 from dataclasses import asdict, dataclass
 from enum import Enum
 from importlib import import_module
 from pathlib import Path
+from threading import Lock
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast, overload
 from urllib.error import URLError
 
@@ -851,6 +853,34 @@ def _load_checkpoint_cached(path: str, _identity: tuple[int, int]) -> dict:
     return Checkpoint(path).load()
 
 
+# Bounded, opt-in cache of *built* models (architecture + loaded weights),
+# keyed by (resolved path, file identity). Enabled by setting the env var
+# ``TABPFN_MODEL_CACHE_SIZE`` to a positive integer (an LRU of that size;
+# default 0 disables it, preserving prior behaviour). Only the non-mutating
+# build is cached: with ``cache_trainset_representation`` the model accumulates
+# the train-set representation during fit, so a shared instance can't be reused
+# across fits. The cached model is shared by reference and left in ``eval()``
+# mode — intended for repeated sequential fit/predict (cross-validation,
+# per-group models, or servers that manage their own concurrency). RES-2422
+# tracks the follow-up that externalises per-fit state so a single backbone can
+# be shared across threads too.
+_BUILT_MODEL_CACHE: OrderedDict[tuple[str, tuple[int, int]], tuple] = OrderedDict()
+_BUILT_MODEL_CACHE_LOCK = Lock()
+
+
+def _built_model_cache_size() -> int:
+    try:
+        return max(0, int(os.environ.get("TABPFN_MODEL_CACHE_SIZE", "0")))
+    except ValueError:
+        return 0
+
+
+def clear_built_model_cache() -> None:
+    """Drop all entries from the built-model cache (see ``TABPFN_MODEL_CACHE_SIZE``)."""
+    with _BUILT_MODEL_CACHE_LOCK:
+        _BUILT_MODEL_CACHE.clear()
+
+
 def load_model(
     *,
     path: Path,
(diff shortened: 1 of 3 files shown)


ggprior:
cc @jmkuebler

src/tabpfn/model_loading.py
```diff
@@ -863,17 +893,57 @@ def load_model(
...
+    )
+
+    if use_cache:
+        size = _built_model_cache_size()
+        with _BUILT_MODEL_CACHE_LOCK:
+            _BUILT_MODEL_CACHE[key] = result
+            _BUILT_MODEL_CACHE.move_to_end(key)
+            while len(_BUILT_MODEL_CACHE) > size:
+                _BUILT_MODEL_CACHE.popitem(last=False)
+    return result
```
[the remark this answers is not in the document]
jmkuebler: Maybe we can just add a check that fine-tuning needs to ensure the checkpoint is not cached? On the other hand, since it is opt-in, even now the risk seems small

src/tabpfn/model_loading.py
```diff
@@ -851,6 +853,34 @@ def _load_checkpoint_cached(path: str, _identity: tuple[int, int]) -> dict:
...
+# across fits. The cached model is shared by reference and left in ``eval()``
+# mode — intended for repeated sequential fit/predict (cross-validation,
+# per-group models, or servers that manage their own concurrency). RES-2422
+# tracks the follow-up that externalises per-fit state so a single backbone can
+# be shared across threads too.
+_BUILT_MODEL_CACHE: OrderedDict[tuple[str, tuple[int, int]], tuple] = OrderedDict()
+_BUILT_MODEL_CACHE_LOCK = Lock()
+
+
+def _built_model_cache_size() -> int:
```
bejaeger: nit: `_get_built_model_cache_size()` ?

src/tabpfn/model_loading.py
```diff
@@ -863,17 +893,57 @@ def load_model(
...
+    *,
+    cache_trainset_representation: bool = True,
+) -> tuple[
+    Architecture,
+    nn.BCEWithLogitsLoss | nn.CrossEntropyLoss | FullSupportBarDistribution | None,
+    ArchitectureConfig,
+    InferenceConfig,
+]:
+    """Build a model from a resolved checkpoint path (no built-model cache)."""
+    checkpoint = _load_checkpoint_cached(resolved, identity)
```
bejaeger, line 946: this fn already uses the `@functools.lru_cache(maxsize=1)`. So if we set `TABPFN_MODEL_CACHE_SIZE`> 0 it seems we're caching the model twice?<br>Was that intentional? Also note that this means that the default is to have a single model cached.

src/tabpfn/model_loading.py
ggprior: addressed

src/tabpfn/model_loading.py
ggprior: * this functools lru cache is the raw checkpoint dict. it can also hold at most one entry
* The cache module proposed here caches at a model level. it can hold up to TABPFN_MODEL_CACHE_SIZE entries (think fully ready-to-use model instances of `tabpfn_2_6` and `tabpfn_3`, instead of a dict representing the checkpoint that still has to be loaded)

ggprior: I agree there is one level of duplication here for a cache miss, but it is bounded at one state dict worth of memory
But we could special case it and disable the LRU cache if the opt-in cache is set -- but arguably increases complexity

bejaeger: Makes sense.
I wonder if we should then throw out the LRU cache entirely and set the model cache's default size to 1 to preserve behavior? what do you think?

bejaeger: I leave it up to you.

bejaeger (approved):

PR #1164, opened 2026-08-06
Add TabPFNRegressor.predict_batched: fused multi-dataset regression

eliott-kalfon:
TLDR; `predict_batch` for regression, mirroring `predict_proba_batched` for classification. This is a rebased version of @ggprior's branch on that topic with a few changes shown below. I also added simple example scripts. I added errors for fp64 inference (as fp32 is hardcoded for the batch).
The current implementation of the `predict_batched` is minimal, and does not touch the predict method. This is done at the cost of some repeated code. Abstracting this away would require a wider refactoring that I would leave for a later date (similar to the classifier case).

## Claude Generated

Mirror predict_proba_batched for regression: preprocess each (X, y, X_test) exactly as fit + predict (own target standardisation, per-estimator target transforms, CPU+GPU preprocessing), stack along the model batch dim and score with a single fused forward per estimator, then decode each dataset's slice with its own raw-space bar distribution and per-estimator border transforms.

Unlike the classifier there is no shared-output constraint: the bar distribution the model predicts over is fixed by the checkpoint, so datasets need not share a target scale. Returns one entry per dataset in input order, each with the structure predict() would return, so "main"/"full" work too. Constant-target datasets are answered analytically and stitched back in input order.

Rebased onto main from georg/regressor-batched-inference, with changes:

- Run on clone(self) so the estimator is not mutated. The original left self in fit_mode="batched" with a batched executor, silently breaking any later predict().
- Take the fused batch's configs from the fitted ensemble members, not from ensemble_configs_. target_transform is fitted in place by _transform_labels_one, and under n_preprocessing_jobs > 1 that happens in a joblib worker, so ensemble_configs_ keeps an unfitted copy and the border mapping would be wrong. This is where copying the classifier verbatim breaks.
- Reuse executor_.ensemble_members instead of rebuilding a TabPFNEnsemblePreprocessor, which preprocessed every dataset twice.
- Clean X_test with ensure_compatible_predict_input_sklearn / fix_dtypes / process_text_na_dataframe as predict() does, so DataFrames, categoricals and NaNs behave identically.
- Move the concatenated block to the fit device before GPU preprocessing.
- Reject ragged batches rather than relying on collator padding, which would feed the model fake context rows.
- zip(..., strict=True) over the input lists.

Tests in test_regressor_interface.py (the classifier equivalents live in test_classifier_interface.py, not the finetuning suite): per-dataset parity for mean and quantiles, per-dataset criterion under "full", fitted-target-transform parity at n_preprocessing_jobs=2, DataFrame parity, constant-target mixing, fp16, ragged and argument rejection, and non-mutation of the estimator.

## Issue

## Motivation and Context

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested? Locally and CI

---

## Checklist

---


src/tabpfn/regressor.py
```diff
@@ -1307,6 +1311,311 @@ def predict(  # noqa: C901, PLR0912
...
+            worker.fit_from_preprocessed(
+                batch.X_context,
+                batch.y_context,
+                batch.cat_indices,
+                batch.configs,
+                performance_options=PerformanceOptions(),
+            )
+            # One fused forward per estimator, each output
+            # (n_test, n_fused_datasets, n_buckets) with one config per dataset.
+            raw_outputs = list(
```
ggprior: this materializes all outputs at once. Can we iterate estimators on the outside, and datasets on the outside, and yield one estimator at a time, we would get around having all predictions in memory

ggprior (dismissed):
LGTM with one minor memory related comment

src/tabpfn/regressor.py
eliott-kalfon: Thanks Georg, such a cool suggestion, I have just pushed a commit with it.

ggprior (approved):

PR #1165, opened 2026-08-07
Add an attention-backend registry

jmkuebler:
This PR does a larger refactor for the attention backends. The dispatching behavior is unchanged wrt to main. 
It serves a few purposes:
- it defines an attention registry where users can also register a custom attention backend if they want
  - the customer registered users take precedence, whereas built-in backends follow an explicit order. Fallback is always torch.sdpa (which itself has backends, but that is left untouched).
  - moves backend specific operations into their modules.
  - [new in Revision 1] checks imports for backends once at registration
- <del>it moves the backend selection out of the individual forward calls and ahead of each stage. Here the shapes, devices and dtypes are known, and we cna select the backend once for the entire transformer. Note ICL stage can still use a different backend as feature aggregator for example, as they have quite different shapes. </del> [Revision 1: I moved this back into the per-call selection, since I realized this is not causing the graph breaks and otherwise causes a lot of code to maintain and model-specific logic wihtout any clear gains]
- The logic of the FA3 selection currently breaks the compile graphs in many pieces and prevents compilation gains. This is now circumvented by chcking importability once during backed registration. We add a seperate test for this here https://github.com/PriorLabs/TabPFN/pull/1166. On main this currently fails, but with the new backend selection upfront it will work (and can then be merged).

❗ **@Reviewers:❗**:
- Please make only a high level pass right now while we are collecting first feedback.
- Do not review the test cases in detail, I still need to do this once we agree that the approach is good.
 

<details><summary> Claude Description </summary>

## What

Replaces the hardcoded FA3 / torch-MPS / MLX dispatch inside the shared
`scaled_dot_product_attention` with a small explicit registry of attention
backends. Each in-tree backend now lives in its own module and is registered —
only if its dependency is actually importable — with the shared SDPA module
owning the consult order.

## Why

- An alternative attention implementation can be added without editing the
  dispatch code: a backend registers itself and decides per call whether to
  take it. This includes backends that are not part of this package.
- Backend-specific logic (capability gates, seqlen crossovers, layout
  conversions) moves out of the shared chokepoint and into the backend that
  owns it.
- Gating registration on availability removes the `import` probe from the
  attention hot path. That probe was the reason the current dispatch fragments
  `torch.compile` graphs — Dynamo cannot trace an `import` statement, so every
  attention call broke the graph. Measured on the case where the import fails
  (torch 2.13, no FA3 wheel): 20 graph breaks / 43 frames on main → 0 breaks /
  2 frames on this branch. Verified by the test in #1166.

## Design notes

- `is_preferred` receives an `AttentionSpec` — the call's geometry (sequence
  lengths, head counts, head dim, dtype, device, whether grad is enabled) — not
  the tensors. A backend cannot tell which stage of
  belongs to, only what the call looks like. Resolving a call this way costs
  ~2.5 µs and traces cleanly.
- Registration is explicit, with no discovery, entry points, or env vars: one
  variadic `register_attention_backend()` call. Bac
  consulted first, so an externally registered backend takes precedence over
  the in-tree ones, which keep the order the old in
  Re-registering the same object is a no-op, so setup hooks may run repeatedly.
- Availability (`is_available()`, an import/version
  preference (`is_preferred(spec)`, per-call geometry). Availability is checked
  once at import; it deliberately does not touch CU
  still does not initialize a device.
- Quantized KV cache entries are dequantized exactl
  backend that sets `consumes_quantized_kv = True` receives the entry as stored
  instead, so it can feed its kernel without materi
- `scaled_dot_product_attention` takes a three-valued `backend=` kwarg:
  `"auto"` (default, consult the registry), `None`
  backend object (force that one). Tests use it to pin a path.
- Backends fail open: nothing registered, or none p
  hardened torch-SDPA path as before, chunking included.
- The selected backend is debug-logged per call (sh
  the stage is recognizable from the geometry. Skipped while tracing, since
  `logging` is itself untraceable.

## Behavior

With nothing extra registered, selection is identic
predicates, same order, same fallback. The v2 family is untouched, and
`inference.py` is unchanged.

## Testing

- Registry contract tests: a preferred backend's ou
  one falls through, consult order is newest-first, registration is re-entrant
  and rejects name conflicts, quantized KV is passe
  once, an explicit `backend=` bypasses the registry, and the spec describes
  the live call.
- Dispatch tests through a real v3 forward: a backend receives the calls it
  prefers, and the cached-predict path reports the
- The existing FA3 and MLX tests are adapted to the spec-based predicates;
  their Hopper/MPS coverage is unchanged.

🤖 Generated with [Claude Code](https://claude.com/
</details>

### Additional Testing:
Since are touching backends etc that run only on specific hardware, I also run tests on those platforms explicitly.

- [x] Test on Mac
Main:
```
============================================== 1486 passed, 174 skipped, 8 xfailed, 2691 warnings in 308.60s (0:05:08) ===============================================
```
This branch
```
============================================ 1503 passed, 174 skipped, 8 xfailed, 2691 warnings in 303.63s (0:05:03) =============================================
```
- [x] Test on H100
main (without FA3 installed)
```
=========================================== 1580 passed, 112 skipped, 12 xfailed, 4527 warnings in 658.08s (0:10:58) ============================================
```
main (with FA3)
```
==================================================== 1587 passed, 105 skipped, 12 xfailed, 4529 warnings in 635.79s (0:10:35) =====================================================
```
this PR (with FA3)
```
============================================= 1607 passed, 103 skipped, 12 xfailed, 4529 warnings in 608.05s (0:10:08) ==============================================
```


src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1967,6 +2071,223 @@ def get_default_performance_options(self) -> PerformanceOptions:
...
+                icl_attn.planned_backend = icl_backend
+                icl_attn.planned_backend_test = icl_test_backend
+
+        # ---- Decoder (multiclass retrieval attention) ----
+        if self.task_type == "multiclass":
+            decoder = self.many_class_decoder
+            if cache_is_populated:
+                assert kv_cache is not None
+                num_test = num_rows
+                decoder_kv_len = kv_cache.train_shape[1]
```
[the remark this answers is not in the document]
jmkuebler: fixed with [2d7fed2](https://github.com/PriorLabs/TabPFN/pull/1165/commits/2d7fed25ce52c3c34e81ee6b641a3a15a21d672b)

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -507,6 +519,7 @@ def __init__(
         self.q_projection = nn.Linear(self.input_size, self.attention_size)
         self.k_projection = nn.Linear(self.input_size, self.attention_size)
         self.softmax_scaling_layer = softmax_scaling_layer
+        self.planned_backend: PlannedAttentionBackend = "auto"
```
ggprior: How do we set PlannedAttentionBackend to something different than auto from a library user perspective? Perhaps we need to funnel this through to the sklearn interfaces or accept environment variables? Or would it be possible to do a module-level override?

src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -0,0 +1,260 @@
...
+    runs at plan time, before the tensors of the call exist. A backend that
+    can be planned into a ``torch.compile``-d region must have a
+    Dynamo-traceable ``run`` — the registry does not check this; it is the
+    registrant's responsibility.
+    """
+
+    name: str
+    """Unique backend name."""
+
+    def is_preferred(self, spec: AttentionSpec) -> bool:
```
ggprior, line 100: what if multiple backends are preferred?

src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -0,0 +1,260 @@
...
+                raise ValueError(
+                    f"A different attention backend is already registered under "
+                    f"{backend.name!r}: {existing!r}"
+                )
+            if existing is None:
+                _registry[backend.name] = backend
+                new.append(backend)
+        _consult_order = (*new, *_consult_order)
+        _logger.debug(
+            "registered attention backends, consult order: %s (registration "
```
ggprior, line 150: we log the registered backends but we never seem to log what ends up being chosen?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1837,6 +1899,54 @@ def forward(
             torch._dynamo.mark_dynamic(x_RiBC, index=1)
             torch._dynamo.mark_dynamic(x_RiBC, index=2)
 
+        # On the cache path the stages drop the train rows unless the caller
+        # already passed test rows only (see _stages_0_to_2), so plan for the
+        # rows the attention calls will actually see.
+        num_rows = x_RiBC.shape[0]
+        if kv_cache is not None and not kv_cache.is_empty() and not x_is_test_only:
+            num_rows -= num_train
+        self._plan_attention_backends(
```
ggprior: how expensive is this plan op itself?

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1967,6 +2077,225 @@ def get_default_performance_options(self) -> PerformanceOptions:
...
+                    num_heads=decoder.num_heads,
+                    num_kv_heads=decoder.num_heads,
+                    head_dim=decoder.head_dim,
+                    # Class chunks are folded into the batch dimension.
+                    batch=batch_size
+                    * math.ceil(decoder.max_num_classes / decoder.head_dim),
+                )
+            )
+
+    def _reset_attention_plan(self) -> None:
```
ggprior: this and above is quite verbose, is it actually the case that we might have different backends per module?
even if there a benefit, does it outweigh the complexity added here?

ggprior (commented):
drive-by comments

tests/test_architectures/test_attention_backends.py
ggprior, line 1: i feel something here should fail loudly here if any SDPA op gets hardcoded in an architecture, if that's easily possible

src/tabpfn/architectures/shared/attention_backends.py
jmkuebler: Then the order matters, the first one that wants it, picks. 

The ones that are registered in tree have a hardcoded order, see https://github.com/PriorLabs/TabPFN/pull/1165/changes/BASE..2d7fed25ce52c3c34e81ee6b641a3a15a21d672b#diff-6d343bc8f7cbc7ea2976639778508cc004dc786e142af66cf5ffc95f7fda0316R28-R31 

(it reflects the order of checks we had so far).

Externally registered backends take preference and are added on top of the in-tree order

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: we currently are not exposing this to the user (same as we did not expose anything for the backends so far).

I actually don't see much user who would do that rn, so maybe we can add it when needed?

src/tabpfn/architectures/shared/attention_backends.py
jmkuebler: Good point. Since it is only once per forward, that would be nice for debug. (I was a bit afraid of overlogging, but should be fine.) will add that

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: super cheap, its just a few checks on availability and and size limits. Note that we have done this similarly for each attention forward (so for example 24 times in the ICL stage of v3).

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: yes, it is possible and happens a lot. For example FA3 only kicks in after q_len >10K or sth similar. So this only ever happens in the ICL stage.

I'll make less verbose. Maybe I can define first a set of attn layers once and then always iterate over it.

tests/test_architectures/test_attention_backends.py
jmkuebler: hm i tried with claude but it did not come up with anything better than clumsily  checking the actual code strings... would not get blocked on this for now

jmkuebler:
@ggprior thanks for your comments. FYI: I did quite a big revision. I removed the entire logic to preselect the backends upfront. This was causing very verbose code in the model file itself. And i realized that I can work around compilation graph breaks also via checking imports once upfront instead on each forward pass. So overall the diff is reduced now.

ggprior (dismissed):

src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -0,0 +1,238 @@
...
+    """Number of query heads."""
+
+    num_kv_heads: int
+    """Number of key/value heads (< ``num_heads`` under GQA/MQA)."""
+
+    head_dim: int
+    """Dimensionality of each head."""
+
+    dtype: torch.dtype
+    """Effective dtype of q/k/v at the call (autocast-aware)."""
```
oscarkey: could you expand the comment to explain what "autocast-aware" means?

src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -0,0 +1,238 @@
...
+    """Whether grad mode is on, i.e. whether a backward may follow."""
+
+
+@runtime_checkable
+class AttentionBackend(Protocol):
+    """A drop-in replacement for one attention call.
+
+    All tensors use the ``(B, S, H, D)`` layout of
+    ``shared.scaled_dot_product_attention``. Backends receive dense
+    ``k``/``v`` — the chokepoint dequantizes a quantized KV cache entry
```
oscarkey: what's "the chokepoint" here?

src/tabpfn/architectures/shared/fa3_backend.py
```diff
@@ -97,3 +86,52 @@ def fa3_attn_func(
             "see fa3_setup.md (next to this file)."
         )
     return fn(q, k, v)
+
+
+class FA3Backend:
```
oscarkey: These implementations may as well extend the `AttentionSpec` protocol (unless it causes a loop)? Just for ease of finding all the impls in the future if we change the interface.

src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+_logger = logging.getLogger(__name__)
+
+_registry: dict[str, AttentionBackend] = {}
+_consult_order: tuple[AttentionBackend, ...] = ()
+_lock = threading.Lock()
+
+
+def register_attention_backend(*backends: AttentionBackend) -> None:
```
oscarkey, line 123: I'm wondering if this might go wrong if we start lazily importing `scaled_dot_product_attention.py` for some reason, in which case the external packages might register their backends before the internal ones. Do you think this is possible? We could add a `priority` option, but that does make it less elegant.

oscarkey (commented):
seems pretty cool & elegant to me! Just some minor comments, and a question re import ordering.

src/tabpfn/architectures/shared/attention_backends.py
jmkuebler: I actually removed the autocast part, as it doesn't matter at all at this point.

It was leftover from the first design where I planned beforehand, and there it was important to reflect what autocast would do when the q/k/v hit the attention interface. But since, planning is removed, we use the actual dtype they arrive in, and it is completely irrelelvant whether autocast was at play or not.

src/tabpfn/architectures/shared/attention_backends.py
jmkuebler: Sorry, that was leftover claude jargon. Replaced it with "SDPA wrapper" which should be clearer.

src/tabpfn/architectures/shared/fa3_backend.py
jmkuebler: Thank you, but You mean `AttentionBacken`, no ? I added that, but let me know if I misunderstood.

oscarkey: oh yes sorry, that is what I meant.

src/tabpfn/architectures/shared/attention_backends.py
jmkuebler: Hmmm. I looked a bit into this, but rn I'd say let's keep it. It currently cannot happen, because TabPFN import triggers the registration of internal backends (in principle this could change in the future), and that it the prerequisite for registration of an external backend.

I tried to put the backends in buckets (internal vs external), but that creates quite a bit of code overhead as you have to discriminate at registration time.

I also tried to guard the above mentioned import dependency, but the resulting test runs 3.5s and needs to create it's own subprocess, so it is also pretty ugly.

Lastly, rn we would only run into troubles if we register another backend that runs on the same hardware and would like to take a certain workload. Also not the case.

So I'd leave it as is.

oscarkey: okay sgtm! Thank you for looking into it.

oscarkey (approved):

PR #1167, opened 2026-08-08
fix seed inside truncated svd

jmkuebler:
## Issue
On main the Truncated SVD is not deterministic. But it cannot directly be seeded. This leads to nondeterministic behavior. On large datasets of BeyondArena, which uses this path this lead to heavily non-deterministic results (that seemed statistically significant far apart).

This PR adds a test (that fails on main) and the locally forces a fixed random seed for the truncated SVD, I don't think there is any need for randomness in the SVD operation.

## Motivation and Context

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

---

## Checklist

---

### DIFF
--- a/src/tabpfn/preprocessing/torch/torch_svd.py
+++ b/src/tabpfn/preprocessing/torch/torch_svd.py
@@ -111,10 +111,12 @@ class TorchTruncatedSVD:
         )
 
         if use_lowrank:
-            # torch.svd_lowrank returns (U, S, V) with A ≈ U diag(S) V^T.
-            # fork_rng keeps the seeding local: the global stream is restored
-            # afterwards, so no other consumer of torch's RNG is affected.
-            with torch.random.fork_rng(devices=[]):
+            # torch.svd_lowrank returns (U, S, V) with A ≈ U diag(S) V^T. It draws a
+            # random projection from the global RNG and takes no generator, so an
+            # unseeded call returns different components every time. Seed it, and fork
+            # every device so the caller's streams are restored on exit
+            # (torch.manual_seed reseeds CPU *and* all CUDA devices).
+            with torch.random.fork_rng(devices=range(torch.cuda.device_count())):
                 torch.manual_seed(0)
                 u, s, v = torch.svd_lowrank(x_filled, q=q, niter=2)
             # Truncate oversampling dimensions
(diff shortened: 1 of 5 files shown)


src/tabpfn/preprocessing/torch/torch_svd.py
```diff
@@ -111,8 +111,22 @@ def fit(self, x: torch.Tensor) -> dict[str, torch.Tensor]:
...
+            # projection from the default generator of x_filled.device and takes no
+            # generator argument, so an unseeded call returns different components
+            # every time. Fork and seed only that device's generator: torch.manual_seed
+            # would also reseed CUDA/MPS/XPU, which a CPU-only fork_rng does not
+            # restore, and forking every device would initialise a context on each.
+            device = x_filled.device
+            fork_devices = [] if device.type == "cpu" else [device]
+            with torch.random.fork_rng(devices=fork_devices, device_type=device.type):
+                if device.type == "cpu":
+                    torch.default_generator.manual_seed(0)
```
LennartPurucker: So, to my understanding, this would break with the (implicit) API convention of scikit-learn models and randomness in tabular machine learning models. 

Any (potentially useful) randomness should be controllable by the user at the API level via the random seed. I am not sure how strict the rest of our package was regarding this behavior, but I think we follow this convention. Hence, I would argue that this should also depend on the user's input seed. 

Moreover, we should also follow the packages' torch seeding practices. Otherwise, overriding this here in the code will likely break outer experiments that wish to test the model with their own, separate global state and would suddenly be reset in the pipeline. E.g., in TabArena, we have a test to check if the model changes our global variables, which this would fail (I think?).

jmkuebler: Thanks a lot @LennartPurucker that was really helpful.
I reworked it and am now passing through the random state. Now it is aligned with the user control, and in particular if the global seed is set to None, then also the SVD is random (which my previous attempt did not bring).

It also gates the seed in a way that does not interact with outside torch seeding. So imo the TabArena concern is fine.

LennartPurucker (approved):
Awesome, thank you! LGTM

jmkuebler:
@cursor review

PR #1171, opened 2026-08-11
Only auto-scale n_estimators when it is left at "auto"

klemens-floege:
## What

`n_estimators` now defaults to `"auto"` on `TabPFNClassifier` / `TabPFNRegressor`. Feature-coverage auto-scaling applies **only** to `"auto"` — an explicitly passed integer is used exactly as given, never raised.

- `scale_n_estimators_for_feature_coverage` takes `int | Literal["auto"]`; `"auto"` resolves to the new `DEFAULT_N_ESTIMATORS = 8` and may then be scaled up.
- `create_default_for_version(...)` passes `"auto"` instead of a hardcoded 8, so per-version defaults keep the scaling behaviour.
- `auto_scale_n_estimators` is unchanged and still works: `False` pins `"auto"` to `DEFAULT_N_ESTIMATORS`.

## Why

Per discussion with @JonasKuebler: keeping the auto-scaling default is good for users on wide datasets, but the package shouldn't silently override `n_estimators=8` when someone asked for 8. This is the same fix as PriorLabs/fomo-fitting#3279 but at the right layer — with this in, that PR's `auto_scale_n_estimators=False` plumbing becomes unnecessary.

## Behaviour change

| construction | before | after |
|---|---|---|
| `TabPFNClassifier()` | 8, scaled up on wide data | same (`"auto"`) |
| `TabPFNClassifier(n_estimators=8)` | scaled up on wide data | exactly 8 |
| `get_params()["n_estimators"]` | `8` | `"auto"` |

Users who *want* scaling from an explicit base should pass `n_estimators="auto"`.

## Test plan

- Updated + new unit tests in `tests/test_preprocessing/test_ensemble.py`: `"auto"` scales and warns, explicit `2`/`8` never scale and never warn, `auto_scale_n_estimators=False` pins to the default.
- `tests/test_estimators.py` (52 passed), version-default tests in both interface suites updated to expect `"auto"`.
- Smoke-fitted both estimators with `"auto"` and explicit values; `clone()`/`get_params()` round-trip fine.

https://claude.ai/code/session_01LGbpojWL3qbBkAYYCAejdn


src/tabpfn/preprocessing/ensemble.py
```diff
@@ -1217,28 +1217,44 @@ def _resolve_feature_subsampling_method(
...
-    ``n_estimators`` is returned unchanged (this is the ``auto_scale_n_estimators``
-    constructor argument on the estimator); some features may then never be sampled.
+    When ``auto_scale_n_estimators`` is False the scaling is skipped and ``"auto"``
+    simply resolves to ``DEFAULT_N_ESTIMATORS`` (this is the
+    ``auto_scale_n_estimators`` constructor argument on the estimator); some
+    features may then never be sampled.
     """
+    if n_estimators != "auto":
+        # A value the user chose explicitly is never overridden.
+        return n_estimators
```
jmkuebler, line 1284: Hmmm. So far a user could have auto scaling enabled starting from `n_estimators=4` as an example. With the code now, that doesn't work anymore. I am wondering whether we want to support autoscaling from an other value as the default 8.

I did not consider that myself before, but it _might_ actually be used. Let's check!

The other way could be that the autoscaling has value `None` by default. If `n_estimators="auto"` then it will autoscale, if an int comes in it won't. And if `auto_scale_n_estimator` is set, than that takes effect.

jmkuebler: ok if we accept that we loose this corner case, then wouldn't it make sense to remove the `auto_scale_n_estimator` argument entirely?

`n_estimators="auto"` => auto-scaling with minimum 8 estimators

`n_estimators` set to an integer ==> no auto scaling

I think right now the `auto_scale_n_estimator` only buys us that we could run default without auto-scaling, but that would be easier to do with just setting it to an integer.

I'd also add a warning if `n_estimators` is set explicitly, but doesn't suffice to cover the entire table.

klemens-floege: I am removing this feature now: auto_scale_n_estimator

klemens-floege: > n_estimators="auto" => auto-scaling with minimum 8 estimators

Right now we do autoscaling with 8 estimators and not minimum 8

jmkuebler (approved):
ok LGTM now with the deprecation warning as we discussed offline.

PR #1172, opened 2026-08-11
Add temperature calibration to TabPFNRegressor

tobias-priorlabs:
# Add temperature calibration to TabPFNRegressor

RES-830.

`TabPFNRegressor` gains `eval_metric` and `tuning_config`, mirroring
`TabPFNClassifier`. When `tuning_config={"calibrate_temperature": True}`, `fit()`
calibrates a single temperature on held-out data and stores it in
`ensemble_softmax_temperature_`. It is applied to the aggregated ensemble logits
inside `predict()`.

```python
reg = TabPFNRegressor(
    eval_metric="nll",                              # or "crps"
    tuning_config={"calibrate_temperature": True},
)
reg.fit(X_train, y_train)
reg.ensemble_softmax_temperature_                   # e.g. 1.12 -> widened
```

## Why

The regressor sharpens every ensemble member by a fixed `softmax_temperature` of
0.9 and has no way to correct that per dataset post-hoc. The classifier already solves the
analogous problem on holdout data; this reuses that infrastructure for the regressor.

## Method

The calibrated temperature acts *after* ensemble aggregation, `softmax(log(sum_i p_i) / T)`. 
`T > 1` widens the distribution, `T < 1` sharpens it. The per-estimator
`softmax_temperature` is untouched.

`eval_metric` selects what the sweep minimises:

- `"nll"` (default) — negative log density under the bar distribution.
- `"crps"` — delegates to `_ranked_probability_score_loss_from_bar_logits`, the
  same implementation used as a finetuning loss.

Both are evaluated in raw target units against the fold's `raw_space_bardist_`.

Degenerate cases fall back to `T = 1.0`, a no-op.

## Task-agnostic shared infrastructure

- `resolve_tuning_config` takes a `config_cls` instead of inferring the config
  type from dict keys. `TabPFNClassifier` passes `ClassifierTuningConfig`
  explicitly and narrows the result with an assert.
- `get_tuning_splits` takes `task_type`. Regression uses a shuffled `KFold` — a
  continuous target has nothing to stratify on — keeping the guarantee that no
  sample is held out twice.
- `get_tuning_feature_flags` reads the enabled features off the config's boolean
  fields, so the "no tuning features enabled" warning names the options that
  config actually has rather than the classifier's hardcoded pair.
- `RegressorTuningConfig` adds no fields over `TuningConfig`. It exists so the
  task type is recoverable from the config alone, and to give regression-only
  options somewhere to live.

##### Classifier's temperature grid

Both searches now use `get_tuning_temperatures()` — 0.60 to 1.40 in steps of
0.01. The classifier previously swept `np.linspace(0.6, 1.4, 82)`, which never lands on the default value 1.0.

##### `predict_batched` rejects `tuning_config`

`TabPFNRegressor.predict_batched` (added in #1164, which landed on `main` while
this branch was open) now raises `NotImplementedError` when the estimator was
constructed with a `tuning_config`, where it previously returned numbers.

`predict_batched` carried its own copy of the aggregation tail, so it never
applied `ensemble_softmax_temperature_`: with calibration enabled it silently
returned **uncalibrated** predictions while its docstring claimed parity with
`predict`. It also paid for a discarded calibration sweep per dataset, since the
clone it fits inherits `tuning_config`.

There is no correct fused answer to return. The temperature is calibrated on each
dataset's own holdout, so a shared batch has no single temperature to apply —
two datasets in a quick check calibrated to 1.4 and 1.06. This mirrors the
existing guard in `TabPFNClassifier.predict_proba_batched`, which rejects
`tuning_config` for exactly the same reason. Score such datasets individually
with `predict`.

## Refactor

Aggregation is extracted from `predict()` into `_compute_aggregated_logits()` and
shared with the calibration path, which needs the same logits on holdout rows.
`predict()` loses its `noqa: C901, PLR0912`.

The tail of that aggregation — averaging, `log`, the float16 fix and the
temperature — is further extracted into `_reduce_accumulated_logits()` and shared
with `predict_batched`'s decode path. That duplication is what let the two paths
diverge in the first place; sharing it means a future change to the reduction
reaches both.

## Tests

Roughly 40 new tests:

- `test_inference_tuning.py` — regression splits (disjoint holdouts, generator
  random states, constant targets, classification unchanged); the temperature
  search widens overconfident and sharpens underconfident distributions, pools
  differing bar distributions, weights folds by holdout size, is invariant to
  target units, leaves the caller's bar distributions unmutated, and beats no
  calibration for every metric; both searches only return grid values, and the
  grid contains 1.0.
- `test_regressor_interface.py` — holdout triples are self-consistent, constant
  training targets are skipped, each `eval_metric` is honoured end to end, no
  `tuning_config` leaves the temperature a no-op, an invalid metric raises, and
  the new params survive `sklearn.clone`. Plus `predict_batched` rejecting
  `tuning_config`, and a test that spies on `_reduce_accumulated_logits` so
  re-inlining a private copy of the reduction fails the suite.
- `test_save_load_fitted_model.py` — the calibrated temperature round-trips and
  still changes predictions after loading.

## Example

`examples/tabpfn_regressor_with_tuning.py` compares an uncalibrated baseline
against calibration on NLL and on CRPS, reporting test NLL, CRPS and empirical
quantile coverage, and showing that point predictions barely move — calibration
reshapes the distribution rather than shifting its centre. The classifier example
is renamed `tabpfn_classifier_with_tuning.py` for symmetry (rename only).

| dataset | T(NLL) | T(CRPS) | NLL @ T=1 | NLL @ T(NLL) | CRPS @ T=1 | CRPS @ T(CRPS) |
| --- | --- | --- | --- | --- | ---  | --- |
| friedman1 noise=1 | 1.25 | 1.09 | 1.48080 | **1.47082** | 0.56505 | 0.56505 |
| friedman1 noise=10 | 1.27 | 1.05  | 3.76264 | **3.75023** | 5.55744 | **5.55635** |
| make_regression noise=25 | 1.30  | 1.13 | 4.70279 | **4.68943** | 14.14720 | **14.13665** |
| diabetes (real, n=442) | 1.24 | 1.07 | 5.43564 | **5.40758** | 31.54079 | **31.41526** |

---

**Note for reviewers:** this branch was rebased onto `main` and force-pushed, so
anyone with it checked out needs to reset rather than pull. The rebase replaced
an earlier merge commit whose conflict resolution had left a malformed import
block in `regressor.py` that did not parse.


src/tabpfn/regressor.py
```diff
@@ -1219,6 +1290,196 @@ def predict(  # noqa: C901, PLR0912
...
+            median_out = typing.cast(
+                "np.ndarray", logit_to_output(output_type="median")
+            )
+            mode_out = typing.cast("np.ndarray", logit_to_output(output_type="mode"))
+            quantiles_out = typing.cast(
+                "list[np.ndarray]",
+                logit_to_output(output_type="quantiles"),
+            )
+
+            # Create our typed dictionary
```
oscarkey:
```suggestion
```

src/tabpfn/regressor.py
```diff
@@ -1261,6 +1522,32 @@ def predict(  # noqa: C901, PLR0912
...
+        """Average the ensemble's accumulated output and apply the temperature.
+
+        The tail shared by `_compute_aggregated_logits` (the `predict` path) and
+        `_decode_batched_dataset` (the `predict_batched` path). Both reach this
+        point with the same quantity -- the sum over estimators of one dataset's
+        bucket probabilities on the `znorm_space_bardist_` borders, already in
+        log space when `average_before_softmax` is set -- so the reduction must
+        stay identical between them. Keeping it in one place is what makes that
+        true; `predict_batched` previously carried its own copy and silently
+        omitted the temperature.
```
oscarkey: Rather than this comment, I'd just mention in the `accumulated_logits` docstring that it's in log-space when `average_before_softmax=True`. Claude likes to do it, but we don't need to mention the call sites in the docstring, and we also don't need to discuss the history of the code, just the current state. This is in our [internal](https://github.com/PriorLabs/fomo-fitting/blob/main/.claude/CLAUDE.md#comments--docstrings) Claude config, but we haven't shared that with this repo yet.

src/tabpfn/regressor.py
```diff
@@ -1219,6 +1290,196 @@ def predict(  # noqa: C901, PLR0912
...
+
+    def _compute_aggregated_logits(self, X: XType) -> torch.Tensor:
+        """Run the ensemble and aggregate it into one log-probability tensor.
+
+        Each estimator's bucket probabilities are translated onto the
+        `znorm_space_bardist_` borders, averaged across the ensemble
+        (before or after the softmax, per `average_before_softmax` flat),
+        and returned in log space.
+
+        Shared by `predict` and `_maybe_calibrate_ensemble_temperature`.
```
oscarkey, line 1467:
```suggestion
```

src/tabpfn/inference_tuning.py
```diff
@@ -148,39 +177,126 @@ def compute_metric_to_minimize(
...
+    )
+
+
 def get_tuning_splits(
     X: np.ndarray,
     y: np.ndarray,
     holdout_frac: float,
     n_splits: int = 1,
     random_state: int | np.random.RandomState | np.random.Generator | None = 0,
+    task_type: Literal["classifier", "regressor"] = "classifier",
```
oscarkey, line 246: As neither classifier nor regressor is more important, I would remove the default value here so we always have to be explicit. This makes future accidents where we forget to specify the task type less likely.

src/tabpfn/inference_tuning.py
```diff
@@ -390,42 +596,65 @@ def get_default_tuning_n_folds(n_samples: int) -> int:
...
+        if isinstance(value := getattr(tuning_config, field.name), bool)
+    }
+
+
 def resolve_tuning_config(
-    tuning_config: dict | ClassifierTuningConfig | None,
+    tuning_config: dict | TuningConfig | None,
     num_samples: int,
-) -> ClassifierTuningConfig | None:
+    config_cls: type[TuningConfig] = ClassifierTuningConfig,
```
oscarkey, line 599: again I wouldn't have the default to avoid forgetting it in the future.

src/tabpfn/classifier.py
```diff
@@ -1256,6 +1257,10 @@ def _maybe_calibrate_temperature_and_tune_decision_thresholds(
                 )
             return
 
+        # `resolve_tuning_config` is task-agnostic; narrow it back down so the
+        # classification-only options below are visible.
```
oscarkey:
```suggestion
```

src/tabpfn/regressor.py
```diff
@@ -1219,6 +1290,196 @@ def predict(  # noqa: C901, PLR0912
...
+
+            if tuning_regressor.is_constant_target_:
+                # This fold's training split is single-valued, so its regressor
+                # returned early with a degenerate two-border distribution and no
+                # executor. There is nothing to calibrate against; skip the fold.
+                continue
+
+            # The tuning regressor has no `ensemble_softmax_temperature_`, so these
+            # are the untempered aggregated logits, with the per-estimator
+            # `softmax_temperature` correctly still applied.
```
oscarkey, line 1440: Should this comment be above line 1445 instead?

src/tabpfn/inference_tuning.py
```diff
@@ -354,6 +470,96 @@ def find_optimal_temperature(
...
+        if logits.shape[0] > 0
+    ]
+    if not folds:
+        return current_default_temperature
+
+    # Imported here rather than at module level so that this module stays free of
+    # torch at import time, as it was before regression calibration existed. `torch`
+    # appears throughout the signatures above, but `from __future__ import
+    # annotations` keeps those unevaluated, and `no_grad` is the only runtime use.
+    import torch  # noqa: PLC0415
```
oscarkey: we can just import at the top of the file, `torch` is imported immediately by other stuff anyway.

src/tabpfn/inference_tuning.py
```diff
@@ -148,39 +177,126 @@ def compute_metric_to_minimize(
...
+    """The bin-width-weighted CRPS already used as a finetuning loss.
+
+    Delegates to `finetuning.finetuned_regressor`'s implementation.
+
+    Imported lazily: `finetuned_regressor` imports `TabPFNRegressor` at module level,
+    and `regressor` imports this module, so a top-level import raises ImportError on a
+    partially initialized `tabpfn` package.
+    """
+    from tabpfn.finetuning.finetuned_regressor import (  # noqa: PLC0415
+        _ranked_probability_score_loss_from_bar_logits,
```
oscarkey: we should probably move this function into a different module and make it public, rather than this hack?

src/tabpfn/inference_tuning.py
```diff
@@ -148,39 +177,126 @@ def compute_metric_to_minimize(
...
+# every holdout row or, where the underlying implementation only exposes it, the mean
+# loss over the fold;
+REGRESSION_METRIC_NAME_TO_OBJECTIVE: dict[
+    str,
+    Callable[
+        [FullSupportBarDistribution, torch.Tensor, torch.Tensor],
+        torch.Tensor,
+    ],
+] = {
+    # `forward` is the negative log density of each target under the bar distribution.
```
oscarkey: there is no "`forward`"?

examples/tabpfn_regressor_with_tuning.py
```diff
@@ -0,0 +1,116 @@
...
+
+QUANTILE_LEVELS = [0.1, 0.25, 0.5, 0.75, 0.9]
+
+
+def evaluate(regressor: TabPFNRegressor) -> tuple[float, float, dict[float, float]]:
+    """Return test-set NLL, CRPS and the empirical coverage of the quantiles."""
+    output = regressor.predict(X_test, output_type="full", quantiles=QUANTILE_LEVELS)
+
+    # `criterion` is the bar distribution in the units of the original target, so it
+    # scores the raw `y_test` directly. Both metrics return one loss per row.
```
oscarkey, line 41: Claude pointed out that this isn't true, for CRPS we get a scalar

src/tabpfn/inference_tuning.py
```diff
@@ -390,42 +596,65 @@ def get_default_tuning_n_folds(n_samples: int) -> int:
     return 1
 
 
+def get_tuning_feature_flags(tuning_config: TuningConfig) -> dict[str, bool]:
```
oscarkey, line 574: maybe a `get_enabled_feature_flags() -> list[str]` on `TuningConfig` would be neater?

src/tabpfn/regressor.py
```diff
@@ -1219,6 +1290,196 @@ def predict(  # noqa: C901, PLR0912
...
+                    **main_outputs,
+                    criterion=self.raw_space_bardist_,
+                    logits=logits,
+                )
+
+            return main_outputs
+
+        return logit_to_output(output_type=output_type)
+
+    def _maybe_calibrate_ensemble_temperature(self, X: XType, y: YType) -> None:
```
oscarkey, line 1334: did you time how long this tends to take? Especially as there's a bit more going on in the regression case compared to classification. I'm just wondering if we might want to add a progress bar as a follow-up.

src/tabpfn/regressor.py
```diff
@@ -1219,6 +1290,196 @@ def predict(  # noqa: C901, PLR0912
...
+            # `raw_space_bardist_` undoes this fold's normalisation, so the targets
+            # are scored in their original units and need no transformation --
+            # only a tensor of the dtype and device the bar distribution expects.
+            raw_space_bardist = tuning_regressor.raw_space_bardist_
+            y_holdout_Nh_tensor = torch.as_tensor(
+                y_holdout_Nh,
+                dtype=raw_space_bardist.borders.dtype,
+                device=logits_NhB.device,
+            )
+            holdout_folds.append((logits_NhB, raw_space_bardist, y_holdout_Nh_tensor))
```
oscarkey, line 1455: how much memory does this use?
If it's a problem, can we make this function a generator, and then update `find_regression_optimal_temperature()` to loop first over folds and then over temperatures (not sure if the maths works out), so we only have to have one fold in memory at a time? Maybe fine as it is, though.

oscarkey (commented):
functionally lgtm, aside from time and memory questions!
I left a load of minor style comments, although I'm never sure how useful they are these days.

examples/tabpfn_regressor_with_tuning.py
tobias-priorlabs: I agree. I removed the comment on the shape of each metric.

src/tabpfn/inference_tuning.py
tobias-priorlabs: I have done the suggested refactor in 5ad39cb, with a bug fix in aeb02ca

src/tabpfn/finetuning/finetuned_regressor.py
```diff
@@ -103,7 +106,7 @@ def _compute_regression_loss(  # noqa: C901
         total_loss = total_loss + crps_loss_weight * crps_loss
 
     if crls_loss_weight > 0.0:
-        crls_loss = _ranked_probability_score_loss_from_bar_logits(
+        crls_loss = ranked_probability_score_loss_from_bar_logits(
```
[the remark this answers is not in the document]
tobias-priorlabs: Fixed in aeb02caf4d03171bfbe98de41f568f5cd24c5a06

src/tabpfn/inference_tuning.py
tobias-priorlabs: FullSupportBarDistribution.forward() returns the log-density of the bar distribution defined by logits.

src/tabpfn/inference_tuning.py
tobias-priorlabs: agreed

src/tabpfn/inference_tuning.py
tobias-priorlabs: agreed, dealt with in the refactor 5ad39cb

src/tabpfn/inference_tuning.py
tobias-priorlabs: agreed

src/tabpfn/inference_tuning.py
tobias-priorlabs: potentially? I suggest to differ for now.

src/tabpfn/regressor.py
tobias-priorlabs: agreed

src/tabpfn/regressor.py
tobias-priorlabs: Claude investigated this and found that the overwhelming time cost comes from the cross-validation, which exists for all types of tuning. I suggest making a progress bar a separate PR.
| n | folds | fit+predict plain | fit calibrated | of which: fold fits | of which: T sweep |
|---|---|---|---|---|---|
| 2000 (CPU) | 10 | 15.6 s | 66.4 s | 65.8 s | 0.35 s |
| 5000 (CPU) | 5 | 34.5 s | 112.3 s | 111.7 s | 0.36 s |
| 2000 (GPU) | 10 | ~2 s | 5.6 s | 4.6 s | 0.79 s |
| 20000 (GPU) | 2 | ~1.6 s | 3.5 s | 3.1 s | 0.17 s |

src/tabpfn/regressor.py
tobias-priorlabs: Peak memory is reached for large datasets, but for those we only do a single fold, i.e. the generator won't save us:

```
def get_default_tuning_n_folds(n_samples: int) -> int:
    n_samples_to_n_folds = {
        2_000: 10,
        5_000: 5,
        10_000: 3,
        20_000: 2,
        50_000: 1,
    }
    for n_samples_threshold, n_folds in n_samples_to_n_folds.items():
        if n_samples <= n_samples_threshold:
            return n_folds
    return 1
```

src/tabpfn/regression_metrics.py
oscarkey, line 1: Maybe this should just be `tabpfn.regression_metrics`? It's not really related to the architectures.

src/tabpfn/regressor.py
oscarkey: thank you for measuring these!
I guess we don't include a progress bar for plain fit/predict, which can also take ages, so I actually suggest we just leave it how it is for now.

src/tabpfn/regressor.py
oscarkey: ahhh okay well good as it is then :)

oscarkey (dismissed):
lgtm! Just a couple of minor things. Unfortunately I'll have to reapprove once those are fixed (I'm fixing that atm), but that will be super fast.

oscarkey (approved):

PR #1173, opened 2026-08-11
perf: drop the redundant fp64 copies in `clean_data`

arthur-priorlabs:
## Issue

## Motivation and Context

**What**

- `clean_data` no longer makes two redundant full-size float64 copies of the table.
- The predict-time frozen-encoder shortcut is now taken only for a frame that lines up with the fitted one.

**Why**

- Those copies set the RAM peak and most of the remaining CPU: 32 GB transient RSS and 16.8 s on a 5.33 GB float32 table.

**How**

- Drop the unconditional closing `astype(np.float64)`, stop re-copying an already-materialised multi-block frame, and skip the numeric reassign in `fix_dtypes`.
- Gate the shortcut on feature count and feature names, the same two grounds sklearn uses; anything else goes back through `transform`.

## Public API Changes

- [x] No Public API changes

### Highest supported version

Benchmarked with `scripts/bench_clean_data.py --reference main` from the `arthur/benchmark-clean-data` branch on a `cpuhighmem16spot` node (pandas 3.0.3). The harness also compares outputs against the reference on every run.

> [!WARNING]
> Time deltas are noisy (up to ~7% delta when A/A testing), but RSS isn't (0.5% delta)

Shapes:

- `numeric` is 666,667 x 2,000 float32 (5.33 GB);
- `half-string` and `numeric-object` are 333,333 x 400 object arrays.

"Transient RSS" is peak minus entry, i.e. what the call adds on top of what it was handed; wall time is the median of 3 repeats after 1 warmup.

| mix | transient RSS (main) | transient RSS (this) | Δ RSS | median wall (main) | median wall (this) | Δ time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` | 31.99 GB | 10.67 GB | **−66.7%** | 16.787 s | 5.472 s | **−67.4%** |
| `half-string` | 2.74 GB | 2.74 GB | +0.0% | 14.678 s | 13.990 s | −4.7% |
| `numeric-object` | 3.84 GB | 2.77 GB | **−27.9%** | 10.312 s | 9.418 s | −8.7% |

<details><summary>Benchmark details</summary>

| mix | side | entry RSS | peak RSS | retained | cold call | median | min | max | spread | stdev |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` | main | 6.00 GB | 37.99 GB | 10.67 GB | 17.370 s | 16.787 s | 16.598 s | 17.041 s | 2.6% | 222.0 ms |
| `numeric` | this | 6.00 GB | 16.67 GB | 10.67 GB | 5.436 s | 5.472 s | 5.433 s | 5.474 s | 0.7% | 22.9 ms |
| `half-string` | main | 3.88 GB | 6.63 GB | 1.06 GB | 14.595 s | 14.678 s | 14.652 s | 14.800 s | 1.0% | 78.7 ms |
| `half-string` | this | 3.88 GB | 6.63 GB | 1.06 GB | 14.382 s | 13.990 s | 13.806 s | 14.084 s | 2.0% | 141.5 ms |
| `numeric-object` | main | 5.97 GB | 9.82 GB | 1.06 GB | 10.340 s | 10.312 s | 10.291 s | 10.337 s | 0.4% | 22.7 ms |
| `numeric-object` | this | 5.97 GB | 8.74 GB | 1.06 GB | 9.695 s | 9.418 s | 9.411 s | 9.420 s | 0.1% | 5.0 ms |

</details>

Peak RSS on `numeric` falls 37.99 → 16.67 GB (−56.1%), and on `numeric-object` 9.82 → 8.74 GB (−11.0%).

### Lowest supported versions

| mix | transient RSS (main) | transient RSS (this) | Δ RSS | median wall (main) | median wall (this) | Δ time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` (66,667 x 2,000) | 4.31 GB | 1.07 GB | **−75.2%** | 253.827 s | 0.547 s | **−99.8%** |
| `half-string` | 3.36 GB | 3.36 GB | +0.0% | 196.368 s | 197.231 s | +0.4% |
| `numeric-object` | 4.40 GB | 3.32 GB | **−24.5%** | 23.706 s | 22.841 s | −3.6% |

<details><summary>Benchmark explanation</summary>

Re-run against the floors in `pyproject.toml`, resolved with `uv pip install --resolution lowest-direct`: **pandas 1.4.0, numpy 1.21.6, scikit-learn 1.2.0, torch 2.5.0, Python 3.10**. Same harness, same node type, outputs identical to `main` again on all three mixes. `numeric` is measured at 66,667 x 2,000 rather than 666,667 x 2,000 -- at the full shape `main` does not finish one side inside a three-hour allocation, for the reason below.

**Why `numeric` is a different kind of number there.** `fix_dtypes` ends in `X[numerical_columns] = X[numerical_columns].astype(numeric_dtype)`, and pandas < 3 implements a list-key assignment as one single-column `__setitem__` per column (`for k1, k2 in zip(key, value.columns): self[k1] = value[k2]`, `pandas/core/frame.py:3687` in 1.4.0). Each of those does block-manager work proportional to the block count, so the reassign is quadratic in column count.
Timed at 25,000 rows, one call:

| columns | `main` | this branch |
| ---: | ---: | ---: |
| 250 | 0.389 s | 0.022 s |
| 500 | 5.276 s | 0.052 s |
| 1,000 | 28.513 s | 0.143 s |
| 2,000 | 124.470 s | 0.303 s |

Every doubling costs `main` about 4x and this branch about 2x. Skipping the reassign
when it has nothing to do turns it off entirely, which is why the same change is worth
−67% on pandas 3 and −99.8% here.

`half-string` gains nothing on pandas 1.4 (+0.4%) because what it removes is a rounding
error at that scale: the same table takes 196 s there against 14 s on pandas 3, a
difference neither branch controls.

</details>

Added tests:

- `tests/test_preprocessing/test_data_cleaning.py`: the numeric shortcut converts without an intermediate copy, passes non-finites through, and matches the encoder path; `_owned_float64_values` hands a multi-block frame through and still copies a single-block one.
- The frozen-encoder shortcut is pinned three ways -- taken when the frame lines up, rejected for a narrower frame, and reordering like sklearn otherwise -- plus the inf-handling suite over the new numeric path.

## Checklist

### DIFF
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -425,49 +425,34 @@ def _encoding_is_identity(
     that selection has to be empty as well -- and the frame has to be one sklearn
     would have accepted, since skipping `transform` skips its checks along with it.
     """
-    if any(dtype != _FLOAT64 for dtype in X.dtypes):
-        return False
-    if fit_encoder or ord_encoder is None:
-        return True
-    # `transformers_` read directly rather than defaulted: only a fitted encoder gets
-    # this far -- `_align_columns_to_fitted_dtypes` runs first and goes through
-    # `named_transformers_`, which raises without it -- and a default would answer
-    # "identity" for an encoder that selected nothing only because it was never fit.
-    if any(
-        len(columns) > 0
-        for name, _, columns in ord_encoder.transformers_
-        if name != "remainder"
-    ):
-        return False
-    return _matches_fitted_frame(ord_encoder, X)
-
-
-def _matches_fitted_frame(
-    ord_encoder: OrderPreservingColumnTransformer, X: pd.DataFrame
-) -> bool:
-    """Whether `X` lines up with the frame the encoder was fitted on.
-
-    The shortcut hands back `X`'s own columns in `X`'s own order, so it is only the
-    same answer as `transform` for a frame that lines up with the fitted one:
-
-    * A different width `transform` refuses outright -- `columns are missing` when it
-      has feature names to check against, sklearn's feature-count error when not.
-    * The same names in another order it accepts and *reorders*, selecting by name
-      and returning the columns as they sat at fit. Handing back `X`'s order instead
-      would be a silently transposed result, so this is the case worth catching.
-
-    When only one of the two sides carries names sklearn falls back to lining the
-    columns up positionally, which is what the shortcut does as well, so that is let
-    through.
-    """
-    if getattr(ord_encoder, "n_features_in_", None) != X.shape[1]:
-        return False
-    fitted_names = getattr(ord_encoder, "feature_names_in_", None)
-    if fitted_names is None or not all(isinstance(col, str) for col in X.columns):
-        return True
-    # Compared as plain lists: the two sides can hold the same names in a numpy
-    # string array and a pandas Index, whose own equality is dtype-sensitive.
-    return list(X.columns) == list(fitted_names)
+    return (
+        # only passthrough non-nullable all-fp64 dataframes
+        all(dtype == _FLOAT64 for dtype in X.dtypes)
+        and (
+            # trainable encoders are ok
+            fit_encoder
+            or ord_encoder is None
+            or (
+                # condition 1: the ordinal encoder needs to be reductible to its
+                # passthrough remainder
+                all(
+                    len(columns) == 0
+                    for name, _, columns in ord_encoder.transformers_
+                    if name != "remainder"
+                )
+                # condition 2: X needs to line up with what the encoder was fitted
+                # condition 2.1: encoder needs to have the same input feature shape
+                and getattr(ord_encoder, "n_features_in_", None) == X.shape[1]
+                # condition 2.2: either no fitted feature names, or they match 1:1
+                and (
+                    getattr(ord_encoder, "feature_names_in_", None) is None
+                    or not all(isinstance(col, str) for col in X.columns)
+                    # compared as plain lists to be dtype-sensitive:
+                    or list(X.columns) == list(ord_encoder.feature_names_in_)  # ty: ignore[unresolved-attribute]
+                )
+            )
+        )
+    )
 
 
 def _to_numpy_may_alias(X: pd.DataFrame) -> bool:
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -433,21 +433,21 @@ def _encoding_is_identity(
             fit_encoder
             or ord_encoder is None
             or (
-                # condition 1: the ordinal encoder needs to be reductible to its
-                # passthrough remainder
+                # condition 1: the ordinal encoder has been fitted and needs to
+                # be reducible to its passthrough remainder
                 all(
                     len(columns) == 0
                     for name, _, columns in ord_encoder.transformers_
                     if name != "remainder"
                 )
-                # condition 2: X needs to line up with what the encoder was fitted
+                # condition 2: X needs to line up with what the encoder was fitted on
                 # condition 2.1: encoder needs to have the same input feature shape
                 and getattr(ord_encoder, "n_features_in_", None) == X.shape[1]
                 # condition 2.2: either no fitted feature names, or they match 1:1
                 and (
                     getattr(ord_encoder, "feature_names_in_", None) is None
                     or not all(isinstance(col, str) for col in X.columns)
-                    # compared as plain lists to be dtype-sensitive:
+                    # compared as plain lists to be dtype-insensitive:
                     or list(X.columns) == list(ord_encoder.feature_names_in_)  # ty: ignore[unresolved-attribute]
                 )
             )
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -435,21 +435,47 @@ def _encoding_is_identity(
     )
 
 
-def _owned_float64_values(X: pd.DataFrame) -> np.ndarray:
-    """`X`'s values as a freshly allocated, writeable float64 array.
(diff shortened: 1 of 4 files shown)


arthur-priorlabs:

- #1173 ◀
    - #1174

src/tabpfn/preprocessing/clean.py
```diff
@@ -112,6 +143,14 @@ def fix_dtypes(  # noqa: D103, C901, PLR0912
...
             # It's a numeric type, just wrap the array in pandas with the correct dtype
+            #
+            # A wrap, not a copy: when the dtype already matches, the frame returned
+            # from here is backed by the caller's buffer, so it does not own what it
+            # holds. Copy-on-write makes that safe by construction, but pandas 2 is
+            # still supported and there it is not -- a write into such a frame lands
+            # in the caller's array. Every write below either copies the frame first
+            # or replaces whole columns with a new dtype, and `_owned_float64_values`
+            # copies before handing the values back out; keep it that way.
             X = pd.DataFrame(X, copy=False, dtype=numeric_dtype)
```
jmkuebler: I'd tend to remove that comment. IIUC on Pandas 2, this is a reason it is slower. While we want to support older versions, I would say for performance, we do not specifically need to drag their legacy along. (If we deprecate pandas 2 support, this comment will become stale)

src/tabpfn/preprocessing/clean.py
```diff
@@ -173,7 +212,15 @@ def fix_dtypes(  # noqa: D103, C901, PLR0912
...
+    # it rewrites them as one block per column, and a fragmented frame has to be
+    # re-materialised (a full extra copy) by every later `to_numpy`. Skip it when
+    # they already hold the target dtype -- the common case for a numeric ndarray
+    # input, whose DataFrame was constructed with `numeric_dtype` above.
+    if (
+        len(numerical_columns) > 0
+        and not (X.dtypes[numerical_columns] == np.dtype(numeric_dtype)).all()
+    ):
         X[numerical_columns] = X[numerical_columns].astype(numeric_dtype)
     return X
```
jmkuebler, line 218: Comment is quite verbose. yet, without it, it is very hard to get this in the future. so maybe it is ok to keep. Your choice.

src/tabpfn/preprocessing/clean.py
```diff
@@ -368,6 +415,156 @@ def _inf_masks_dataframe(X: pd.DataFrame) -> tuple[np.ndarray, np.ndarray]:
...
+    through.
+    """
+    if getattr(ord_encoder, "n_features_in_", None) != X.shape[1]:
+        return False
+    fitted_names = getattr(ord_encoder, "feature_names_in_", None)
+    if fitted_names is None or not all(isinstance(col, str) for col in X.columns):
+        return True
+    # Compared as plain lists: the two sides can hold the same names in a numpy
+    # string array and a pandas Index, whose own equality is dtype-sensitive.
+    return list(X.columns) == list(fitted_names)
```
jmkuebler: maybe worthwhile to just do it inline with one or two comments. Otherwise too many lines of code comments for a simple function IMO.

src/tabpfn/preprocessing/clean.py
arthur-priorlabs: Maybe now is the time to drop `pandas<3` altogether

jmkuebler (commented):
Thanks a lot Arthur! The big gains seem to be achieved by the shorttrack for the cases where there is nothing to encode. This part is also simple to read and maintain. 

Does this already explain the speedups and savings we see for the case of numeric inputs?

On the other inputs the savings are arguably moderate and I am not sure it is then really worth it do introduce that much new code. What's your feeling about this?

If you think it is worthwhile, I'd still aim to get rid of some of the private methods and their verbose docstrings and just do them inline at their unique call sites.

src/tabpfn/preprocessing/clean.py
jmkuebler: I think pandas 3 was only released this year, so i guess we will support v2 for a while.

arthur-priorlabs:
Thanks for the review!

The simple fast-track path indeed corresponds to the huge `numeric` (=float-only) gains.

> On the other inputs the savings are arguably moderate

The reason the transient RAM numbers are much lower on `half-string` and `numeric-object` is that the input/output data itself takes much more memory per cell (cf. details section), and so I've had to use much smaller shapes.

I'd argue that -28% RAM usage is pretty significant at the 20B cell scale we're targeting.

Fyi, the mixed dtype paths come into play whenever we have categoricals (either numerical or string), which is pretty common.

src/tabpfn/preprocessing/clean.py
arthur-priorlabs: Fixed in b44676ef.

src/tabpfn/preprocessing/clean.py
arthur-priorlabs: Without it, I'd be the first to always cast everything in the next PR 😅

src/tabpfn/preprocessing/clean.py
arthur-priorlabs: Fixed in https://github.com/PriorLabs/TabPFN/pull/1173/commits/e01d537b8be3a811b584b195f38c0f3fbe208f91 + https://github.com/PriorLabs/TabPFN/pull/1173/commits/1d0f80bf5d2ea8a8dbb1dfa92fbb31f50b0fa2e8

jmkuebler (approved):
Thanks @arthur-priorlabs let's go ahead with this one already (no pandas versioning discussion here IIUC)

PR #1174, opened 2026-08-11
perf: assemble `clean_data`'s encoded array instead of stacking and reordering it

arthur-priorlabs:
## Issue

## Motivation and Context

**What**

- The encoded array is assembled column by column instead of being stacked by `ColumnTransformer` and then reordered.
- The assembly's precondition is checked before the inner encoder is fitted, not after.

**Why**

- The stack and the reorder gather each reached the RAM peak, which is why removing a later copy (#1173) moved the timing but not the peak.
- An input that fails the precondition was paying for a full pass over the categorical columns and then discarding it.

**How**

- Encoded columns are written straight into the output array at their own positions, so nothing needs stacking and nothing needs reordering.
- Column bookkeeping comes from a one-row structural fit (it makes no pass over the data); the inner encoder is then refit on every row for its categories.

## Public API Changes

- [x] No Public API changes

### Highest supported version

Benchmarked with `scripts/bench_clean_data.py --reference <parent branch>` from the `arthur/benchmark-clean-data` branch on a `cpuhighmem16spot` node (pandas 3.0.3). The harness also compares outputs against the reference on every run.

> [!WARNING]
> Time deltas are noisy (up to ~7% delta when A/A testing), but RSS isn't (0.5% delta)

Shapes:

- `numeric` is 666,667 x 2,000 float32 (5.33 GB);
- `half-string` and `numeric-object` are 333,333 x 400 object arrays, `half-string` with every other column a 20-level string.

"Transient RSS" is peak minus entry, i.e. what the call adds on top of what it was handed; wall time is the median of 3 repeats after 1 warmup.

| mix | transient RSS (parent) | transient RSS (this) | Δ RSS | median wall (parent) | median wall (this) | Δ time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` | 10.67 GB | 10.67 GB | +0.0% | 5.409 s | 5.373 s | −0.7% |
| `half-string` | 2.74 GB | 2.09 GB | **−23.6%** | 14.189 s | 13.602 s | −4.1% |
| `numeric-object` | 2.79 GB | 2.79 GB | −0.0% | 9.847 s | 9.687 s | −1.6% |

<details><summary>Benchmark details</summary>

| mix | side | entry RSS | peak RSS | retained | cold call | median | min | max | spread | stdev |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` | parent | 6.01 GB | 16.67 GB | 10.67 GB | 5.410 s | 5.409 s | 5.385 s | 5.415 s | 0.6% | 15.9 ms |
| `numeric` | this | 6.00 GB | 16.67 GB | 10.67 GB | 5.417 s | 5.373 s | 5.350 s | 5.381 s | 0.6% | 15.7 ms |
| `half-string` | parent | 3.88 GB | 6.62 GB | 1.06 GB | 14.305 s | 14.189 s | 14.076 s | 14.259 s | 1.3% | 92.0 ms |
| `half-string` | this | 3.88 GB | 5.97 GB | 1.06 GB | 13.736 s | 13.602 s | 13.506 s | 13.733 s | 1.7% | 113.7 ms |
| `numeric-object` | parent | 5.97 GB | 8.76 GB | 1.06 GB | 9.818 s | 9.847 s | 9.828 s | 9.851 s | 0.2% | 12.1 ms |
| `numeric-object` | this | 5.97 GB | 8.76 GB | 1.06 GB | 9.891 s | 9.687 s | 9.604 s | 9.703 s | 1.0% | 52.6 ms |

</details>

`half-string` is the only mix with categorical columns to encode, and the only one that moves: its peak RSS falls 6.62 → 5.97 GB (−9.8%). `numeric` and `numeric-object` never reach the new path.

### Lowest supported versions

| mix | transient RSS (parent) | transient RSS (this) | Δ RSS | median wall (parent) | median wall (this) | Δ time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `numeric` (66,667 x 2,000) | 1.07 GB | 1.07 GB | +0.0% | 0.528 s | 0.526 s | −0.4% |
| `half-string` | 3.36 GB | 2.71 GB | **−19.2%** | 187.979 s | 185.406 s | −1.4% |
| `numeric-object` | 3.32 GB | 3.32 GB | +0.1% | 24.408 s | 21.797 s | −10.7% |

<details><summary>Benchmark explanation</summary>

Re-run against the floors in `pyproject.toml`, resolved with `uv pip install --resolution lowest-direct`: **pandas 1.4.0, numpy 1.21.6, scikit-learn 1.2.0, torch 2.5.0, Python 3.10**. Same harness, same node type, outputs identical to the parent branch again on all three mixes. `numeric` is measured at 66,667 x 2,000, the shape #1173's lowest-version run is pinned to.

The memory win mostly survives on the old stack (−19.2% against −23.6%). The timing win does not (−1.4% against −4.1%): `half-string` costs 188 s there against 14 s on pandas 3, and the assembly's saving is small against that.

`numeric-object`'s −10.7% is not a win this PR can claim. That mix takes the identity path, which neither commit here touches, and the identical parent-branch code measured 22.841 s on one node and 24.408 s on another.

</details>

Added tests:

- `tests/test_preprocessing/test_data_cleaning.py`: the assembly path is taken for a mixed frame, and its result matches `ColumnTransformer.fit_transform` exactly -- array and fitted state (`n_features_in_`, `output_indices_`, `sparse_output_`, `transformers_`, `categories_`).
- Inputs that fail the precondition (`object` passthrough, duplicate column names) fall back to sklearn, and a test counts encoder fits over the full row set: two before, one now.

## Checklist

### DIFF
--- a/changelog/1174.changed.md
+++ b/changelog/1174.changed.md
@@ -1,0 +1,1 @@
+`fit()` uses less memory on tables with categorical columns: the encoded array is now assembled in place instead of being stacked and then reordered, cutting transient memory by about a quarter on a half-string table.
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -605,12 +605,14 @@ def _apply_ordinal_encoder(
         # inner encoder its categories from every row.
         ord_encoder.fit(X.iloc[:1])
         selected = _encoder_selection(ord_encoder)
+        if not _can_write_encoded_columns(X, selected):
+            # Bail before learning any categories: `fit_transform` learns them again
+            # from scratch, so doing it first would be a wasted pass over the data.
+            # Only the one-row fit above is lost, which costs no pass at all.
+            return ord_encoder.fit_transform(X)
         if selected:
             ord_encoder.named_transformers_["encoder"].fit(X[selected])
-        if _can_write_encoded_columns(X, selected):
-            return _encode_into_preallocated(X, ord_encoder, selected)
-        # Not assemblable: refit properly and let sklearn do the whole thing.
-        return ord_encoder.fit_transform(X)
+        return _encode_into_preallocated(X, ord_encoder, selected)
     if ord_encoder is not None:
         # Left on sklearn deliberately. `transform` also validates the frame against
         # the one seen at fit -- column count, names, order -- and the assembly above
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -224,6 +224,14 @@ def _column_kind(dtype: Any) -> str:
     return dtype.kind
 
 
+def _encoder_selection(ord_encoder: OrderPreservingColumnTransformer) -> list[Any]:
+    """The columns a fitted encoder takes, in the order it holds them."""
+    return next(
+        (cols for name, _, cols in ord_encoder.transformers_ if name == "encoder"),
+        [],
+    )
+
+
 def _is_single_float_block(X: pd.DataFrame) -> bool:
     """True if ``X`` is backed by a single contiguous numpy float block.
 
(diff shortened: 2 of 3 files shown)


arthur-priorlabs:

- #1173
    - #1174 ◀
        - #1180

jmkuebler (dismissed):
That one looks also good to me, thank you!

jmkuebler (approved):
reapprove after rebase

PR #1180, opened 2026-08-12
perf: efficient pandas dtype casting

arthur-priorlabs:
## Issue

## Motivation and Context

**Why**

- Assigning back is one `__setitem__` per column before pandas 3, and each rebuilds the block its column sits in: casting `c` columns out of a block costs `c(c+1)/2` column copies.
- `astype` holds a second full copy of a frame that already has a block per column, which cost `numeric-object` 47.7% more transient RSS when applied everywhere.

**How**

2 code paths:

- A shared block takes `astype` with a dtype mapping, which casts each column once (so it's never slower).
- When each column has its own block, we can release each input column as we replace it, which costs less memory.

## Public API Changes

- [x] No Public API changes

## How Has This Been Tested?

Benchmarked with `scripts/bench_clean_data.py --reference <parent branch>` from the `arthur/benchmark-clean-data` branch on a `cpuhighmem16spot` node. Outputs are identical to the parent branch on all six pairs.

| environment | mix | Δ transient RSS | Δ median wall |
| --- | --- | ---: | ---: |
| pandas 1.4.0 | `half-string` | +0.2% | 191.3 s → 19.2 s (**−90.0%**) |
| pandas 1.4.0 | `numeric-object` | 3.33 GB → 2.80 GB (**−16.0%**) | −2.3% |
| pandas 3.0.3 | `half-string` | +0.5% | −0.8% |
| pandas 3.0.3 | `numeric-object` | +0.1% | −5.8% |
| pandas 3.0.3 | `numeric` | +0.0% | −0.0% |

An A/A run of this harness puts its floor at ±3% on wall time and ≤0.5% on transient RSS, so the two bolded numbers are the result and the rest is the point: nothing on pandas 3 moves.

<details><summary>Details</summary>

The cast, timed on its own at 200,000 x 400 on pandas 1.4:

| frame the cast is handed | assign back | `astype` mapping |
| --- | ---: | ---: |
| one block for 400 columns (`pd.DataFrame(object_array)`, what the categorical cast sees) | 347.7 s, +805 MB | 9.5 s, +1 MB |
| a block per column (what `convert_dtypes` leaves, what the numeric cast sees) | 247 ms, +24 MB | 152 ms, +21 MB |

`X[cols] = X[cols].astype(dtype)` reaches `_setitem_array`, which loops `for k1, k2 in zip(key, value.columns): self[k1] = value[k2]` (`pandas/core/frame.py:3687` in 1.4.0). Each single-column assignment deletes that column out of its block, and `Block.delete` is an `np.delete` over the whole block -- free when the block holds one column, a full-block rebuild when it holds 400.

pandas 3 splits an object array into one block per column, so nothing there shares a block and the two spellings measure the same (133 ms against 128 ms). This is why the switch is on layout rather than on version: the layout is what the cost depends on, and pandas 3 simply never presents the bad one.

`half-string` on pandas 1.4 now takes 19.2 s against 12.9 s on pandas 3, where it took 191 s before. The remaining gap is the pure-pandas inf-mask path `PANDAS_FASTER_THAN_MIXED_PATH` selects, not this cast.

`numeric-object`'s −16.0% is not from the routing: with no categorical columns the old code still ran `X[[]] = X[[]].astype("category")`, and the helper returns early on an empty selection. Inferred from the code path rather than measured separately.

</details>

Added tests:

- `_cast_columns` is pinned from both sides: a shared-block frame must not be assigned into one column at a time (`__setitem__` patched to fail; skipped on pandas 3, which has no such frame), and a block-per-column frame must come back cast without the caller's dtypes changing.
- `fix_dtypes` no longer casts into a `DataFrame` handed to it, duplicate labels are cast rather than refused, and `coerce_nullable_dtypes_to_numpy` converts nullable and boolean columns while leaving its input alone.

## Checklist

### DIFF
--- a/changelog/1180.changed.md
+++ b/changelog/1180.changed.md
@@ -1,0 +1,1 @@
+`fit()` no longer slows to a crawl on wide tables with categorical columns under pandas < 3: the dtype casts no longer rebuild the frame one column at a time, which took minutes on a 333,333 x 400 table and now takes seconds.
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -55,18 +55,13 @@ def _cast_columns_share_a_block(
 ) -> bool:
     """Whether any of `columns` sits in a block that holds more than one column.
 
-    That is what makes assigning a cast back expensive: the block a column is deleted
-    from is rebuilt whole, so a column with a block to itself costs nothing and one
-    sharing a block of 400 costs 400 columns' worth of copying. Blocks holding no
-    column being cast are not touched, so they do not count.
-
-    A frame that does not expose its blocks in the shape expected here answers `True`
-    as well, sending it down the route that cannot go quadratic.
+    That would make assigning a cast back expensive: the block a column is deleted
+    from is rebuilt whole.
+    Since columns are assigned one at a time, casting `c` columns out of a block
+    costs `c(c+1)/2` column copies.
     """
-    blocks = getattr(getattr(X, "_mgr", None), "blocks", None)
-    if blocks is None:
-        return True
     try:
+        blocks = X._mgr.blocks
         columns_in_block = np.zeros(X.shape[1], dtype=np.intp)
         for block in blocks:
             positions = block.mgr_locs.as_array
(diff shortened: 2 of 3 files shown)


arthur-priorlabs:

- #1173
    - #1174
        - #1180 ◀

jmkuebler:
Let's park (close) this PR for now so we do not add specialization onto older pandas versions.

jmkuebler (dismissed):
LGTM, we aligned offline to add the specialization, given that pandas v2 will be used for the foreseeable future.

jmkuebler (approved):
stamp

PR #1183, opened 2026-08-14
Reduce peak memory during KV cache construction

eliott-kalfon:
## Issue + Motivation
- We observe a 50% peak GPU memory increase when building the cache
- This creates OOM when building large caches (e.g. 1M per estimator, 8 estimators, 33GB to 55 GB peak memory)
- This is inefficient and will put load on our servers too
- It increases hardware requirements of whoever uses the model

To fix this, we introduce two main changes:
- Quantize each layer at a time instead of materialising the full cache then compressing
- Save estimator cache to CPU after build, then load back to GPU to avoid having to maintain all estimator caches at the same time in memory

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

- Tested locally, removes the peak memory spike

---

## Checklist

---

Closes [RES-2162](https://linear.app/priorlabs/issue/RES-2162/quantize-each-layers-kv-cache-directly-as-opposed-to-once-after-the)

### DIFF
--- a/src/tabpfn/inference.py
+++ b/src/tabpfn/inference.py
@@ -1054,20 +1054,17 @@ class InferenceEngineExplicitKVCache(MultiDeviceInferenceEngine):
             save_peak_memory_factor=DEFAULT_SAVE_PEAK_MEMORY_FACTOR
             if save_peak_mem
             else None,
+            kv_cache_dtype=(
+                KV_CACHE_PRECISION_DTYPES[kv_cache_precision]
+                if kv_cache_precision != "auto"
+                else None
+            ),
         )
 
         with (
             get_autocast_context(device, enabled=autocast),
             torch.inference_mode(),
         ):
-            cache_dtype = (
-                KV_CACHE_PRECISION_DTYPES[kv_cache_precision]
-                if kv_cache_precision != "auto"
-                else None
-            )
-            cache_build_kwargs = {}
-            if "kv_cache_dtype" in signature(model.forward).parameters:
-                cache_build_kwargs["kv_cache_dtype"] = cache_dtype
             _, cache = model(
                 X,
                 y,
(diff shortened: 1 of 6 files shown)


eliott-kalfon:
@jmkuebler CI for TabPFN extensions is an issue again, I am fixing the issue by incorporating the latest TabPFN changes into it so that we do not need to compute the read out in the extension repo anymore

src/tabpfn/inference.py
```diff
@@ -977,7 +987,13 @@ def __init__(  # noqa: PLR0913
             ),
             devices,
         )
-        self.kv_caches: list = list(timed_caches)
+        built_caches = list(timed_caches)
+        if stage_caches_on_cpu:
+            # Keep already-built caches out of device memory while later ensemble
+            # members are constructed, then restore their original placement.
+            self.kv_caches: list = [cache.to(device) for cache, device in built_caches]
```
priorphil, line 995: Does this match the comment? IIUC this moves the cache to device, so doesn't keep it out of device memory?

src/tabpfn/inference.py
```diff
@@ -991,7 +1007,8 @@ def _build_cache(
         gpu_preprocessor: TorchPreprocessingPipeline | None,
         autocast: bool,
         save_peak_mem: bool,
-    ) -> KVCache:
+        stage_cache_on_cpu: bool,
+    ) -> tuple[KVCache, torch.device]:
```
priorphil, line 1012: Can we add a Returns section to explain which device is returned and how it differs from the device of the tensors in the cache? (E.g., if it just matches KVCache[first_key].device, we could just read it from there)

priorphil (dismissed):
LGTM once comments are resolved

jmkuebler:
Thanks a lot for doing this @eliott-kalfon ! LGTM, I also did some measurements on RTX Pro 6k to check it doesn't make things relevantly slower ✅ 

I think the introduction of the `kv_compute_dtype` is a little clumsy, but that's what it is given that we need to support autocast, fp32, fp16 etc.. so that's the easiest way to detect the relevant precision here.

eliott-kalfon:
> Thanks a lot for doing this @eliott-kalfon ! LGTM, I also did some measurements on RTX Pro 6k to check it doesn't make things relevantly slower ✅
> 
> I think the introduction of the `kv_compute_dtype` is a little clumsy, but that's what it is given that we need to support autocast, fp32, fp16 etc.. so that's the easiest way to detect the relevant precision here.

Yes, I did the same calculations on RTX, with my build, this PR made it even slightly faster. Curious to hear what you found. Also, if you have a less clumsy method, feel free to let me know :)

priorphil (approved):
Awesome, thanks for clarifying the comments :)

PR #1186, opened 2026-08-18
perf: remove redundant copies in ensemble preprocessing

arthur-priorlabs:
## Issue

[RES-2592: optimize `TabPFNEnsemblePreprocessor.fit_transform_ensemble_members_iterator` RAM usage](https://linear.app/priorlabs/issue/RES-2592/optimize-tabpfnensemblepreprocessorfit-transform-ensemble-members) — follows the wrapper host-RAM work: with `clean_data` brought down, this call is the largest remaining contributor to the peak.

## Motivation and Context

**What**

-   `TabPFNEnsemblePreprocessor.fit_transform_ensemble_members_iterator` held four full-size copies of its input above the table it hands back, none of which changed a value.
-   Dropping them takes transient RSS from **42.7 GB to 12.0 GB** on the profiled shape, and the wall time with it.

**Why**

-   At 666,667 x 2,000 float64 this call was measured holding 4.0x the table it returns — the single largest remaining contributor to the wrapper's host-RAM peak.
-   The copies dominated its runtime too, so this is as much a latency change as a memory one.

**How**

-   Skip the rebuild wherever it cannot change a value: the redundant copy in `_fit_preprocessing_one`, an all-true mask select, an identity reshape `ColumnTransformer` (what a GPU-scheduled transform leaves behind), and a whole-table encoder passthrough.
-   First pin Fortran order at the ARPACK SVD's entrance, so those skips cannot rotate its basis — see the note below, this is the one part that is not obviously free.

|                              | numeric                       | half-string                | numeric-object             |
| :--------------------------- | :---------------------------- | :------------------------- | :------------------------- |
| input size              | 10.67 GB (666,667 x 2,000)    | 1.07 GB (333,333 x 400)    | 1.07 GB (333,333 x 400)    |
| categoricals after cleaning   | 0 of 2,000                    | 200 of 400                 | 0 of 400                   |
| transient RSS, highest       | 42.66 -> 12.00 GB (**-71.9%**) | 4.40 -> 3.33 GB (**-24.2%**) | 4.26 -> 1.07 GB (**-74.9%**) |
| transient RSS, lowest-direct | 42.66 -> 12.00 GB (**-71.9%**) | 3.73 -> 3.20 GB (**-14.3%**) | 4.26 -> 1.07 GB (**-75.0%**) |
| median wall, highest         | 34.80 -> 11.09 s (**-68.1%**)  | 7.71 -> 7.37 s (-4.4%)     | 2.86 -> 0.70 s (**-75.6%**)  |
| median wall, lowest-direct   | 36.22 -> 11.28 s (**-68.8%**)  | 14.52 -> 14.37 s (-1.0%)   | 2.86 -> 0.76 s (**-73.4%**)  |

Each cell is a complete A/B against `main` on a dedicated `cpuhighmem16spot` node, three timed repeats, with the ensemble members compared cell by cell: **bit-identical in all six**. Mixes carry their own default shapes, so cells compare down a column, not across one. The two half-string wall-time deltas are inside the ±3% run-to-run noise floor and should be read as "unchanged".

<details><summary>Why half-string gains least, and where the remaining 12 GB is</summary>

`half-string` is the case where both skips are inapplicable by construction: it has genuine categorical columns, so the encoder really runs, and the reshape layout puts those columns first, so it is not the identity permutation either. Only the two copy removals apply, and its peak is set by the encoder's `ColumnTransformer`, which this PR does not touch — so it is the honest floor of the change rather than a surprise. Its RSS delta also moves with the dependency set (-24.2% vs -14.3%) because the baseline peak itself does (4.40 vs 3.73 GB): older pandas/numpy build that path differently.

After this change the 12.00 GB is the 10.67 GB table the call returns plus ~1.3 GB of one-byte-per-cell boolean masks (`isinf`, `isnan`), i.e. the peak is now the output itself. The `ENABLE_GPU_PREPROCESSING=False` path is unchanged at 4.95x (measured neutral: RSS +0.1%, wall -2.1%), because its peak sits in the squashing-scaler `ColumnTransformer` that none of these skips reach. That is the next target on that path, and it is the path released presets take.

</details>

<details><summary>The ARPACK layout trap (why the first commit exists)</summary>

`TruncatedSVD(algorithm="arpack")` is an iterative Lanczos solver. On the near-degenerate spectra wide tables produce — measured here at 124.14 / 123.65 / 123.08 — the basis it converges to depends on whether the array it is handed is C- or Fortran-contiguous: identical values, identical subspace (principal angles below 2e-6 degrees) and identical singular values (1.9e-15 relative), different basis.

Nothing picks that layout deliberately: sklearn's column indexing and `hstack` both return Fortran-contiguous arrays, so every step that rebuilds the table happens to hand ARPACK Fortran order. Removing such a rebuild therefore rotates these features, and every released preset configures this step (`global_transformer_name="svd"`, `ENABLE_GPU_PREPROCESSING` defaults to False), so it would have reached users. Neither the fit-side benchmarks nor a local test run would have shown it: `tests/test_consistency.py`'s reference predictions are skipped off the enabled platforms.

`_pin_layout` therefore pins Fortran order at the solver's entrance. It is a literal no-op today — verified on numpy 1.22.0 and 2.5.2, where `X[:, mask]` is Fortran-contiguous either way — and it means the next change to an earlier step no longer has to know any of this.

</details>

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

| suite                                                                                             | result                    |
| :------------------------------------------------------------------------------------------------ | :------------------------ |
| `tests/test_preprocessing`, `tests/test_torch_preprocessing`                                      | 760 passed, on both ends of the dependency range |
| `tests/test_consistency.py`, classifier/regressor interface, `tests/test_inference.py`            | 314 passed (reference predictions skipped on this platform) |
| A/B vs `main`, 6 cells above, plus `--n-estimators 3` and `--no-gpu-preprocessing`                 | members bit-identical, no regression |

-   No new tests: every change is output-preserving by construction and gated on that by the A/B above, which compares each ensemble member's preprocessed matrix, labels, schema, feature indices and pipeline shape against the reference commit.
-   Checked out-of-band that the torch (GPU-scheduled) pipeline is bitwise identical across input layouts, since the fit-side arrays now arrive C- rather than Fortran-contiguous, and that path runs at inference time where the benchmarks do not look.

<details><summary>Environments and how to reproduce</summary>

-   `highest`: python 3.14.5, numpy 2.5.2, pandas 3.0.5, scikit-learn 1.9.0, torch 2.13.0
-   `lowest-direct`: python 3.10.20, numpy 1.22.0, pandas 1.4.0, scikit-learn 1.2.0, torch 2.5.0 — matching the `lowest-direct` leg of CI

Measured with `scripts/bench_ensemble_preprocessing.py --reference main`, which is not part of this branch. Note for anyone rebuilding the floor environment: `torch.utils.benchmark` imports `cpp_extension` -> `setuptools` on torch 2.5, and nothing in the dependency floor pulls `setuptools` in, so the benchmark cannot import until it is installed. CI does not hit this because its tests never import that module.

</details>

---

## Checklist

---

### DIFF
--- a/changelog/1186.changed.md
+++ b/changelog/1186.changed.md
@@ -1,0 +1,1 @@
+Reduced peak host memory during preprocessing: the ensemble preprocessor no longer rebuilds the feature matrix in steps that cannot change it, taking transient RSS from 42.7 GB to 12.0 GB (-72%), and wall time with it, on a 666,667 x 2,000 float64 fit. Preprocessed outputs are unchanged.
--- a/src/tabpfn/preprocessing/steps/add_svd_features_step.py
+++ b/src/tabpfn/preprocessing/steps/add_svd_features_step.py
@@ -4,9 +4,10 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Literal
+from typing import Literal
 from typing_extensions import override
 
+import numpy as np
 from sklearn.decomposition import TruncatedSVD
 from sklearn.pipeline import Pipeline
 from sklearn.preprocessing import StandardScaler
@@ -16,8 +17,24 @@ from tabpfn.preprocessing.pipeline_interface import PreprocessingStep
 from tabpfn.preprocessing.steps.utils import make_scaler_safe
 from tabpfn.utils import infer_random_state
 
-if TYPE_CHECKING:
-    import numpy as np
+
+def _pin_layout(X: np.ndarray) -> np.ndarray:
+    """Return ``X`` in the memory layout this step's solver is calibrated on.
+
+    ``TruncatedSVD(algorithm="arpack")`` is an iterative Lanczos solver, and on a
+    near-degenerate spectrum -- singular values within a fraction of a percent of each
+    other, which wide tables routinely produce -- the basis it converges to depends on
+    the order the underlying BLAS accumulates in, and so on whether the array it is
+    handed is C- or Fortran-contiguous. The values are unaffected and so is the
+    subspace; which basis of it comes back is not.
+
+    Upstream steps decide that layout only by accident (sklearn's column indexing and
+    ``hstack`` happen to return Fortran-contiguous arrays), so without pinning it here
+    an unrelated change to an earlier step silently rotates these features. Fortran
+    order is what the pipeline has always produced, so pinning it keeps the components
+    that were previously returned. A no-op when the input already has it.
+    """
+    return np.asfortranarray(X)
 
 
 def get_svd_n_components(
@@ -99,7 +116,7 @@ class AddSVDFeaturesStep(PreprocessingStep):
             n_features,
             random_state=static_seed,
         )
-        transformer.fit(X)
+        transformer.fit(_pin_layout(X))
 
         self.transformer_ = transformer
         self.feature_schema_updated_ = feature_schema
@@ -116,7 +133,8 @@ class AddSVDFeaturesStep(PreprocessingStep):
         assert self.feature_schema_updated_ is not None
         assert self.transformer_ is not None
 
-        return X, self.transformer_.transform(X), FeatureModality.NUMERICAL
+        # Only the solver's input is pinned; `X` itself is handed back as it came.
+        return X, self.transformer_.transform(_pin_layout(X)), FeatureModality.NUMERICAL
 
 
 def get_svd_features_transformer(
--- a/src/tabpfn/preprocessing/steps/add_svd_features_step.py
+++ b/src/tabpfn/preprocessing/steps/add_svd_features_step.py
@@ -19,20 +19,17 @@ from tabpfn.utils import infer_random_state
 
 
 def _pin_layout(X: np.ndarray) -> np.ndarray:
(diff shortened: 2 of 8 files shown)


arthur-priorlabs:

- #1186 ◀
    - #1187

bejaeger (dismissed):
Nice!

bejaeger:
One question, do we have a test that ensures we never mutate the input dataset? If not, let's add it here as a safeguard.

arthur-priorlabs:
We have a defensive copy in `PreprocessingPipeline._process_steps`, so we don't mutate the dataset input to the wrapper, cf. [code.](<https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/preprocessing/pipeline_interface.py#L455-L455>)

On the other hand, we don't check that individual steps don't mutate their outputs, which IMO is a good thing: that's where we could get further speedups.

bejaeger (approved):

PR #1189, opened 2026-08-19
cache decoder keys instead of the embeddings [RES-2474]

jmkuebler:
This PR changes the caching to only cache the training keys used for the many-class decoder instead of the train embeddings as this is 25% smaller than the full embeddings. For regression we completely remove the caching of the embeddings as they are  not used anymore. 
We adjust the kv_cache_size computation accordingly.

**This change breaks one functionality that we had before,** namely reading out the _train_ embeddings when the model was fitted with cache. You can still read out the test embeddings though. We now fail louldy if users try that and tell them to extract train embeddings from a fit without cache.

I checked w/ TabPFN-extensions and it does not seem to be an issue there.

## Motivation and Context

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?
I tested also in TabPFN-extensions and found no users of the corner case cache + train embeddings.
---

## Checklist

---

### DIFF
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -517,15 +517,7 @@ class ManyClassDecoder(nn.Module):
         self.softmax_scaling_layer = softmax_scaling_layer
 
     def project_keys(self, train_embeddings: torch.Tensor) -> torch.Tensor:
-        """Project train embeddings to per-head keys: (B, N, E) -> (B, N, H, D).
-
-        Split out of the forward pass so cached inference can run it once when
-        the cache is built and store the result (see
-        :attr:`TabPFNV3Cache.decoder_keys`), instead of re-projecting every
-        train row on every predict. The projection is a plain linear map with
-        no dependence on the test rows or on ``N``, so the cached keys are
-        exactly what the forward pass would have recomputed.
-        """
+        """Project train embeddings to per-head keys: (B,N,E)->(B,N,H,D)."""
         k_BNE = self.k_projection(train_embeddings)
         h, d = self.num_heads, self.head_dim
         return k_BNE.view(*k_BNE.shape[:2], h, d).contiguous()
--- a/src/tabpfn/architectures/tabpfn_v3.py
+++ b/src/tabpfn/architectures/tabpfn_v3.py
@@ -203,8 +203,12 @@ class TabPFNV3Cache(KVCache):
 
     Attributes:
         kv: Per-layer KV cache for the ICL transformer blocks.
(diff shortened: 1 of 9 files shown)


jmkuebler:
@cursor review

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -357,8 +359,14 @@ def get_cache_size(
         # One scalar scale per key and per value tensor.
         total_bytes += model_config.nlayers * 2 * kv_scale_dtype.itemsize
 
-    # 2. Train activations, (N_train, icl_emsize). Cached for both tasks.
-    total_bytes += n_train * icl_emsize * train_emb_dtype.itemsize
+    # 2. Many-class decoder keys, (N_train, H_dec * D_dec). Classification only:
+    # regression has no many-class decoder, so its cache omits this term. This
+    # mirrors how the architecture picks its task type from the config.
+    if model_config.max_num_classes >= 2:
```
bejaeger: nit: could add a small helper "\_is_classification" and re-use

src/tabpfn/architectures/tabpfn_v3.py
```diff
@@ -1938,30 +1951,50 @@ def forward(
...
+        if self.task_type == "multiclass":
+            if running_from_cache:
+                assert kv_cache is not None
+                assert kv_cache.decoder_keys is not None, (
+                    "A multiclass KV cache must carry the decoder keys."
+                )
+                train_keys = kv_cache.decoder_keys
+            else:
+                assert train_emb is not None
+                train_keys = self.many_class_decoder.project_keys(train_emb)
```
bejaeger, line 1979: nit: is there a reason I'm missing why we build the cache before the many_class_decoder runs?<br>Would maybe be cleaner to call the many_class_decoder once only in this main forw and return the keys in the forward call, similar to how we deal with it in the blocks. I'll let you decide.

bejaeger (dismissed):
Looks great! Approving but please consider my two comments.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: I opted for this design, because then the ManyClassDecoder can be completely agnostic to the cache. The downside imo is that now that train keys need to be computed outside the MCD. But I'd prefer to keep it as is.

src/tabpfn/architectures/tabpfn_v3.py
jmkuebler: done added as a property

bejaeger (approved):
Thanks!

PR #1195, opened 2026-08-21
Raise an exception if download_all_models() fails.

oscarkey:
At the moment we log but don't raise. This causes the true problem to be hidden in our CI pipelines, as the download stage appears to succeed and then the test fails later.

This is a potentially breaking change for existing users, as the function will now raise if they haven't accepted the license for one of the models, even if they're not using that model. However:
- This function is not used by TabPFNClassifier or TabPFNRegressor. The user has to explicitly discover it from the readme and then call it (e.g. if they want to work offline later).
- The function now only raises at the end, so it will still try to download all models.

### DIFF
--- a/changelog/1195.breaking.md
+++ b/changelog/1195.breaking.md
@@ -1,0 +1,1 @@
+`tabpfn.model_loading.download_all_models()` now raises an exception if one or more of the models fails to download. It will still download all possible models before raising the exception.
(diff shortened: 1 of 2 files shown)


ggprior (dismissed):
I remember some back and forth on this back when we tried to get this behavior the last time, but I'm strongly in favour of failing here, so LGTM

PR #1196, opened 2026-08-21
Apply regression target transforms to the unnormalized target [RES-2639]

bejaeger:
Closes RES-2639.

## Problem

The regressor z-normalises the target before the ensemble preprocessing, so every entry of `REGRESSION_Y_PREPROCESS_TRANSFORMS` acted on standardized values rather than on the target. `1_plus_log` was `log1p` of a z-score, not of the target — undefined wherever the z-score drops below -1, which is 15% of the rows for a symmetric target, and those NaNs reached the model unguarded because the log presets have no trailing scaler.

## Change

`wrap_target_transform` composes each transform into a pipeline that

1. undoes the z-normalisation, so the transform sees the target in its original units,
2. applies the transform,
3. standardizes the result — the scale the bar distribution's borders assume.

The pipeline's input and output stay in the z-normalised space, so nothing downstream moves: `inverse_transform` maps the model's bar-distribution borders straight back into z-space, exactly as an unwrapped transform did, and the `±1e3` sanity limits applied to those borders keep the z-units they were tuned in. That is why `predict`, `predict_batched` and the border mapping are untouched.

The fine-tuning data pipeline re-splits and re-normalises per split, so it rebinds the statistics before fitting. `"none"` is now treated as no transform at all, since composing the identity would only add float noise.

Also fixes the docstring of `_transform_labels_one`, which called its input "unprocessed" (the second item on the ticket).

## Impact

Changes the v2 / v2.5 / v3 defaults, whose ensembles use `safepower` for half of their estimators — up to ~4% relative prediction change on california housing. **Bit-identical** for the v2.6 default (`("none",)`), for `n_estimators=1` on any version, and for every estimator whose target transform is `None`.

## Tests

New `tests/test_preprocessing/test_target_transform.py` (17 cases) plus 4 regressor-level tests; two of them fail on `main` and pass here. Full local sweep of regressor + classifier interface, preprocessing, consistency, finetuning-regressor, save/load and inference-tuning suites passes.


bejaeger:
Superseded by the #1198 → #1199 stack, which removes the estimator's up-front z-normalisation of the target instead of composing it into each target pipeline. Same semantics as this PR (verified to 1e-7 relative on five datasets, so the TabArena validation transfers), without the baked-in statistics or the rebinding they forced on the fine-tuning path.

PR #1201, opened 2026-08-23
Convert columns to the type their contents imply, using skrub

alanprior:
## Issue
No corresponding issue. skrub's own type detection already runs server-side; this brings it to the public library.

## Motivation and Context

The public package ordinal-encodes anything that is not already numeric, so a column of `"39.00"` or a date string reaches the model as alphabetical rank instead of its real value, and a `datetime64` column raises outright. `fit()` now converts each column to the type its contents imply, using [skrub](https://skrub-data.org), before validation: a numeric string becomes a number regardless of any setting, a datetime column can expand into calendar features, and a text column can be encoded with tf-idf over character n-grams followed by a truncated SVD. Numeric columns and low-cardinality strings are passed through, so categorical detection and encoding are unchanged.

Date expansion and text encoding ship off by default. They are real, tested, and available now, but off until they have run more broadly, so they can be switched on later without a release that changes predictions for everyone.

---

## Public API Changes

-   [ ] No Public API changes

Four new `InferenceConfig` fields, so each checkpoint can decide what it needs rather than adding constructor parameters: `USE_DATES` and `USE_TEXT` (both default `False`), `MIN_CARDINALITY_FOR_TEXT` (default 30, named to match the existing `MAX_UNIQUE_FOR_CATEGORICAL_FEATURES`/`MIN_UNIQUE_FOR_NUMERICAL_FEATURES`), and `TEXT_N_COMPONENTS` (default 30). Set via `inference_config={"USE_DATES": True, ...}`. The fitted converter is exposed as `input_converter_`. Turning `USE_DATES`/`USE_TEXT` on grows `n_features_in_` and `feature_names_in_`, since they describe the converted frame. A column of numbers stored as strings is read as numbers regardless of these settings, which changes predictions on tables that have one. Minimum supported versions rise to match skrub's own floors: `scikit-learn>=1.4.2`, `pandas>=1.5.3`, `numpy>=1.23.5`.

---

## How Has This Been Tested?

Unit tests in `tests/test_preprocessing/test_input_conversion.py`, including every `USE_DATES`/`USE_TEXT` combination against a frame that mixes a native datetime column, date strings, numeric strings, text, categories and an all-null column at once, both standalone and through `TabPFNClassifier.fit`/`predict` with `inference_config`. Full suite passes locally.

---

## Checklist

---


alanprior:
@bejaeger Hi! This is right now a high-level proposal of implementation, we could think about it as a draft. 

To avoid optimization of the wrong implementation, I would love to know if the general direction makes sense. Don't bother on reviewing small specifics - this would be improved and handed to you later.

This seems to be working - there might be edge-cases, but I've run it over of dozen of datasets with dates/texts and performance does improve consistently as expected.

alanprior:
@bejaeger after a few iterations, I reimplemented things and starting to do in smaller chunks. This PR will either be made a draft, closed or evolve; for now, removing you from it as a reviewer.

alanprior:
We broke this down to many components and smaller PRS, and changed a lot of the code, closing

PR #1203, opened 2026-08-24
Stop pandas.to_numeric from segfaulting on adversarial strings

### LINKED ISSUE #1202: Segmentation fault in modality detection on certain numeric-looking strings (pandas.to_numeric overflow)
### Describe the bug

`fit()` segfaults on some tables, with no Python exception and no traceback, when a string column reaches `tabpfn.preprocessing.modality_detection._is_numeric_pandas_series`. That function calls `pandas.to_numeric(column, errors="coerce")` on the raw column with no guard. On the dependency versions below, `pandas.to_numeric` itself crashes the process (`SIGSEGV`) on any string matching `<digit>e<exponent>` where `exponent` is in `[2**31, 2**32)` — a signed 32-bit integer overflow in the underlying C parser for scientific notation, unrelated to TabPFN's own logic. A real-world trigger is a hash-like ID column that happens to contain such a substring, e.g. `"8e2569614270f3d8b9e7038efac9f116"` (read as mantissa `8`, exponent `2569614270`, which is in the crashing range).

This is a segfault, so it cannot be caught with `try`/`except`; the process just dies. I have not reproduced it on newer `pandas`/`numpy` (`3.0.5`/`2.5.2` on a different machine/OS), so it may already be fixed upstream, but I have not confirmed that against a changelog, and it may also be an OS/architecture difference rather than a version fix.

### Steps/Code to Reproduce

```python
import numpy as np
import pandas as pd

n = 60
ids = [f"id_{i}" for i in range(n)]
ids[5] = "8e2569614270f3d8b9e7038efac9f116"  # crashes pandas.to_numeric on affected versions
X = pd.DataFrame({"id": ids, "x": np.arange(n, dtype=float)})
y = np.arange(n, dtype=float)

TabPFNRegressor(device="cpu", use_text=False).fit(X, y)
```

Narrower repro of the actual crash, no TabPFN involved at all:

```python
import pandas as pd
pd.to_numeric(pd.Series(["8e2569614270"]))  # segfaults
```

Boundary check on the same machine:

```python
import pandas as pd
for exp in (2147483647, 2147483648, 4294967295, 4294967296):  # 2**31-1, 2**31, 2**32-1, 2**32
    print(exp, pd.to_numeric(pd.Series([f"1e{exp}"])).tolist())
```
`2147483647` and everything at/after `4294967296` return safely; `2147483648` segfaults.

### Expected Results

`fit()` either succeeds, or raises a catchable Python exception. No table content should be able to crash the interpreter.

### Actual Results

```
Fatal Python error: Segmentation fault

Current thread ...:
  File ".../pandas/core/tools/numeric.py", line 235 in to_numeric
  File ".../tabpfn/preprocessing/modality_detection.py", line 229 in _is_numeric_pandas_series
  File ".../tabpfn/preprocessing/modality_detection.py", line 199 in _detect_feature_modality
  File ".../tabpfn/preprocessing/modality_detection.py", line 80 in detect_feature_modalities
  File ".../tabpfn/regressor.py", line 921 in _initialize_dataset_preprocessing
  File ".../tabpfn/regressor.py", line 1205 in fit
```
(paths trimmed to the site-packages-relative parts; captured with `faulthandler` enabled, otherwise the process just dies with exit code -11 and no output)

### Versions

```
tabpfn: 8.4.0
torch: 2.9.1
numpy: 2.1.3
scipy: 1.15.3
pandas: 2.3.3
scikit-learn: 1.6.1
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 0.36.2
Python: 3.11.14, Linux-6.14.0-1021-gcp-x86_64, glibc-2.39
```

alanprior:
## Issue

Fixes PriorLabs/TabPFN#1202.

## Motivation and Context

`fit()` could crash the interpreter outright (`SIGSEGV`, no Python traceback) on a table where a string column holds a hash-like value such as `"8e2569614270f3d8b9e7038efac9f116"`. Modality detection called `pandas.to_numeric` on the raw column to ask whether it was numeric; below pandas 3.0 that function has a signed 32-bit integer overflow parsing a scientific-notation exponent in `[2**31, 2**32)`, and segfaults rather than raising ([pandas-dev/pandas#63650](https://github.com/pandas-dev/pandas/issues/63650), fixed upstream in pandas 3.0). A segfault cannot be caught with `try`/`except`, so the only fix is to never hand such a value to pandas in the first place.

## How

The change is confined to `_is_numeric_pandas_series`, the one function implicated in the crash. It only ever consumed `coerced.notna()`, so it now answers with a bool and never builds a numeric series:

```python
def _is_numeric_pandas_series(s: pd.Series) -> bool:
    if pd.api.types.is_numeric_dtype(s.dtype):
        return True
    if PANDAS_BELOW_3:
        return all(_is_numeric_or_missing_for_old_pandas(value) for value in s)
    coerced = pd.to_numeric(s, errors="coerce")
    is_numeric_or_missing = coerced.notna() | s.isna()
    return bool(is_numeric_or_missing.all())
```

On pandas 3.0 and later the function is byte-identical to `main`, since the upstream fix makes the workaround unnecessary there.

`_is_numeric_or_missing_for_old_pandas` reads one value with the built-in `float`, which does not share the bug. `float` is more permissive than pandas' parser, so four spellings are rejected explicitly: the literal `"nan"`, PEP 515 digit separators (`"1_000"`), non-ASCII digits and spaces (`"٣"`, `"\xa0 5"`), and finite literals that overflow to infinity (`"1e400"`). Buffers other than `bytes` are rejected for the same reason.

## Decisions

**Why the version gate, rather than one code path.** The pandas majors genuinely disagree, so a single hand-rolled parse cannot match both: `"1e400"` reads as NaN on 2.x and as infinity on 3.x, from the same rewritten parser that fixed the segfault. Gating keeps exact pandas semantics on 3.x and confines the stand-in to the version that actually crashes.

**Why only modality detection.** An earlier revision routed both `pandas.to_numeric` call sites through a shared helper. Only modality detection was ever implicated in the crash, and sharing forced the two call sites to agree on an output dtype they do not need to share. That produced an unconditional `float64` cast which, as @arthur-priorlabs found, collides int64 categories above `2**53` in `_align_columns_to_fitted_dtypes` and silently reassigns their encoder codes. `clean.py` is therefore untouched apart from renaming `PANDAS_FASTER_THAN_MIXED_PATH` to `PANDAS_BELOW_3`, which is what it actually tests.

**Reading one value at a time is not the cost it looks like.** Returning a bool lets the scan stop at the first non-numeric value, which is the first value for a text column, i.e. the shape that carries the crashing token. Benchmarked with `torch.utils.benchmark` on pandas 2.x over 1M-row object columns, against `main` as the baseline: `non-numeric` and `mostly-numeric` drop to ~0, `int-strings` 0.7x, `with-missing` 1.0x, `float-strings` and `sci-strings` 1.2x. The one regression is `numbers-as-objects` at 2.9x (+48 ns/value), roughly 0.5 s per 10M cells. Restoring a gate to skip it was measured and rejected: it recovers that kind but gives up the free exit on `non-numeric`, which is the crash case.

## Follow-ups

`_align_columns_to_fitted_dtypes` can still reach `pandas.to_numeric` with a crashing value, exactly as on `main` today. Fixing it needs the int64 narrowing preserved, so it is a separate change rather than part of this one.

---

## Public API Changes

- [X] No Public API changes

---

## How Has This Been Tested?

`tests/test_preprocessing/test_numeric_parsing_below_pandas_3.py` is a dedicated module, deletable in one go once the pandas floor reaches 3.0. It covers the spellings pandas accepts and rejects, missing values, non-string scalars, bytes and other buffers, and container cells, plus both branches of `_is_numeric_pandas_series` and a stub asserting the below-pandas-3 branch never reaches `pandas.to_numeric`. Its expectation table is asserted twice, once against the parse and once against `pandas.to_numeric` itself, so a wrong expectation cannot be satisfied by a matching bug in the stand-in.

The parse was additionally checked against `pandas.to_numeric` on pandas 2.3.3 over 20075 values (75 hand-picked plus 20000 fuzzed) with no disagreements, and the original #1202 reproducer was rerun on pandas 2.3.3 without crashing.

Full suite passes locally on pandas 3.0.5: 1816 passed, 162 skipped, 8 xfailed. The xfail count is one lower than before because the pandas-floor tripwire went away with the helper it guarded.

---

## Checklist

- [X] Documentation has been updated (if the public API or usage changes).

---


alanprior:
Hello @arthur-priorlabs, assigning this to you as you are a pandas 3 fan :-)

Should be a short review, LMK if the implementation makes sense.

alanprior:
TL;DR: an edge case that fails for pandas < 3 and was fixed for newer versions. I propose the fix, with enough loud signs that will allow us to remove when we upgrade the package.

src/tabpfn/preprocessing/clean.py
```diff
@@ -298,6 +298,51 @@ def _is_single_float_block(X: pd.DataFrame) -> bool:
...
 
 
+#: https://github.com/pandas-dev/pandas/issues/63650, fixed upstream in pandas 3.0:
+#: on pandas < 3, `to_numeric`'s scientific-notation parser has a signed 32-bit
+#: integer overflow for a string whose exponent falls in `[2**31, 2**32)`, e.g.
+#: `"8e2569614270"`, which crashes the interpreter outright rather than raising.
+#: Delete `_to_numeric_or_nan_below_pandas_3` and this flag once the floor in
+#: `pyproject.toml` reaches pandas 3, and `to_numeric_or_nan` with it: at that point
+#: it is exactly `pandas.to_numeric(s, errors="coerce")`.
+_PANDAS_TO_NUMERIC_HAS_OVERFLOW_BUG = Version(pd.__version__) < Version("3.0.0")
```
arthur-priorlabs: I'd shorten the comment (the issue reference is enough + the migration path is obvious), and group this flag with the other similar constants around line 48 so we can easily grep what to migrate in one go.

src/tabpfn/preprocessing/clean.py
```diff
@@ -298,6 +298,51 @@ def _is_single_float_block(X: pd.DataFrame) -> bool:
...
+    return pd.to_numeric(s, errors="coerce")
+
+
+def _to_numeric_or_nan_below_pandas_3(s: pd.Series) -> pd.Series:
+    """The pandas < 3 workaround: parse one value at a time with the built-in `float`.
+
+    The built-in does not share `to_numeric`'s overflow bug (see
+    `_PANDAS_TO_NUMERIC_HAS_OVERFLOW_BUG`), at the cost of no longer being
+    vectorized. TabPFN's own row limits keep a column short enough that this is not
+    a real cost.
```
arthur-priorlabs: Not sure I agree with this comment.

Benchmarking per column on pandas 2.2.3 / python 3.12, 1M-row columns:

|    column kind     |  pandas  | workaround | ratio | added ns/value |
|--------------------|----------|------------|-------|----------------|
| numbers-as-objects |  41.3 ms |   157.7 ms |  3.8x |            116 |
| sci-strings        | 145.3 ms |   262.0 ms |  1.8x |            117 |
| with-missing       | 151.2 ms |   235.1 ms |  1.6x |             84 |
| float-strings      | 142.9 ms |   232.7 ms |  1.6x |             90 |
| mostly-numeric     | 153.9 ms |   240.9 ms |  1.6x |             87 |
| non-numeric        | 369.1 ms |   536.5 ms |  1.5x |            167 |
| int-strings        | 241.3 ms |   227.0 ms |  0.9x |            −14 |

tests/test_preprocessing/test_data_cleaning.py
```diff
@@ -661,6 +664,90 @@ def test__classifier_predict__numpy_array_against_string_fit_categories() -> Non
...
+)
+def test__numeric_workaround__pandas_version__no_longer_needs_the_workaround() -> None:
+    """A forcing function so the pandas < 3 workaround does not outlive its cause.
+
+    `xfail_strict = true` (pyproject.toml) turns an unexpected pass into a hard
+    failure: once the installed pandas actually reaches 3.0, this assertion stops
+    failing, which flips this from a quiet, expected xfail into a build break that
+    says exactly what to remove.
+    """
+    assert Version(pd.__version__) >= Version("3.0.0")
```
arthur-priorlabs: Makes the CI crash on pandas 3.x

src/tabpfn/preprocessing/clean.py
```diff
@@ -298,6 +298,51 @@ def _is_single_float_block(X: pd.DataFrame) -> bool:
...
+    """`pandas.to_numeric(s, errors="coerce")`, without the risk of it segfaulting.
+
+    Args:
+        s: The values to parse. Already-numeric values pass through unchanged.
+
+    Returns:
+        A float64 series the same length as `s`, `NaN` wherever a value is missing
+        or does not parse as a number.
+    """
+    if _PANDAS_TO_NUMERIC_HAS_OVERFLOW_BUG:
```
arthur-priorlabs: Since the conversion is a bit slower, can we do some narrowing on `s` itself?
IUC the crash is only caused by series that contain strings in scientific notation (although checking for strings would probably suffice/be fastest).

arthur-priorlabs (changes requested):
Thanks a lot for the hotfix!
I have a few nitpicks though :)

src/tabpfn/preprocessing/clean.py
alanprior: Good call. I renamed the old flag to `PANDAS_BELOW_3` to be shared, hope it makes sense.

tests/test_preprocessing/test_data_cleaning.py
alanprior: Yeah, thanks, this one would have been caught :) fixed

src/tabpfn/preprocessing/clean.py
alanprior: Amazing catch! Addressing this in the other related comment.

src/tabpfn/preprocessing/clean.py
alanprior: Great call. I tried also your alternative of scientific notations and it was slower. Could be potentially optimized, LMK what you think.

alanprior:
@arthur-priorlabs thanks, great review - not nit at all when comes to runtime. Could you share how did you run the analysis and if this is something we have natively somewhere? Would be good to verify current implementation is OK in that front

arthur-priorlabs:
Nothing native about it; I just had Claude use `torch.utils.benchmark` on `to_numeric_or_nan` under pandas 2.x with different input mixes that register as object dtypes.

As expected, we mostly recover the old performance on object columns without strings (but all cases except `numbers-as-objects` have at least 1 string).

<details><summary>Benchmark on pandas 2.2.3 / python 3.12, 1M-row columns</summary>

|          kind           |        branch         |  pandas  | to_numeric_or_nan | ratio | added ns/value |
|-------------------------|-----------------------|----------|-------------------|-------|----------------|
| numbers-as-objects      | vectorized (floating) |  38.2 ms |           42.5 ms |  1.1x |              4 |
| int-strings             | per-value (string)    | 235.6 ms |          213.4 ms |  0.9x |            −22 |
| non-numeric             | per-value (string)    | 367.8 ms |          523.5 ms |  1.4x |            156 |
| mostly-numeric          | per-value (string)    | 144.7 ms |          231.2 ms |  1.6x |             87 |
| float-strings           | per-value (string)    | 138.4 ms |          231.5 ms |  1.7x |             93 |
| with-missing            | per-value (string)    | 131.3 ms |          226.4 ms |  1.7x |             95 |
| sci-strings             | per-value (string)    | 139.6 ms |          251.1 ms |  1.8x |            112 |
| numbers-with-one-string | per-value (mixed)     |  38.1 ms |          143.5 ms |  3.8x |            105 |
</details>

For future reference, the non-vectorized code path takes ~1-2 additional second per 10M cell column; this is obviously a very acceptable trade-off for not crashing the interpreter (!), but makes an even bigger case for users to switch to pandas 3.x

src/tabpfn/preprocessing/clean.py
```diff
@@ -298,6 +297,62 @@ def _is_single_float_block(X: pd.DataFrame) -> bool:
...
+        s: The values to parse. Already-numeric values pass through unchanged.
+
+    Returns:
+        A float64 series the same length as `s`, `NaN` wherever a value is missing
+        or does not parse as a number.
+    """
+    # https://github.com/pandas-dev/pandas/issues/63650, fixed upstream in pandas 3.0.
+    if PANDAS_BELOW_3 and _may_hold_a_string(s):
+        return _to_numeric_or_nan_below_pandas_3(s)
+    return pd.to_numeric(s, errors="coerce")
```
arthur-priorlabs: Ok, there's a nit and a real bug on this one:
1. nit: we promise a `float64` output in the docstring, and sometimes either of the code paths returns an `int64` output
2. 🐛: the two code paths disagree on the output dtype for a `["1", "2", "3"]` column 🫠

Note that this is only a downstream problem for `_align_columns_to_fitted_dtypes`.

| column | `infer_dtype` | gated to per-value? | `pd.to_numeric` | `to_numeric_or_nan` |
|---------------------------|---------------------|---------------------|---------------|------------------|
| `["1","2","3"]` | string | yes | int64 | float64 |
| `["1","2","n/a"]` | string | yes | float64 | float64 |
| `["1","2",None]` | string | yes | float64 | float64 |
| `["1.5","2.5"]` | string | yes | float64 | float64 |
| `[1, 2, 3]` (object ints) | integer | no | int64 | int64 |
| `[1, 2.5]` (object) | mixed-integer-float | no | float64 | float64 |

arthur-priorlabs (changes requested):
The new changes themselves are great, but I found a bug I hadn't caught before 😖

src/tabpfn/preprocessing/clean.py
alanprior: Thanks :-(

Changed so now to_numeric_or_nan now always returns floats, added also a test.

arthur-priorlabs: Not sure I agree with this, as some `_align_columns_to_fitted_dtypes` outputs that would have been `int64` prior to this PR now become `float64`

alanprior: Right, but from my understanding this doesn't lead to real implications and from a test it seems predictions don't change. Would you rather to just break down those functions instead of entangling them?

alanprior: I agree with you that this is becoming a weird and unexpected rabbithole

arthur-priorlabs: This specific code path is tricky, because it only gets triggered when categorical dtypes differ at train and test (on pandas 3.x).

First, this will cause runtime increases, since we're adding explicit casts to float64.

Second, we can get a collision because two int categories would share a float64 value:
```python
X_train = pd.DataFrame({"id": [9007199254740992, 9007199254740993]})
X_test  = pd.DataFrame({"id": ["9007199254740993"]})
```
The expected output would be encoding `X_test` as category 1 by the ordinal encoder, but it'll get encoded as 0.
We can also get the same float64 collision to turn an unknown category (w.r.t. ordinal encoder) into a known one.

Happy to share a repro script if needed.

alanprior: Thanks - yes, share the script please! :)

I'm trying to reduce the issue as much as possible so we stop entangling things. Ideally, I would like to detect only the cases that are crashing, and for them and only for them apply a fix. It doesn't sound impossible.

arthur-priorlabs:
```python
r"""Where the unconditional float64 cast in `to_numeric_or_nan` changes an encoding.

`to_numeric_or_nan` now ends its vectorized branch with `.astype("float64")`, so that
both of its bodies return the same dtype whatever pandas is installed. `pandas`
otherwise narrows an all-integer column to int64, and that narrowing was load-bearing
in one place: `_align_columns_to_fitted_dtypes`, which coerces a column that was
numeric at fit and arrives as strings at predict, and then hands the result to an
`OrdinalEncoder` holding int64 `categories_`.

Above `2**53` a float64 cannot hold every int64, so the coerced values no longer land
on the categories they were learned from. `numpy` promotes the int64 `categories_` to
float64 to compare against them, which hides the mismatch instead of raising it: the
value silently takes a neighbouring category's code.

Two shapes of that, both printed by this script:

* *collision* -- two fit categories that share one float64. The second one's code is
  reassigned to the first.
* *unknown becomes a match* -- a test value in neither category, which the encoder
  used to report as unknown (`-1`) and now rounds onto a real category.

Both need the same three things, which is what `CASES` below builds:

1. a column declared categorical whose values are integers, so `fix_dtypes` gives it
   `category` dtype with int64 categories and the encoder's `categories_` is int64
   (`fit_kind in "iuf"`);
2. the same column arriving as strings at predict (`values_kind in "OUS"`), which is
   what puts it in `_align_columns_to_fitted_dtypes`'s `to_numeric` list;
3. values above `2**53`, where int64 -> float64 loses exactness.

Only reachable on **pandas >= 3**. Below it, `_may_hold_a_string` sends a string
column to the per-value body, which has always ended in `.astype("float64")` -- so
pandas 2 already returned the post-cast answers and prints `SAME` here. The cast does
not introduce the loss; it propagates pandas 2's loss to pandas 3, which was exact.
Run it under both to see that:

    uv run scripts/repro_to_numeric_float64_cast.py

Cheap and single-threaded -- a handful of rows, no allocation needed.
"""

import warnings
from typing import TYPE_CHECKING, NamedTuple

import numpy as np
import pandas as pd

import tabpfn.preprocessing.clean as clean_module
from tabpfn.preprocessing.clean import fix_dtypes, process_text_na_dataframe
from tabpfn.preprocessing.steps.preprocessing_helpers import get_ordinal_encoder

if TYPE_CHECKING:
    from collections.abc import Callable

#: The column is declared categorical, as `categorical_features_indices=[0]` would.
CAT_INDICES = (0,)


class Case(NamedTuple):
    """One fit/predict pair, and what the cast does to it."""

    name: str
    #: Integer values of the categorical column at fit.
    fit_values: list[int]
    #: The same column at predict, as strings.
    test_values: list[str]


CASES = (
    Case(
        name="collision: two categories share a float64",
        # 2**53 and 2**53 + 1, which are one float64 apart, i.e. not apart at all.
        fit_values=[9007199254740992, 9007199254740993],
        test_values=["9007199254740993"],
    ),
    Case(
        name="unknown becomes a match",
        # Both exactly representable, so the categories survive the cast ...
        fit_values=[9007199254740992, 9007199254740994],
        # ... but this value, in neither of them, rounds onto the first.
        test_values=["9007199254740993"],
    ),
)


def to_numeric_or_nan_before_the_cast(s: pd.Series) -> pd.Series:
    """`to_numeric_or_nan` as it stood before the unconditional float64 cast."""
    if clean_module.PANDAS_BELOW_3 and clean_module._may_hold_a_string(s):
        return clean_module._to_numeric_or_nan_below_pandas_3(s)
    return pd.to_numeric(s, errors="coerce")


def frames(case: Case) -> tuple[pd.DataFrame, pd.DataFrame]:
    """The train and test frames for one case, as a user would pass them to `fit`."""
    n_fit = 2 * len(case.fit_values)
    X_train = pd.DataFrame(
        {"id": case.fit_values * 2, "x": [float(i) for i in range(n_fit)]}
    )
    X_test = pd.DataFrame({"id": case.test_values, "x": [0.5] * len(case.test_values)})
    return X_train, X_test


def encode(
    X_train: pd.DataFrame,
    X_test: pd.DataFrame,
    variant: Callable[[pd.Series], pd.Series],
) -> np.ndarray:
    """Fit an encoder on `X_train`, then transform `X_test` through `variant`.

    These are the two calls `clean_data` and the estimators' predict paths make; the
    coercion under test sits inside the second one.
    """
    original = clean_module.to_numeric_or_nan
    clean_module.to_numeric_or_nan = variant
    try:
        encoder = get_ordinal_encoder()
        process_text_na_dataframe(
            X=fix_dtypes(X=X_train, cat_indices=list(CAT_INDICES)),
            ord_encoder=encoder,
            fit_encoder=True,
        )
        with warnings.catch_warnings():
            # The dtype-drift warning is the point of the fit/predict pair, not news.
            warnings.simplefilter("ignore")
            return process_text_na_dataframe(
                X=fix_dtypes(X=X_test, cat_indices=list(CAT_INDICES)),
                ord_encoder=encoder,
                fit_encoder=False,
            )
    finally:
        clean_module.to_numeric_or_nan = original


def main() -> int:
    """Print both encodings for every case, and whether the cast moved them."""
    print(
        f"pandas {pd.__version__} | PANDAS_BELOW_3 {clean_module.PANDAS_BELOW_3}"
        f" -> the cast is {'invisible' if clean_module.PANDAS_BELOW_3 else 'live'}"
        " at this call site"
    )
    differing = 0
    for case in CASES:
        X_train, X_test = frames(case)
        before = encode(X_train, X_test, to_numeric_or_nan_before_the_cast)
        after = encode(X_train, X_test, clean_module.to_numeric_or_nan)
        same = np.array_equal(before, after, equal_nan=True)
        differing += not same
        print(f"\n{case.name}")
        print(f"  fit categories {case.fit_values}")
        print(f"  test values    {case.test_values}")
        print(f"  before: code {before[:, 0].tolist()}")
        verdict = "SAME" if same else "DIFFERS"
        print(f"  after:  code {after[:, 0].tolist()}   -> {verdict}")
    print(f"\n{differing} of {len(CASES)} case(s) re-encoded by the cast.")
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

arthur-priorlabs: Sorry to be annoying on this PR (and really appreciate your work on it!), but this is the type of issue that could come bite us later :/

alanarazi7: On the contrary, I'm impressed by being thorough and ashamed of my CC slop :)

tests/test_preprocessing/test_numeric_parsing_below_pandas_3.py
```diff
@@ -0,0 +1,168 @@
...
+    (np.nan, True),
+    (pd.NA, True),
+    (pd.NaT, True),
+    # Already a number, so there is no spelling to check.
+    (1, True),
+    (2.5, True),
+    (True, True),
+    (np.int64(7), True),
+    (np.float64(1.5), True),
+    (b"1", True),
```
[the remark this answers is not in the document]
alanarazi7: CC fixed it.

"The bot's specific claim is false, but its probe sits next to a real one: float() accepts any buffer, pandas only accepts bytes."

src/tabpfn/preprocessing/clean.py
alanprior: Got back to the basics, and now it seems to work

alanprior:
@arthur-priorlabs I hope now it's good! I came back to a fix that should be much more conservative. CCs summary:

> **Final design**
> 
> `clean.py` is back at `origin/main`: `to_numeric_or_nan`, the `_align_columns_to_fitted_dtypes` change and the flag rename are all gone. #1202 was only ever modality detection, and sweeping in the second call site is what created the dtype question in the first place.
> 
> The whole fix now lives in `_is_numeric_pandas_series`, which only ever consumed `coerced.notna()`. It answers with a bool and never builds a numeric series:
> 
> ```python
> def _is_numeric_pandas_series(s: pd.Series) -> bool:
>     if pd.api.types.is_numeric_dtype(s.dtype):
>         return True
>     if PANDAS_BELOW_3:
>         return all(_is_numeric_or_missing_for_old_pandas(value) for value in s)
>     coerced = pd.to_numeric(s, errors="coerce")
>     is_numeric_or_missing = coerced.notna() | s.isna()
>     return bool(is_numeric_or_missing.all())
> ```
> 
> On pandas 3 this is byte-identical to `main`. Your 2\*\*53 collision cannot occur, because there is no float64 series left to collide in, and the float64 cast that caused it is gone.
> 
> **Why the version flag stayed**
> 
> The two majors genuinely disagree, so a single hand-rolled parse cannot match both: `"1e400"` reads as NaN on 2.x and as infinity on 3.x, from the same rewritten parser that fixed the segfault. Gating keeps exact pandas semantics on 3.x and confines the stand-in to the version that actually crashes.
> 
> `_is_numeric_or_missing_for_old_pandas` parses one value with the built-in `float`, then rejects the four spellings `float` accepts and pandas does not: the literal `"nan"`, PEP 515 separators (`"1_000"`), non-ASCII digits and spaces (`"٣"`, `"\xa0 5"`), and finite literals that overflow to infinity (`"1e400"`). Checked against `pd.to_numeric` on 2.3.3 over 20,075 values (75 hand-picked plus 20,000 fuzzed): 0 disagreements. Its tests are isolated in `test_numeric_parsing_below_pandas_3.py`, deletable in one go once the floor is 3.0.
> 
> **Runtime, using your methodology**
> 
> `torch.utils.benchmark`, pandas 2.x, 1M-row object columns, your eight kinds. Absolute ms are not comparable to your table (different pandas patch and machine), but the ratio is, since each is normalised against its own `pd.to_numeric` baseline.
> 
> | kind | yours | this branch |
> |---|---|---|
> | numbers-as-objects | 1.1x | **2.9x** |
> | int-strings | 0.9x | 0.7x |
> | non-numeric | 1.4x | **0.0x** |
> | mostly-numeric | 1.6x | **0.0x** |
> | float-strings | 1.7x | 1.2x |
> | with-missing | 1.7x | 1.0x |
> | sci-strings | 1.8x | 1.2x |
> | numbers-with-one-string | 3.8x | **0.9x** |
> 
> Better on 7 of 8, and your worst case drops from 3.8x to 0.9x. Returning a bool is what buys it: `all()` stops at the first non-numeric value, which is index 0 for a text column, i.e. exactly the shape carrying the crashing hash ID.
> 
> The one regression is `numbers-as-objects` (+48 ns/value), from dropping `_may_hold_a_string`. I measured restoring it: it helps that kind (74 to 30 ms) but gives up the free exit on `non-numeric` (0.0 to 3.1 ms) and hurts `numbers-with-one-string` (74 to 83 ms), so I left it out. Worst kind is ~0.5 s per 10M cells, inside the budget you already called acceptable.
> 
> **Two latent bugs in the previous version, now gone**
> 
> - It raised `ValueError` on a non-scalar cell, since `pd.isna([1, 2])` returns an array, where `main` returns NaN cleanly.
> - `float` accepting `"nan"` and `"1_000"` silently moved such columns from TEXT to NUMERICAL.

arthur-priorlabs (approved):
Very well done, thanks a lot!

PR #1204, opened 2026-08-24
Pick the loss criterion from the requested task, not from max_num_classes

bejaeger:
Generated by Claude Code

## Why

`TabPFNRegressor` cannot load a checkpoint whose config also describes a classification head:

```
ValueError: The model loaded, '...', was expected to have a FullSupportBarDistribution
criterion, but instead had a CrossEntropyLoss criterion.
```

`get_loss_criterion` infers the task from `max_num_classes` alone. That works only because every shipped checkpoint is single-task — a regression one has `max_num_classes=0`. A checkpoint carrying both heads has that field set for the classification head, so the inference can only ever answer "classification", whatever the caller asked for.

Architectures that do serve both tasks are loadable today only by accident. `load_model` sniffs `"test_targets_MB" in signature(model.forward).parameters` to decide a checkpoint has no external criterion, and that branch is the only route to building the bar distribution from the model's `regression_borders`. `test_targets_MB` is the training-time loss hook, so the sniff really means "computes its own loss" — it just happens to coincide. `tabpfn_v3.py` carries the parameter as dead weight purely to trip it, with `# TODO: test_targets_MB needed because model_loading has a condition`.

## What

`load_model_criterion_config` already receives the task. This threads it to `load_model` and `get_loss_criterion` and lets it choose the criterion. The parameter is called `estimator_type` throughout, matching `TabPFNClassifier.estimator_type`, `TabPFNRegressor.estimator_type` and `InferenceConfig.get_default(estimator_type=...)`; `load_model_criterion_config`'s pre-existing `which` is renamed to match.

Regression borders now come from wherever the checkpoint keeps them. The two layouts never coexist:

| checkpoint | `criterion.borders` | model `regression_borders` |
|---|---|---|
| v2 / v2.5 / v2.6 regressor | yes | no |
| v3 regressor | no | yes |

So no version switch is needed — `_resolve_regression_borders` takes the criterion state when present and the model buffer otherwise. That retires the dummy borders `get_loss_criterion` invented for the state dict to overwrite (`arange(n+1) * 30000`, with its `NOTE: This just seems to get overriddden` comment) and the `assert config.max_num_classes == 0`.

The `test_targets_MB` sniff goes, and with it the dead parameter in `tabpfn_v3.py` and its TODO. The `criterion is None` fallback in `load_model_criterion_config` goes too, its job having moved into the resolver, so `load_model` no longer returns an optional criterion.

`estimator_type` joins the built-model cache key: the criterion now differs per task for the same file, so without it a checkpoint built for one task could be served for the other. `test_load_model_signature_is_tracked_by_the_cache` exists to force exactly that consideration and caught it.

## Call sites

`load_model` gains a required `estimator_type`, `get_loss_criterion` becomes `(config, *, estimator_type, regression_borders=None)`, and `load_model_criterion_config`'s `which` is renamed to `estimator_type`. The `which` on `download_model` and `resolve_model_path` is left alone — there it selects a HuggingFace repo and filename rather than describing an estimator — as is `initialize_tabpfn_model`'s own. Neither has callers outside this repo that I could find — `tabpfn-extensions` has none. One in fomo-fitting does need the extra argument: `experimental/anurag/per_dataset_finetuning/finetune_slice.py` calls `load_model(path=..., cache_trainset_representation=False)` for its inference config.

Also, pointing `TabPFNClassifier` at a regression checkpoint (`max_num_classes=0`) now raises with a clear message instead of silently building a bar distribution it discards. One existing test relied on that mismatch being ignored and is corrected to `which="regressor"`.

<details><summary>Verification</summary>

Every cached checkpoint version x both tasks x both fit modes — 20 combinations, all passing. v2, v2.5, v2.6 and v3 predictions are byte-identical to before the change.

```
### fit_mode=fit_preprocessors ###        ### fit_mode=fit_with_cache ###
  v2   clf  OK                             v2   clf  OK
  v2   reg  OK                             v2   reg  OK
  v2.5 clf  OK                             v2.5 clf  OK
  v2.5 reg  OK                             v2.5 reg  OK
  v2.6 clf  OK                             v2.6 clf  OK
  v2.6 reg  OK                             v2.6 reg  OK
  v3   clf  OK                             v3   clf  OK
  v3   reg  OK                             v3   reg  OK
  v3.5 clf  OK                             v3.5 clf  OK
  v3.5 reg  OK  <- was ValueError          v3.5 reg  OK  <- was ValueError
```

`380 passed, 79 skipped` across `test_model_loading`, `test_model_cache`, `test_inference`, `test_checkpoint`, `test_architectures`, `test_classifier_interface` and `test_regression_metrics`. Repo-wide ruff clean.

New tests: `test__load_multitask_ckpt__criterion_follows_the_requested_task` (one file backs both estimators) and `test__load_classification_only_ckpt__as_regressor__raises`.

The both-heads checkpoint used for the end-to-end runs is the v3.5 pre-release, loaded through the `tabpfn_v3_5_prerelease` architecture registered from fomo-fitting.
</details>


src/tabpfn/model_loading.py
```diff
@@ -697,18 +696,9 @@ def load_model_criterion_config(  # noqa: PLR0912
 
         loaded_model, criterion, architecture_config, inference_config = load_model(
             path=path,
+            which=which,
```
jmkuebler: `which` doesn't seem very descriptive to me. Could we call it task_type or similar? Or was it a very deliberate choice to call it `which`

jmkuebler (dismissed):
LGTM except for the naming choice. Maybe I am missing sth there.

priorphil (dismissed):
Thanks!

PR #1205, opened 2026-08-26
Recognize date-like strings and decouple text-vs-category detection

alanprior:
## What & why

First slice of a stack: `detect_feature_modalities` now recognizes date-like
strings and decides category-vs-text independently of category-vs-number.
Detection only -- nothing acts on a date yet, so it's still read as a plain
category or text.

## How

- `FeatureModality.DATE`: reserved, not yet produced by anything downstream.
- `_classify_string_like_column` returns `DATE` when detected; a temporary
  `_demote_dates_for_now` (to be deleted once real date expansion lands)
  demotes it to `CATEGORICAL`/`TEXT` right after warning about it by name.
- `InferenceConfig.MIN_CARDINALITY_FOR_TEXT`: new field for category-vs-text,
  independent of `MAX_UNIQUE_FOR_CATEGORICAL_FEATURES`.
- Early-exit fix: the unique-count shortcut now accounts for
  `min_cardinality_for_text` too, not just the other two thresholds.

## Decisions

- `MIN_CARDINALITY_FOR_TEXT` defaults equal to `MAX_UNIQUE_FOR_CATEGORICAL_FEATURES`,
  so this PR changes no default behavior (verified bit-for-bit against `origin/main`).
- No `USE_DATES`/`USE_TEXT` flags yet -- nothing would consume them.
- A native `datetime64` column still crashes (pre-existing, unrelated to this
  PR). A `stringify_datetime_columns` fix was tried and dropped since the
  string round-trip is lossy; a test now documents the crash as known and
  deliberately unfixed here.

## Follow-ups

- Real date/text expansion (`DatetimeEncoder`/`StringEncoder`),
  `USE_DATES`/`USE_TEXT` flags, and raising `MIN_CARDINALITY_FOR_TEXT`'s default.
- Fix the native `datetime64` crash without a lossy round-trip.


src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -80,42 +87,129 @@ def detect_feature_modalities(
...
+    calendar features, so a `DATE` result isn't consumable by anything
+    downstream. This function only exists to make it consumable for this PR --
+    demoting it to whichever of `CATEGORICAL`/`TEXT` a same-shaped non-date
+    string would get, reusing `_detect_numeric_as_categorical` for the
+    low-cardinality case, exactly like a low-cardinality number already gets.
+    A follow-up PR that adds real date expansion should delete this function
+    outright rather than extend it.
+
+    Called after `_warn_about_text_or_dates`, which reports a `DATE` column by
+    name while the schema still says so.
```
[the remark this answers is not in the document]
alanprior: Whatever.

src/tabpfn/preprocessing/clean.py
```diff
@@ -182,6 +182,33 @@ def coerce_nullable_dtypes_to_numpy(X: pd.DataFrame) -> pd.DataFrame:
...
+    cols = [
+        col
+        for col, dtype in X.dtypes.items()
+        if pd.api.types.is_datetime64_any_dtype(dtype)
+    ]
+    if not cols:
+        return X
+    X = X.copy()
+    for col in cols:
+        X[col] = X[col].dt.strftime("%Y-%m-%d %H:%M:%S")
```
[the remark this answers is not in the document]
alanprior: Irrelevant by now

alanprior:
@bejaeger Hi there!

This PR is the first one, expanding our detection capabilities of dates, alongside addingwarnings for dates as well (@vahidblm ).

This PR still decouples action taking of a next PRs which will add both text and date treatments.

src/tabpfn/preprocessing/datamodel.py
```diff
@@ -94,12 +94,16 @@ class FeatureModality(str, Enum):
     instance, a numerical dtype could represent numerical features
     or categorical features, while a string could represent categorical
     or text features.
+
+    ``DATE`` is reserved and not yet produced by anything; a recognized date is
+    always demoted to `CATEGORICAL`/`TEXT` for now.
```
bejaeger: nit: could also remove or make a TODO comment to remember to clean it up later

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -242,6 +318,28 @@ def _is_numeric_or_missing_for_old_pandas(value: object) -> bool:
...
+    non_null = s.dropna()
+    if non_null.empty:
+        return False
+    try:
+        with warnings.catch_warnings():
+            # A mixed-format column makes `to_datetime` fall back to parsing one
+            # value at a time and warn about it; this is only a probe, so the
+            # warning would be noise for the caller regardless of the outcome.
+            warnings.simplefilter("ignore")
+            parsed = pd.to_datetime(non_null, errors="coerce")
```
bejaeger, line 336: We've seen before that something like this could get a performance bottleneck for very large n_rows. Would be good to double-check.

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -242,6 +318,28 @@ def _is_numeric_or_missing_for_old_pandas(value: object) -> bool:
...
+    except (TypeError, ValueError):
+        return False
+    return bool(parsed.notna().all())
```
bejaeger, line 339: I'm wondering in general, do we need these heuristics just yet, or should we start with supporting the case where the user passes a `datetime` column?

bejaeger (dismissed):
Thanks, just some small comments. Otherwise LGTM!

src/tabpfn/preprocessing/modality_detection.py
alanprior: @bejaeger I think that it is very often the case that dates are easily parseable but given as strings. We see it in benchmarks, and it's very common to see as well in databases etc, because it's easier to store as strings.

src/tabpfn/preprocessing/modality_detection.py
alanprior: Good comment, thanks. It seems that we only get here part of the time:

> Design note in our favour: modality_detection.py:28 already defines _EARLY_EXIT_PREFIX_ROWS = 1024 and uses it at line 231 for exactly this shape — compute on the prefix, fall through to the full column only if the prefix answer isn't decisive.

However, this led me think - we might want to parallelize detection?

Anyway, I feel that this goes beyond the scope of this PR, but I will open an internal ticket for this, so we can think about going deeper / optimizing. Please LMK if you disagree.

PR #1206, opened 2026-08-26
Expand a detected date into calendar features behind USE_DATES

alanprior:
## What & why

Second slice of the stack, on top of PriorLabs/TabPFN#1205 (detection): a detected date can<br>now actually be expanded into calendar features via `skrub.DatetimeEncoder`,<br>behind a new `InferenceConfig.USE_DATES` flag that defaults to `False`. Off<br>(default), behavior is unchanged from PriorLabs/TabPFN#1205 -- a date is silently demoted and<br>warned about by name.og

## How

* `InferenceConfig.USE_DATES`: new flag, default `False`.
* `detect_feature_modalities` gains `use_dates`, threaded to a reworked<br>`_resolve_dates` (was `_demote_dates_for_now`): still detection-only in<br>contract (`-> FeatureSchema`, never touches `X`) -- it decides whether to<br>leave a column tagged `DATE` for the caller to expand, or demote it.
* New `date_encoding.py`: `make_datetime_encoder()` (lazy `skrub` import) and<br>`expand_date_features()`, called unconditionally from `clean_data` (a no-op<br>whenever there's no `DATE`-tagged column, always true when `USE_DATES` is<br>off). Fits one `DatetimeEncoder` per date column, removes the raw column,<br>appends the encoded output as `NUMERICAL` columns.
* `date_encoders_` (mirrors `ordinal_encoder_`'s lifecycle): fitted encoders<br>keyed by original column index, reapplied positionally at predict time, on<br>both the direct and ensemble-worker paths in `classifier.py`/`regressor.py`.
* `pyproject.toml`: added `skrub>=0.10,<0.11` as a hard dependency.

## Decisions

* No new "IRRELEVANT" modality: the raw column is fully removed<br>(`FeatureSchema.remove_columns`) and the encoder's output appended<br>(`append_columns`) -- the same pattern `remove_constant_features_step.py`<br>already uses to drop a column outright.
* A low-cardinality date is always demoted to `CATEGORICAL` regardless of<br>`USE_DATES`, via the same threshold a low-cardinality *number* already gets<br>-- a handful of repeating dates is a category, not worth expanding.
* The warn-vs-demote call order flipped from PriorLabs/TabPFN#1205 (was warn-then-demote):<br>the warning now needs to know whether a date was silently demoted<br>(`USE_DATES` off, warn) or genuinely left for expansion (no warning), not<br>just whether it was detected as a date.
* Fixed a real, verified bug found while building this: `pd.to_datetime`<br>infers a format from an early value and silently coerces a later,<br>differently-shaped but valid value to NaT with no warning (e.g. mixing<br>`"2020-01-01"` and `"2020-06-15 13:45:30"`). `format="mixed"` fixes this in<br>both date detection and expansion.

## Follow-ups

* `USE_TEXT` / `StringEncoder` -- separate slice, not touched here.
* The numeric-string-as-text ordinal-encoding bug from the original<br>exploration -- separate, unrelated slice.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


alanprior:
@bejaeger lots of changes given our last round. I will polish a bit, and try to find more bugs.

bejaeger:
superseded by PriorLabs/TabPFN#1236

PR #1207, opened 2026-08-26
Expand a detected date into calendar features behind USE_DATES

alanprior:
## What & why

Second slice of the stack, on top of #1205 (detection): a detected date can
now actually be expanded into calendar features via `skrub.DatetimeEncoder`,
behind a new `InferenceConfig.USE_DATES` flag that defaults to `False`. Off
(default), behavior is unchanged from #1205 -- a date is silently demoted and
warned about by name.

## How

- `InferenceConfig.USE_DATES`: new flag, default `False`.
- `detect_feature_modalities` gains `use_dates`, threaded to a reworked
  `_resolve_dates` (was `_demote_dates_for_now`): still detection-only in
  contract (`-> FeatureSchema`, never touches `X`) -- it decides whether to
  leave a column tagged `DATE` for the caller to expand, or demote it.
- New `date_encoding.py`: `make_datetime_encoder()` (lazy `skrub` import) and
  `expand_date_features()`, called unconditionally from `clean_data` (a no-op
  whenever there's no `DATE`-tagged column, always true when `USE_DATES` is
  off). Fits one `DatetimeEncoder` per date column, removes the raw column,
  appends the encoded output as `NUMERICAL` columns.
- `date_encoders_` (mirrors `ordinal_encoder_`'s lifecycle): fitted encoders
  keyed by original column index, reapplied positionally at predict time, on
  both the direct and ensemble-worker paths in `classifier.py`/`regressor.py`.
- `pyproject.toml`: added `skrub>=0.10,<0.11` as a hard dependency.

## Decisions

- No new "IRRELEVANT" modality: the raw column is fully removed
  (`FeatureSchema.remove_columns`) and the encoder's output appended
  (`append_columns`) -- the same pattern `remove_constant_features_step.py`
  already uses to drop a column outright.
- A low-cardinality date is always demoted to `CATEGORICAL` regardless of
  `USE_DATES`, via the same threshold a low-cardinality *number* already gets
  -- a handful of repeating dates is a category, not worth expanding.
- The warn-vs-demote call order flipped from #1205 (was warn-then-demote):
  the warning now needs to know whether a date was silently demoted
  (`USE_DATES` off, warn) or genuinely left for expansion (no warning), not
  just whether it was detected as a date.
- Fixed a real, verified bug found while building this: `pd.to_datetime`
  infers a format from an early value and silently coerces a later,
  differently-shaped but valid value to NaT with no warning (e.g. mixing
  `"2020-01-01"` and `"2020-06-15 13:45:30"`). `format="mixed"` fixes this in
  both date detection and expansion.

## Follow-ups

- `USE_TEXT` / `StringEncoder` -- separate slice, not touched here.
- The numeric-string-as-text ordinal-encoding bug from the original
  exploration -- separate, unrelated slice.

### DIFF
--- a/changelog/1207.added.md
+++ b/changelog/1207.added.md
@@ -1,0 +1,1 @@
+`fit()` can now expand a detected date column into calendar features (year, month, day, weekday, ...) via `skrub.DatetimeEncoder`, behind a new `InferenceConfig.USE_DATES` flag that defaults to `False`. With the flag off, behavior is unchanged: a date is silently read as a plain category or text and `fit()` warns about it by name.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,8 @@ dependencies = [
   "tqdm>=4.66.0",
   "filelock>=3.11.0",
   "lightgbm>=4.4",
+  # Capped: skrub is pre-1.0 and its minor releases change transformer behavior.
+  "skrub>=0.10,<0.11",
   "mlx>=0.29.3; sys_platform == 'darwin' and platform_machine == 'arm64'",
 ]
 requires-python = ">=3.10"
@@ -339,6 +341,7 @@ module = [
   "torch.*",
   "kditransform.*",
   "lightgbm.*",
+  "skrub.*",
 ]
 ignore_missing_imports = true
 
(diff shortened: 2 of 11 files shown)


alanprior:
Duplicate of #1206 (same branch, this repo's automation auto-opened #1206 against `main` when the branch was pushed). Continuing there.

PR #1211, opened 2026-08-27
Read declared and category columns as categorical, and label unexpanded text NUMERICAL, each behind a flag

alanprior:
## What & why

The detection half of the text work, split from `TRANSFORM_TEXT` (#1240) at review. Three changes to how a string column is labelled, each behind its own `inference_config` flag so their effect can be benchmarked apart. All three default to the new behaviour.

- **`CATEGORY_DTYPE_IS_CATEGORICAL`**: a column with pandas' `category` dtype is a category at any cardinality, like one listed in `categorical_features_indices`, and the tuning estimators get it as such. Validation flattens the dtype away before detection runs, so the positions are read off the frame first, which is what #1236's reverted attempt missed.
- **`DECLARED_STRINGS_ARE_CATEGORICAL`**: a string column listed in `categorical_features_indices` is a category at any cardinality. Above `MIN_CARDINALITY_FOR_TEXT` it used to become text silently, since declaring it also suppressed the free-text warning, so it reached the model neither cast to category nor flagged categorical.
- **`TEXT_AS_NUMERICAL`**: a text column nothing expanded is labelled `NUMERICAL`, which is how the model always read it: ordinal-encoded by the alphabetical rank of each string. Only the free-text warning ever read `FeatureModality.TEXT`; every other consumer keys on `CATEGORICAL` or `NUMERICAL`, so the `TEXT` label mostly meant "skipped by the numeric GPU steps". The enum member stays, and is assigned only with the flag off.

The free-text warning moves from detection into `TextTransformer`, which sees the raw frame before validation and before any expansion, so it also names `object` and undeclared `category` columns and states the remedy that applies.

## How

- `base.expand_dates_and_text` joins the `category` positions to the declared indices right after the date expansion, while the frame still has dtypes; `TextTransformer` then treats them as declared and `categorical_features_indices_` carries them to the tuning estimators.
- `detect_feature_modalities` takes the two labelling flags as keyword arguments; the estimators pass them from the config.
- `TextTransformer._fit` collects every text candidate (`string`, pyarrow string, `object`, `category` dtype above the cutoff, not declared, not all numbers), expands the string ones when `TRANSFORM_TEXT` is on, and warns about the rest.

## Not covered

- An ordered `category` (`ordered=True`) is read as a plain category; its order is not used. Separate PR.
- Inferred categoricals are still not handed to the tuning estimators, only declared ones (now including `category` dtype). Pre-existing.


alanprior:
Both Bugbot findings fixed in d16752dc:

- **Demoted dates get text expansion**: `_demote_dates` now forces `CATEGORICAL` unconditionally when `use_text` is on, so a high-cardinality date demoted because `USE_DATES` is off can never land on `TEXT` and get swept into `StringEncoder`.
- **Declared categoricals still text-encoded**: `expand_text_features` now takes `provided_categorical_indices` and excludes them from its candidates, since detection has never honored that declaration for strings (a pre-existing gap that only became consequential once `TEXT` got a real encoder).

Added regression tests for both at the unit level (`test_modality_detection.py`, `test_string_encoding.py`) and end-to-end through `TabPFNClassifier.fit`.

alanprior:
@bejaeger a lot of debatable decisions. feel free to roast this :)

src/tabpfn/classifier.py
```diff
@@ -844,9 +851,9 @@ def _get_tuning_classifier(self, **overwrite_kwargs: Any) -> TabPFNClassifier:
             "differentiable_input": False,
             "tuning_config": None,  # never tune inside tuning
             # Fit on the already-expanded array, where a declared column may
-            # have moved down past an expanded date.
-            "categorical_features_indices": self.date_transformer_.output_indices(
-                self.categorical_features_indices
+            # have moved down past an expanded date or text column.
+            "categorical_features_indices": self.text_transformer_.output_indices(
+                self.date_transformer_.output_indices(self.categorical_features_indices)
```
bejaeger: What about introducing a `categorical_features_indices_`that we set in `_initialize_dataset_preprocessing` so we can re-use it here?

src/tabpfn/classifier.py
```diff
@@ -768,10 +776,8 @@ def _initialize_dataset_preprocessing(
 
         feature_schema = detect_feature_modalities(
             X=X,
-            feature_names=date_transformer.feature_names_out_,
-            provided_categorical_indices=date_transformer.output_indices(
-                self.categorical_features_indices
-            ),
+            feature_names=text_transformer.feature_names_out_,
```
bejaeger: Can we also return `feature_names_out_`from `expand_dates_and_text` so we don't rely on the order of expanding dates and text here?

src/tabpfn/inference_config.py
```diff
@@ -151,6 +146,19 @@ class InferenceConfig:
...
 
+    TRANSFORM_TEXT: bool = False
+    """Whether a text column, a `string` column with more than
+    `MIN_CARDINALITY_FOR_TEXT` distinct values, is expanded into
+    `TEXT_N_COMPONENTS` numeric features via `skrub.StringEncoder` (tf-idf over
+    character n-grams, truncated SVD). Off, the model reads the alphabetical rank
+    of each string and `fit` warns about it. An `object` column is never
+    expanded. Not run by the fine-tuning estimators."""
+
+    TEXT_N_COMPONENTS: int = 30
```
bejaeger, line 175: nit: why not 32?

src/tabpfn/regressor.py
```diff
@@ -995,9 +1002,9 @@ def _get_tuning_regressor(self, **overwrite_kwargs: Any) -> TabPFNRegressor:
             "differentiable_input": False,
             "tuning_config": None,  # never tune inside tuning
             # Fit on the already-expanded array, where a declared column may
-            # have moved down past an expanded date.
-            "categorical_features_indices": self.date_transformer_.output_indices(
-                self.categorical_features_indices
+            # have moved down past an expanded date or text column.
+            "categorical_features_indices": self.text_transformer_.output_indices(
+                self.date_transformer_.output_indices(self.categorical_features_indices)
```
bejaeger: see above in classifier

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -79,59 +77,22 @@ def detect_feature_modalities(
...
-        # stacklevel=6 reaches the `estimator.fit(X, y)` call site; pinned by the
-        # `warning.filename` asserts in the tests.
-        stacklevel=6,
-    )
+    declared = set(indices or ())
+    if isinstance(X, pd.DataFrame):
+        for i, dtype in enumerate(X.dtypes):
+            if isinstance(dtype, pd.CategoricalDtype):
+                declared.add(i)
+    return sorted(declared)
```
bejaeger: I noticed one subtlety here which is that a Categorical can have "ordered=True". In that case, our current CATEGORICAL handling would be flawed cause we shuffle the categories for estimators.<br><br>I would suggest to deal with this in a separate PR, as it can be well separated from the text transform introduction.

bejaeger (changes requested):
Looks great! I left some small comments.<br><br>Also, one ask:
Let's decouple the "transform text" change from the change to NOT convert categoricals to text. Then we can have a more careful look at the latter.

src/tabpfn/preprocessing/text.py
```diff
@@ -0,0 +1,376 @@
...
+
+        text = _text_positions(
+            X,
+            declared=self._declared_categorical,
+            min_cardinality=self._min_cardinality_for_text,
+        )
+        positions = [
+            i
+            for i in text
+            if self._transform_text and isinstance(X.dtypes.iloc[i], pd.StringDtype)
```
bejaeger: isn't there more string types?

src/tabpfn/inference_config.py
alanprior: This is the standard being used as default for text-tabular literature and implementations (STRABLE paper, skrub's default).

alanprior:
We've covered this in other PRs - closing

PR #1212, opened 2026-08-27
Update example notebook for client v0.5.0 release

ggprior:
Notebook update for client v0.5.0 release (TL;DR now requires TABPFN_TOKEN env variable passed explicitly


oscarkey (commented):
lgtm! But seems like the outputs have disappeared from most of the cells, would you re-run it so people can see what it does even if they don't run it?

ggprior:
ah good catch! btw wdyt on adding CI for making sure the notebook keeps working?

oscarkey (approved):
lgtm thanks!

PR #1215, opened 2026-08-27
Fix FullSupportBarDistribution tail sampling and quantiles

### LINKED ISSUE #1200: FullSupportBarDistribution.sample()/cdf()/icdf() disagree with forward() in outer buckets
# FullSupportBarDistribution.sample()/cdf()/icdf() disagree with forward() in outer buckets

## Summary

`FullSupportBarDistribution.forward()` evaluates the first and last components as
half-normal tails with support on all of `R`, but `cdf()`, `icdf()`, and `sample()`
inherit the finite, within-bin-uniform implementation from `BarDistribution`.

Consequently, values returned by `sample()` are not distributed according to the
density returned by `-forward()`. The same inconsistency affects tail quantiles,
`median()`, and methods such as `ucb()` that call the inherited `icdf()`.

This is still present in PriorLabs/TabPFN commit
`930ea728b40f7f321d965250838bca8507bd70b9` (2026-08-20):

- `FullSupportBarDistribution.forward()` adds the half-normal tail density.
- `FullSupportBarDistribution.sample()` calls `self.icdf(...)`.
- `icdf()` is inherited from `BarDistribution` and linearly interpolates between
  the two finite borders of every selected bucket, including both outer buckets.

## Minimal reproduction

```python
import torch

try:
    from tabpfn.architectures.shared.bar_distribution import FullSupportBarDistribution
except ImportError:
    from tabpfn.model.bar_distribution import FullSupportBarDistribution

torch.manual_seed(7)
dist = FullSupportBarDistribution(torch.tensor([-2.0, -1.0, 1.0, 2.0]))
probs = torch.tensor([0.25, 0.50, 0.25])
logits = probs.log().repeat(20_000, 1)

samples = dist.sample(logits)
buckets = dist.map_to_bucket_idx(samples).clamp(0, dist.num_bars - 1)
reported_logq = -dist(logits, samples)
sample_logq = probs[buckets].log() - dist.bucket_widths[buckets].log()
edge = (buckets == 0) | (buckets == dist.num_bars - 1)

print(samples.min().item(), samples.max().item())
print(((samples < -2.0) | (samples > 2.0)).sum().item())
print((reported_logq[edge] - sample_logq[edge]).min().item())
print((reported_logq[edge] - sample_logq[edge]).max().item())
```

Observed:

```text
sample range: approximately [-2, 2]
outside [-2, 2]: 0 / 20,000
reported_logq - sample_logq in edge buckets: [-0.847, -0.620]
```

The documented full-support construction puts half of each outer component's
mass beyond its finite outer border. With probabilities `[0.25, 0.50, 0.25]`,
the expected outside fraction is therefore

```text
0.5 * (0.25 + 0.25) = 0.25,
```

or about 5,000 of 20,000 samples, rather than zero.

The complete reproducible audit also compares empirical and analytic CDFs. It
obtains KS distance `0.00585` to the inherited finite ICDF distribution versus
`0.125` to the density implemented by `forward()`.

## Expected behavior

For borders `b[0] < ... < b[K]` and component probabilities `pi`, the two outer
components should be sampled from the same half-normal densities used by
`forward()`:

```text
left:  Y = b[1]   - HalfNormal(scale_left)
right: Y = b[K-1] + HalfNormal(scale_right)
```

where each scale keeps half of the component mass inside its nominal outer bin:

```text
scale_left  = (b[1] - b[0])     / HalfNormal(1).icdf(0.5)
scale_right = (b[K] - b[K - 1]) / HalfNormal(1).icdf(0.5)
```

For a uniform variate `u` that selects the left component, with conditional
rank `r = u / pi[0]`, the inverse CDF should be

```text
b[1] - HalfNormal(scale_left).icdf(1 - r).
```

For the right component it should be

```text
b[K-1] + HalfNormal(scale_right).icdf(r).
```

Interior buckets should retain their current within-bin-uniform interpolation.

## Requested fix

1. Override `cdf()` and `icdf()` in `FullSupportBarDistribution` with the
   half-normal formulas already used by `forward()`.
2. Make `sample()` use that corrected inverse CDF, or sample the selected outer
   components directly from their half-normal distributions.
3. Preserve batch shape, device, and dtype; avoid rebuilding the result through
   `torch.tensor(list_of_tensors)`, which can move CUDA results back to CPU.
4. Keep the existing borders, component probabilities, and half-normal scale
   convention so checkpoint likelihoods remain unchanged.
5. Audit inherited methods that depend on `cdf()`/`icdf()`, including
   `quantile()`, `median()`, `ucb()`, and border-probability translation.
6. State the affected releases and include the correction in the release notes.

## Proposed regression tests

1. **Deterministic inverse-CDF checkpoints.** For borders `[-2,-1,1,2]` and
   probabilities `[0.25,0.50,0.25]`, verify
   `icdf(0.125) == -2` and `icdf(0.875) == 2` within numerical tolerance.
   The current inherited implementation returns `-1.5` and `1.5`.
2. **CDF/ICDF round trip.** Check `cdf(logits, icdf(logits, u)) == u` for ranks
   inside both tails and all interior buckets.
3. **Sampling tail mass.** A seeded large sample should place approximately 25%
   of samples outside `[-2,2]` in the reproduction above.
4. **Density/sampling consistency.** A goodness-of-fit test should accept the
   distribution defined by `forward()` and reject the finite inherited-ICDF law.
5. **Device/dtype coverage.** Test CPU and CUDA when available, including batched
   logits and float32/float64.

## Why this matters

The mismatch invalidates any use that assumes samples and density values refer
to the same distribution, including importance sampling, Monte Carlo
expectations, density estimation, generative sampling, and calibrated tail
quantiles. For importance sampling it can additionally violate the proposal
support condition because the implemented sampler has finite support while the
reported proposal density is full-support.

ppguo:
## Summary

- make `FullSupportBarDistribution.cdf()` and `icdf()` use the half-normal outer components already used by `forward()`
- sample the outer components from their half-normal tails and keep uniform sampling for interior buckets
- preserve batch shape, device, dtype, and temperature behavior with vectorized sampling
- leave bounded `BarDistribution` unchanged

Fixes #1200

## Rationale

The original PFN paper defines both bounded and infinite-support Riemann distributions ([arXiv:2112.10510](https://arxiv.org/abs/2112.10510)). `FullSupportBarDistribution` is the infinite-support form used by the regression checkpoints: its existing `forward()` already evaluates the two outer components as half-normal tails, but inherited `cdf()`, `icdf()`, and the old `sample()` kept all mass within the finite borders.

For a component-conditional probability `r`, this change uses `b_1 - H_L^{-1}(1-r)` in the left tail and `b_{K-1} + H_R^{-1}(r)` in the right tail. Interior buckets retain linear interpolation.

With borders `[-2, -1, 1, 2]` and component probabilities `[0.25, 0.50, 0.25]`:

- before: 0/20,000 samples were outside `[-2, 2]`, although the reported density assigns 25% mass there
- after: 4,934/20,000 samples are outside (`0.2467`); empirical-CDF KS distance is `0.00716` to the documented full-support law versus `0.1266` to the old finite law
- analytic checks now give `icdf(0.125) == -2` and `icdf(0.875) == 2`

## Public behavior

This is a public behavior correction: `sample()`, `cdf()`, quantiles, median/UCB, and the class's border translation now agree with the existing full-support log density. Samples from the outer components can now lie beyond the finite borders.

There are no function-signature, checkpoint-format, state-dict, border, component-probability, or `forward()`/log-density changes.

## Tests

Validated in an isolated Linux CPU/CUDA environment with Python 3.10.18, PyTorch 2.6.0+cu124, pytest 8.4.2, and Ruff 0.15.12:

- `pytest tests/test_architectures/test_bar_distribution.py` with CUDA excluded: 19 passed
- `pytest tests/test_architectures/test_bar_distribution.py` with CPU excluded on CUDA: 19 passed
- `ruff check src/tabpfn/architectures/shared/bar_distribution.py tests/test_architectures/test_bar_distribution.py`: passed
- `ruff format --check src/tabpfn/architectures/shared/bar_distribution.py tests/test_architectures/test_bar_distribution.py`: passed
- independent 20,000-sample tail-bucket audit: outside mass `0.2467`, full-support KS `0.00716`

`tests/test_architectures/test_bar_distribution.py` is the repository's existing bar-distribution test file; this PR adds the full-support regression cases there. They cover analytic tail checkpoints, CDF/ICDF round trips including probabilities 0 and 1, float32 cumulative-probability roundoff, interior buckets and borders, inherited quantile helpers and border translation, 1D and multidimensional batches, float32/float64, CPU/CUDA, and empirical tail mass/CDF. The repository's full CI suite is separate from these targeted runs.

## Follow-up audit

The default regressor also has a standalone finite `_cdf()` path in `utils.translate_probs_across_borders()`, and target preprocessing can use the nonlinear `safepower` transform. Those paths are intentionally unchanged here so this fix stays scoped to `FullSupportBarDistribution`; maintainers can decide whether they should be addressed in a follow-up PR.


src/tabpfn/architectures/shared/bar_distribution.py
```diff
@@ -501,15 +501,88 @@ def assert_support(self, *, allow_zero_bucket_left: bool = False) -> None:
...
+        side_normals = (
+            self.halfnormal_with_p_weight_before(self.bucket_widths[0]),
+            self.halfnormal_with_p_weight_before(self.bucket_widths[-1]),
+        )
+        left_tail_values = self.borders[1] - side_normals[0].icdf(
+            1.0 - conditional_prob,
+        )
+        right_tail_values = self.borders[-2] + side_normals[1].icdf(
+            conditional_prob,
+        )
```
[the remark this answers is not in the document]
ppguo: Confirmed, thank you. With float32 logits I reproduced cumsum[-1] = 0.99999994 and a component-conditional probability of 1.00000048, which made icdf(1.0) return NaN. Commit e8312b2 clamps the component-conditional probability to [0, 1] and adds a regression test for this endpoint. The targeted test file now passes 19 tests on both CPU and CUDA; Ruff check/format and the independent 20,000-sample audit also pass.

oscarkey:
hey @eliott-kalfon, I'm happy to take the review on this one, as I was assigned the original ticket, unless you would like to!

eliott-kalfon:
> hey @eliott-kalfon, I'm happy to take the review on this one, as I was assigned the original ticket, unless you would like to!

Hey Oscar, that would nice thanks. Happy to double check after you too :)

oscarkey:
hey @ppguo , sorry it's taking so long for me to get to this PR, it has been a busy couple of weeks. But I will review it as soon as possible.

ppguo:
> hey @ppguo , sorry it's taking so long for me to get to this PR, it has been a busy couple of weeks. But I will review it as soon as possible.

Thanks for the update! No worries at all — I really appreciate you taking the time to review it.

ppguo:
Hi @oscarkey, I’ve synced the PR with the latest `main`; when you have a moment, could you please approve the updated CI workflows so they can run? Thanks!

ppguo:
Quick CI follow-up: the two macOS compatibility jobs completed successfully, including the PR-targeted tests (Python 3.10 and 3.14). The remaining six failed jobs did not reach the test step; they stopped while downloading models with `TabPFNLicenseError`, and the logs show empty `TABPFN_TOKEN` / `HF_TOKEN` values.

Could a maintainer please rerun those jobs in a trusted context with the required credentials, or advise how external-fork PRs should handle this model-download step? Thanks!

oscarkey:
hey, thanks for your patience on this!

@tobias-priorlabs @LeoGrin I was wondering if I could ask your opinion on this change:

This PR updates the `.cdf()`, `.icdf()`, and `.sample()` functions of the `FullSupportBarDistribution` to have inifinite support, adding half-normal tails on each side of the bar distribution. This makes them consistent with `FullSupportBarDistribution.forward()`, which is important for downstream tasks like importance sampling (see the issue https://github.com/PriorLabs/TabPFN/issues/1200).

Updating `.cdf()` and `.icdf()` impacts:
- `TabPFNRegressor.predict()`: if the user asks for the median or any other quantiles, and the quantile falls in the first or last bar.
- `FullSupportBarDistribution.quantile()`, `.median()`, `.ucb()`, 

It does not impact the mean (the default return value of `TabPFNRegressor.predict()`), as this is already computed taking the tails into account.

To assess the impact, I got Claude to TabPFN through ScoringBench, once with the finite bar distribution and once with the half-normal tails (ScoringBench actually uses the bar distribution directly, rather than our functions, so I got Claude to modify ScoringBench to do this). Below is a summary of the results.  "max |%|" is the largest absolute per-split change and "mean %" the signed mean of the per-split percentage changes.
```
                          v2.6 max |%| v2.6 mean % v3 max |%| v3 mean % v3.5 max |%| v3.5 mean %
metric
crps                               0.0        +0.0        0.0      +0.0          0.0        +0.0
interval_score_20                  0.0        -0.0        0.0      +0.0          0.0        +0.0
interval_score_40                  0.0        -0.0        0.0      -0.0          0.0        -0.0
interval_score_60                  0.0        +0.0        0.0      -0.0          0.0        +0.0
interval_score_80                  0.0        -0.0        0.0      +0.0          0.0        +0.0
interval_score_90                  0.0        -0.0        0.0      +0.0          0.0        +0.0
interval_score_95                  0.0        -0.0        0.0      +0.0          0.0        -0.0
crts_alpha_1.01                    0.6        +0.0        1.0      +0.0          0.6        +0.0
crts_alpha_1.2                     0.3        +0.0        0.5      +0.0          0.1        +0.0
crts_alpha_1.5                     0.1        +0.0        0.1      +0.0          0.0        +0.0
crts_alpha_2.0                     0.0        +0.0        0.0      +0.0          0.0        +0.0
wcrps_left                         0.1        +0.0        0.0      +0.0          0.0        +0.0
wcrps_right                        0.0        -0.0        0.0      +0.0          0.0        -0.0
wcrps_center                       0.0        -0.0        0.0      +0.0          0.0        +0.0
energy_score_beta_0.1              0.0        -0.0        0.0      -0.0          0.0        -0.0
energy_score_beta_0.3              0.0        -0.0        0.0      -0.0          0.0        -0.0
energy_score_beta_0.5              0.0        -0.0        0.0      -0.0          0.0        -0.0
energy_score_beta_0.7              0.0        +0.0        0.0      -0.0          0.0        -0.0
energy_score_beta_0.9              0.0        +0.0        0.0      -0.0          0.0        -0.0
energy_score_beta_1.0              0.0        +0.0        0.0      +0.0          0.0        +0.0
energy_score_beta_1.1              0.1        +0.0        0.1      +0.0          0.0        +0.0
energy_score_beta_1.3              0.2        +0.0        0.5      +0.0          0.1        +0.0
energy_score_beta_1.5              0.5        +0.0        1.1      +0.0          0.2        +0.0
energy_score_beta_1.7              1.4        +0.0        2.2      +0.0          0.5        +0.0
energy_score_beta_1.8              2.5        +0.0        2.9      +0.0          0.7        +0.0
energy_score_beta_1.9              4.2        -0.0        3.7      +0.0          0.9        +0.0
cde_loss                           0.0        -0.0        0.0      +0.0          0.0        -0.0
dpd_beta_0.01                      0.0        -0.0        0.0      -0.0          0.0        -0.0
dpd_beta_0.2                       0.0        -0.0        0.0      -0.0          0.0        -0.0
dpd_beta_0.5                       0.0        -0.0        0.0      -0.0          0.0        -0.0
dpd_beta_1.0                       0.0        -0.0        0.0      +0.0          0.0        -0.0
pseudospherical_alpha_1.5          0.0        -0.0        0.0      -0.0          0.0        -0.0
pseudospherical_alpha_2.0          0.0        +0.0        0.0      +0.0          0.0        -0.0
pseudospherical_alpha_3.0          0.0        +0.0        0.0      +0.0          0.0        +0.0
```
The full results are [here](https://github.com/user-attachments/files/32509955/tail_comparison_results.csv). Note that the point metrics (e.g. RMSE) change when adding the tails, which is not expected. This is because ScoringBench computes RMSE directly from the bar distribution (without tails) when computing RMSE, rather than using the mean returned by our code (which does include the tails).

So some metrics get slightly worse with this PR, but only slightly. I'd be in favour of merging?

LeoGrin:
Thanks a lot @oscarkey! I'd also be in favour of merging, the differences on ScoringBench seem very small and the change makes a lot of sense :)

LeoGrin:
And thanks a lot @ppguo for fixing this!!

ppguo:
Thanks @oscarkey for running the benchmarks and checking the impact, and @LeoGrin for the feedback! I really appreciate your time and support.

tobias-priorlabs:
@oscarkey Apologies for the late reply. I think this is a necessary change to make the cdf consistent with the bar distribution forward, so I am fully on board with merging it (thanks @ppguo!). The scoring bench metrics are somewhat sensitive to the output format of the tabular model, so it is not surprising that this can make TabPFN a tiny bit worse on ScoringBench.

src/tabpfn/architectures/shared/bar_distribution.py
```diff
@@ -501,15 +501,89 @@ def assert_support(self, *, allow_zero_bucket_left: bool = False) -> None:
...
+        prob_left_of_ys = torch.where(
+            ys >= self.borders[-2],
+            right_tail_cdf,
+            prob_left_of_ys,
+        )
+        return prob_left_of_ys.clip(0.0, 1.0)
+
+    @override
+    def icdf(self, logits: torch.Tensor, left_prob: float) -> torch.Tensor:
+        """Calculate quantiles using half-normal tails in the outer buckets."""
```
oscarkey: Might be worth adding an early exit here: `if left_prob >= 1.0: return inf`.
At the moment I think we can get a finite value, because we're relying on the cumsum reaching 1.0, which it might not do, due to rounding errors. What do you think?

oscarkey (dismissed):
great!

I got Claude to run a quick speed+memory benchmark, vs the existing functions:
```
(CPU, 20000 rows x 5000 bars, float32)
  ┌────────────────┬─────────────────────┬──────────┬─────────────────────┐
  │     Method     │        main         │    PR    │   Peak RSS delta    │
  ├────────────────┼─────────────────────┼──────────┼─────────────────────┤
  │ icdf           │ 73.7 ms             │ 59.3 ms  │ unchanged           │
  ├────────────────┼─────────────────────┼──────────┼─────────────────────┤
  │ cdf (100 ys)   │ 94.0 ms             │ 130.4 ms │ unchanged │
  ├────────────────┼─────────────────────┼──────────┼─────────────────────┤
  │ sample per row │ 32 us (Python loop) │ 76 us    │ ~3x logits size     │
  └────────────────┴─────────────────────┴──────────┴─────────────────────┘
```
As only `icdf` is used on the main `TabPFNRegressor` path, I think this is fine. Sample uses quite a lot more peak memory, but I'd imagine the overall peak memory usage will be dominated by the model itself.

src/tabpfn/architectures/shared/bar_distribution.py
ppguo: Thanks, good catch! I reproduced this with just three bars: logits `[0.56665051, 0.79350841, 0.59883946]` returned about `8.66` for `icdf(1.0)` in float32 (and a finite value in float64 as well). The conditional probability can round below 1, so clamping alone does not guarantee an infinite endpoint.

Fixed in a9e259c with explicit early returns for `left_prob == 0.0` and `left_prob == 1.0`, preserving batch shape, device, and dtype. The interior quantile calculation is unchanged. Added regression coverage for both endpoints with scalar and multidimensional batches, float32/float64, and the existing device parametrization.

All four new CPU cases failed before the fix; the full `tests/test_architectures/test_bar_distribution.py` now passes: **23 passed** (Python 3.12.14, PyTorch 2.14.0, CPU). Ruff check and format check also pass. CUDA was unavailable in this local test environment; the updated PR CI will provide additional coverage.

oscarkey (approved):
sweet! Thank you @ppguo !

PR #1218, opened 2026-08-31
perf: fuse float64 upcast with defensive copy

arthur-priorlabs:
## Issue

[RES-2439 — Reduce the RAM footprint of the wrapper](https://linear.app/priorlabs/issue/RES-2439/reduce-the-ram-footprint-of-the-wrapper)

## Motivation and Context

**What**

- `clean_data` no longer copies a numeric input to float64. The preprocessing pipeline already copies the table before it runs, so that copy does the conversion instead.
- `predict` now uses the same shortcut, as `clean_data_transform`.

**Why**

- The float64 table was 10.67 GB on a 666,667 x 2,000 float32 fit, and stayed in memory for all of `fit`.
- `predict` built two float64 copies where one was needed.

**How**

- float32 to float64 is exact, so every step sees the same values and the ensemble members are bit-identical.
- The feature-importance model converts its own subsample, so its ranking does not change either.

---

## Public API Changes

-   [x] No Public API changes

`clean_data` / `clean_data_transform` now return a numeric input's own dtype rather than always float64. Neither is exported from `tabpfn`, and `fit` / `predict` outputs are unchanged — but flagging it in case you read that boundary as public.

---

## How Has This Been Tested?

Peak host RSS over a full fit + predict at 1M x 2000 float32, one estimator, `tabpfn_v3`, measured with the wrapper profiler in fomo-fitting:

| stage | before | after |
| :--- | ---: | ---: |
| `total: fit + predict` | 26.10 GB | **15.44 GB** |
| `clean_data` retained | 10.67 GB | 0.01 GB |
| `predict` transient | 10.67 GB | 1.61 GB |

<details><summary>Gates against f213434d, 666,667 x 2,000, outputs/members identical on every cell</summary>

Every cell runs both sides of its own comparison in one allocation, on 16 vCPU / 128 GB with 8 threads, so a delta is never contaminated by which machine it landed on. `current` is the repo environment (python 3.14, numpy 2.5.2, pandas 3.0.5), which is also the `highest` end of the supported range; `lowest` is python 3.10, numpy 1.22.0, pandas 1.4.0.

| gate | mix | transient RSS | median wall time |
| :--- | :--- | :--- | :--- |
| `clean_data --stage predict`, current | numeric | 21.33 → 0.00 GB | 13.486 → 0.000 s |
| | half-string | 27.34 → 27.34 GB | 158.5 → 143.7 s |
| | numeric-object | 22.02 → 22.02 GB | 160.0 → 159.4 s |
| `clean_data --stage predict`, lowest | numeric | 21.33 → 0.00 GB | 13.577 → 0.000 s |
| | half-string | 27.34 → 27.34 GB | 263.5 → 264.0 s |
| | numeric-object | 26.69 → 26.68 GB | 276.2 → 278.7 s |
| `clean_data --stage fit`, current | numeric | 10.67 → 0.00 GB | 6.337 → 0.005 s |
| | half-string | 18.78 → 18.78 GB | 222.5 → 218.7 s |
| | numeric-object | 22.03 → 22.03 GB | 171.6 → 163.9 s |
| `clean_data --stage fit`, lowest | numeric | 10.67 → 0.00 GB | 5.621 → 0.003 s |
| | half-string | 18.78 → 18.78 GB | 344.2 → 347.9 s |
| | numeric-object | 26.70 → 26.70 GB | 345.1 → 337.6 s |
| `bench_ensemble_preprocessing`, current | numeric | 12.01 → 12.01 GB | 10.944 → 5.315 s (**−51.4%**) |
| | half-string | 28.01 → 28.01 GB | no change |
| | numeric-object | 12.01 → 12.01 GB | 11.104 → 11.338 s |
| `bench_ensemble_preprocessing`, lowest | numeric | 12.01 → 12.01 GB | 16.420 → 8.816 s (**−46.3%**) |
| | half-string | 28.01 → 28.01 GB | 231.4 → 240.6 s |
| | numeric-object | 12.01 → 12.01 GB | 16.292 → 16.583 s |

The numeric cells of `clean_data` now measure a no-op: for a numeric input with nothing to encode, cleaning costs nothing at either end of the range, and the widening moved into the per-member copy. `bench_ensemble_preprocessing` is the gate that sees where that work landed — and it halves there, on both pandas 1.4 and pandas 3, with members identical.

The object mixes are unchanged on RAM to three significant figures. Their wall times scatter a few percent either way and should be read as no change: at this shape a single half-string ensemble comparison is not resolvable at the 5% tolerance — five repeats of the same two commits gave +17.0%, +4.3%, +0.02%, +1.7%, −6.3%, with the reference alone spanning 80.6–116.7 s across allocations.

The gate changes these numbers were produced with live on `arthur/benchmarking-update` (`scripts/` only), which is why this branch touches no benchmarks.

</details>

- `tests/test_input_immutability.py`: a sweep over both estimators x five input kinds asserting `X_train`, `X_test` and `y_train` come back bit-identical, plus a check that the pipeline copies before the first in-place writer — verified to fail against a build with that copy removed, since the sweep alone does not (`_extract_inf_masks` corrupts the caller's array and `_restore_inf_masks` tidies up before `fit` returns).
- `test__clean_data_transform__*` in `test_data_cleaning.py`: the shortcut matches the general path exactly, and does not skip an encoder fitted on a *text* column — a bug the first version of this had, since text is encoded at fit yet never appears in `indices_for(CATEGORICAL)`.

Full suite: 1809 passed, 145 skipped, 12 xfailed.

---

## Checklist

---

### DIFF
--- a/tests/test_input_immutability.py
+++ b/tests/test_input_immutability.py
@@ -31,9 +31,12 @@ import pytest
 from tabpfn import TabPFNClassifier, TabPFNRegressor
 from tabpfn.preprocessing import pipeline_interface
 
-N_ROWS = 80
-N_TEST = 20
+N_ROWS = 12
+N_TEST = 6
 N_FEATURES = 5
+# Spacing of the missing values in the object column, kept at or below `N_TEST`
+# so the smaller split cannot come out without one.
+N_MISSING_EVERY = 5
 # +/-inf reaches the ordinal encoder only with this on; without it validation
 # rejects the input before any of the code under test runs.
 PASSTHROUGH_INF = {"PASSTHROUGH_INF": True}
@@ -62,7 +65,7 @@ def _object_mixed() -> np.ndarray:
     # Three levels, so it detects as categorical rather than as free text.
     X[:, 1] = [f"lvl{i % 3}" for i in range(n)]
     X[:, 2] = rng.integers(0, 5, n)
-    X[:, 3] = [None if i % 17 == 0 else f"k{i % 2}" for i in range(n)]
+    X[:, 3] = [None if i % N_MISSING_EVERY == 0 else f"k{i % 2}" for i in range(n)]
     return X
 
 
--- a/tests/test_input_immutability.py
+++ b/tests/test_input_immutability.py
@@ -1,0 +1,229 @@
+#  Copyright (c) Prior Labs GmbH 2026.
+
+"""Fitting and predicting must leave the caller's data exactly as it was.
+
+`clean_data` hands a numeric array straight back rather than casting it into one
+it owns, so the array the wrapper carries through `fit` *is* the caller's. What
+keeps that safe is `PreprocessingPipeline._process_steps` taking a private copy
+before any step mutates anything -- one guarantee, in one place, rather than a
+defensive copy at every boundary. Several things inside cleaning and preprocessing
+write in place quite deliberately (`_extract_inf_masks` NaNs out infinities,
+`process_text_na_dataframe` is documented to mutate the frame it is given), so the
+question is not whether anything mutates but whether any of it reaches back to the
+caller.
+
+Every input is checked, not just the training features: `y` is label-encoded and
+z-normalised on the way through, and the test features take a different route
+again -- through `clean_data_transform` and a fitted encoder rather than a fresh
+one. Each input kind below reaches cleaning by a different path: a plain numeric
+array takes the single-cast shortcut, an object array and a DataFrame go through
+pandas, and NaN and +/-inf are rewritten in place internally before the steps run.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import numpy as np
+import pandas as pd
+import pytest
+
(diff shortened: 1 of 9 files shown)


tests/test_input_immutability.py
```diff
@@ -0,0 +1,229 @@
...
+from typing import Any
+
+import numpy as np
+import pandas as pd
+import pytest
+
+from tabpfn import TabPFNClassifier, TabPFNRegressor
+from tabpfn.preprocessing import pipeline_interface
+
+N_ROWS = 80
```
oscarkey: Might be worth making this a bit smaller to be as fast as possible?

oscarkey (dismissed):
nice!

tests/test_input_immutability.py
arthur-priorlabs: Done in dfb2799d; might be worth investigating the dozen of model instantiations in the test suite in a dedicated PR

PR #1219, opened 2026-08-31
fix: key the built-model cache on device and inference dtype

jmkuebler:
## Issue

RES-2743. Stacked on #1250, which removes the dead `cache_trainset_representation` flag and lets `fit_mode="fit_with_cache"` use the cache.

## Motivation and Context

The built-model cache is opt-in via `TABPFN_MODEL_CACHE_SIZE` and hands one built model to every estimator that asks for it, so a warm fit costs ~0.003 s instead of ~0.12 s.

**It could not be shared across placements.** An estimator applies the placement to the module it is handed, in place — the inference engine moves it with `model.to(device)` and casts it with `model.type(dtype)`. Neither was part of the key, so two estimators wanting different placements were served the same instance:

```python
a = TabPFNClassifier(device="cpu").fit(X, y); a.predict_proba(X_test)
b = TabPFNClassifier(device="mps").fit(X, y); b.predict_proba(X_test)
a.predict_proba(X_test)
# RuntimeError: Placeholder storage has not been allocated on MPS device!
```

```python
TabPFNClassifier(inference_precision=torch.float16).fit(X, y).predict_proba(X_test)
TabPFNClassifier().fit(X, y).predict_proba(X_test)
# RuntimeError: mat1 and mat2 must have the same dtype, but got Float and Half
```

The dtype cannot be repaired instead of keyed: `model.type(torch.float16)` is destructive, so a model cast to fp16 can never serve a full-precision caller.

### The suite is the clearest evidence

Setting `TABPFN_MODEL_CACHE_SIZE` over `tests/test_classifier_interface.py` and `tests/test_regressor_interface.py`:

| | result | time |
|---|---|---|
| cache off | 309 passed | 146.4 s |
| cache on, before this change | **38 failed** | — |
| cache on, after | 309 passed | **116.8 s** |

Every one of the 38 was a placement collision — `Float and Double`, `Float and Half`, `Cannot compare two tensors on different devices`, `Cannot convert a MPS Tensor to float64`. The suite spans cpu/mps and three forced precisions, so it could not opt in at all before. On top of #1250, where `fit_with_cache` also hits the cache, the same suites show 64 collisions before this change and none after.

Those two files are unusually fit-dense. Over the whole suite the effect is smaller — 1861 passed either way, 398 s off versus 368 s on, so ~30 s or 7.5%. Enabling it in `conftest.py` is a follow-up, not part of this PR.

### What a caller still owns

The default is unchanged, so nothing happens unless the env var is set. Mixing devices or precisions now yields separate entries rather than corruption. Two estimators served the *same* entry still share the module, so moving one with `.to()` after fit or training its weights reaches the other — deliberate, and left to the caller. Calling `.to()` before fit is safe: the key is built from the device the estimator has at fit time.

---

## Public API Changes

-   [ ] No Public API changes

- `tabpfn.model_loading.load_model` and `load_model_criterion_config` take optional `devices` and `force_inference_dtype` keyword arguments (both default `None`, so existing calls are unaffected — a caller that places the returned model should pass them).
- A non-integer `TABPFN_MODEL_CACHE_SIZE` warns instead of being silently ignored.
- The cache default is unchanged (off).

---

## How Has This Been Tested?

- Full suite in the shipping configuration (cache off): **1861 passed, 176 skipped, 8 xfailed**.
- Full suite with `TABPFN_MODEL_CACHE_SIZE=8`: the same **1861 passed, 176 skipped, 8 xfailed**, so the cache is exercised across every device and precision the suite covers.
- After rebasing on #1250: `tests/test_classifier_interface.py`, `tests/test_regressor_interface.py` and `tests/test_model_cache.py` with `TABPFN_MODEL_CACHE_SIZE=8` on cpu and cuda (A100): all pass.
- `tests/test_model_cache.py`: separate entries per device and per dtype; unspecified devices kept apart from device-specific ones; disabled by default; two entries holding a classifier and a regressor; invalid env value; LRU eviction; cache-clear forcing a rebuild; and a tripwire on `load_model`'s parameters so a future parameter cannot quietly bypass the key.
- Checked end-to-end on cpu, mps and cuda with the cache on: fitting on one device then the other leaves both estimators correct, and fp16 then full precision works.

---

## Checklist


jmkuebler:
@cursor review

jmkuebler:
@cursor review

jmkuebler:
@cursor review

src/tabpfn/architectures/shared/bar_distribution.py
```diff
@@ -52,8 +52,12 @@ def __init__(self, borders: torch.Tensor, *, ignore_nan_targets: bool = True):
         self.to(borders.device)
 
     def has_equal_borders(self, other: BarDistribution) -> bool:
-        """Check if two BarDistributions have equal borders."""
-        return torch.equal(self.borders, other.borders)  # pyright: ignore[reportArgumentType]
+        """Check if two BarDistributions have equal borders.
+
+        The two may sit on different devices — a cached model carries the device
+        of the fit that placed it — so compare values, not placement.
```
oscarkey, line 58:
```suggestion
        The two may sit on different devices.
```

src/tabpfn/base.py
```diff
@@ -124,6 +126,10 @@ def initialize_tabpfn_model(
             model, or None if they did not ask for one. Only used to decide whether
             checkpoints are allowed to disagree on their temperature; the override
             itself is applied by the caller.
+        devices: The devices the caller will place the loaded models on, used to
+            key the built-model cache. None when the caller will not move them.
+        force_inference_dtype: The dtype the caller will cast the loaded models
+            to, also part of the built-model cache key.
```
oscarkey:
```suggestion
            to. None when the caller will not force a dtype.
```

src/tabpfn/model_loading.py
```diff
@@ -914,26 +928,53 @@ def _load_checkpoint_cached(path: str, _identity: tuple[int, int]) -> dict:
...
+# entry hold the same module, so moving or training either reaches the other.
+#
+# `cache_trainset_representation` is not part of the key: every architecture's
+# ``get_architecture`` deletes it, so one entry serves every fit mode.
+_DEFAULT_BUILT_MODEL_CACHE_SIZE = 0
+
+# Where the module is moved and what dtype it is cast to — applied in place by
+# whoever is handed it, so both belong in the key.
+_Placement = tuple[tuple[str, ...] | None, str | None]
+_BuiltModelCacheKey = tuple[
```
oscarkey, line 946: "Placement" wasn't an obvious word to me, what do you think about not defining it and just including everything directly in the main key, like:
```python
_Devices = tuple[str, ...] | None
_ForceInferenceDType = str | None
_BuiltModelCacheKey = tuple[str, tuple[int, int], Literal["regressor", "classifier"], _Devices, _ForceInferenceDType]
```

src/tabpfn/model_loading.py
```diff
@@ -914,26 +928,53 @@ def _load_checkpoint_cached(path: str, _identity: tuple[int, int]) -> dict:
...
-# per-group models, or servers that manage their own concurrency). RES-2422
-# tracks the follow-up that externalises per-fit state so a single backbone can
-# be shared across threads too.
-_BUILT_MODEL_CACHE: OrderedDict[tuple[str, tuple[int, int]], tuple] = OrderedDict()
+# Bounded LRU of *built* models, keyed by (path, file identity, estimator type,
+# placement). Off unless ``TABPFN_MODEL_CACHE_SIZE`` is a positive integer; 2
+# holds a classifier and a regressor. Entries are shared by reference and left in
+# ``eval()`` mode, for repeated sequential fit/predict.
+#
+# Sharing is what a caller who enables it takes on: two estimators served one
```
oscarkey: Maybe: When the cache is enabled, two estimators may share the same underlying model. Thus, callers must be aware that changing the device of one will affect the other.

src/tabpfn/model_loading.py
```diff
@@ -914,26 +928,53 @@ def _load_checkpoint_cached(path: str, _identity: tuple[int, int]) -> dict:
...
+# entry hold the same module, so moving or training either reaches the other.
+#
+# `cache_trainset_representation` is not part of the key: every architecture's
```
oscarkey, line 939: Maybe: We don't include the `cache_trainset_representation` kwarg of `ArchitectureModule.get_architecture()` in the key, because it is no longer used since `InferenceEngineCacheKV` was removed.

src/tabpfn/model_loading.py
oscarkey, line 936: What do you think about disabling `.to()` when the cache is enabled? Because the key is based on the initialisation options, not the actual device, you can get some confusing behaviour e.g.
```python
a = TabPFNClassifier(device="cpu").to("cuda")
b = TabPFNClassifier(device="cpu") # actually on cuda
```

oscarkey (commented):
hey! One thing re `.to()`, otherwise just comments on claude on comments.

I'm wondering if another option might be to replace the current cache with an extension of `tabpfn.inference._PerDeviceModelCache`, which might allow us to neatly have `.to()` work and properly support multi-device inference (atm the model is only cached on one device?). But, the current approach seems pragmatic as it's not part of the public api.

src/tabpfn/model_loading.py
oscarkey: oh no wait I'm wrong! Actually both a and b will end up on cpu. So that's mostly covered by this comment already.

src/tabpfn/model_loading.py
jmkuebler: Thank you. taking your suggestion

src/tabpfn/model_loading.py
jmkuebler: I now went for the real fix and opened a PR to entirely wipe cache_trainset_representation #1250 

I rebased the current PR on top of it.

src/tabpfn/model_loading.py
jmkuebler: so do you think we are fine as is?

jmkuebler:
@cursor review

jmkuebler:
@cursor review

src/tabpfn/model_loading.py
oscarkey: amazing, thank you!

src/tabpfn/model_loading.py
oscarkey: yes seems fine to me as people know they have to be very careful with this option.

oscarkey (approved):

PR #1220, opened 2026-08-31
Move the softmax temperature into the inference config

bejaeger:
Closes [RES-2744](https://linear.app/priorlabs/issue/RES-2744/move-softmax-temperature-to-inference-config-so-it-can-be-configured).

The softmax temperature was hard-coded in the estimator interface (`softmax_temperature: float = 0.9`), so every checkpoint was run at the same temperature and a checkpoint could not ship the value it was trained for.

## The transition

`InferenceConfig.SOFTMAX_TEMPERATURE` is the new home, with a field default of `0.9` — exactly what the interface used to apply. That single choice covers backwards compatibility for everything that exists today:

| checkpoint | how it gets its value | result |
|---|---|---|
| v2 / v2.5 | `_get_v2_config` / `_get_v2_5_config` don't pass it → field default | 0.9 |
| existing v2.6 / v3 | `inference_config` in the checkpoint has no such key → field default | 0.9 |
| new checkpoints | key present, explicitly declared | whatever they declare |

The field default is frozen at the legacy value forever (that is the contract the class docstring already states), and any checkpoint that knows better overrides it. No version sniffing, no migration table.

The `softmax_temperature` argument is **not** deprecated — it stays as the override, and now defaults to `"auto"`, meaning "use what the checkpoint declares". Temperature is the most-swept knob in the interface, and pushing users to `inference_config={"SOFTMAX_TEMPERATURE": ...}` for a routine sweep would be a downgrade; the sentinel fixes the actual problem (the hard-coded default) without touching the escape hatch. Deprecating the argument later, once there is one obvious way to set this, is still on the table.

### What users observe

- `TabPFNClassifier()` on today's checkpoints → 0.9, unchanged.
- `TabPFNClassifier(softmax_temperature=0.8)` → 0.8, unchanged.
- `TabPFNClassifier()` on a new checkpoint → whatever it declares. This is the point of the change.
- Someone who defensively pinned `softmax_temperature=0.9` keeps 0.9 on new checkpoints. Correct sklearn semantics for an explicit argument, but worth knowing.

## Two ways to name a temperature, never both

There are now two ways to override the checkpoint — the `softmax_temperature` argument and `SOFTMAX_TEMPERATURE` in `inference_config` — and setting both is rejected rather than silently resolved, since the call gives no clue which one wins:

```
ValueError: The softmax temperature was given twice: `softmax_temperature=1.5` and
`inference_config` with SOFTMAX_TEMPERATURE=0.7. Pass it one way or the other, so
which one applies is unambiguous.
```

A hand-built `InferenceConfig` always carries a temperature (the field has a default), so passing one counts as naming it; an `inference_config` dict that does not mention the temperature stays fine alongside the argument. Both estimators' tuning paths rebuild themselves through `get_params(deep=False)`, so they reproduce the user's original combination and cannot trip this.

## Ensembles

One temperature applies to the whole ensemble, so checkpoints that disagree on it are rejected rather than reconciled:

```
ValueError: The given model checkpoints declare different softmax temperatures
([0.9, 1.0]), and one temperature is applied to the whole ensemble. Pick the one to
use for all of them and pass it explicitly, e.g.
`TabPFNClassifier(softmax_temperature=0.9, ...)`.
```

Doing what the error says makes the same ensemble load, so the user's override is resolved *before* loading and passed down as `softmax_temperature_override`; it is what decides whether the checkpoints have to agree. Any other config mismatch keeps its existing error, since naming a temperature cannot fix those.

## Release ordering

`InferenceConfig` is `extra="forbid"`, so a checkpoint carrying `SOFTMAX_TEMPERATURE` will fail to load on any release predating the field (the same situation as `MIN_CARDINALITY_FOR_TEXT` in #1205). This library release has to land before such a checkpoint is published.

## Testing

- New tests in `test_inference_config.py` cover the resolution order, the legacy default, `equals_ignoring_softmax_temperature`, both raises (named twice, and disagreeing checkpoints), both ways of resolving the ensemble case, an `inference_config` that does not name a temperature still being allowed alongside the argument, and that a checkpoint's temperature reaches the predictions identically to passing the same value.
- New tests in `test_model_loading.py` cover the ensemble raise, loading with an override, agreeing checkpoints, a non-temperature mismatch still raising even with an override, and a checkpoint with the key deleted.
- `test_consistency.py` (reference predictions), both interface suites, save/load, finetuning, and inference tuning all pass unchanged.


src/tabpfn/finetuning/finetuned_classifier.py
```diff
@@ -235,10 +235,11 @@ def _create_estimator(self, config: dict[str, Any]) -> TabPFNClassifier:
 
     @override
     def _setup_estimator(self) -> None:
```
LeoGrin: I think we can now remove this thing entirely, as it was already not doing anything for regression

LeoGrin (dismissed):
LGTM! one small nit.

One thing I realize while reviewing this is that passing an InferenceConfig object as inference_config overrides all of a checkpoint default config with the InferenceConfig class defaults, which can be misleading. Opened #1224 as a follow-up

src/tabpfn/finetuning/finetuned_classifier.py
bejaeger: Done — removed in 44e1d20d. Both implementations were empty (the classifier's was the only one that ever did anything, and this PR made it a no-op), so the abstract method, both overrides, and the call site are all gone. `_initialize_model_variables()` ran on the very next line anyway.

PR #1223, opened 2026-08-31
perf: chunked TorchSquashingScaler

arthur-priorlabs:
## Issue

[RES-2731 — SquashingScaler 10M x 2k OOM](https://linear.app/priorlabs/issue/RES-2731/squashingscaler-10m-x-2k-oom)

## Motivation and Context

**What**

-   `TorchSquashingScaler` now operates in fixed-byte blocks, so the VRAM peaks are roughly constant instead of proportional to the input.
-   `TorchSquashingScaler.fit`: -97% transient VRAM; -7.5% wall time in torch 2.10+, **but +5% wall time on torch ≤2.9**.

**Why**

-   `TorchSquashingScaler.fit` needed **13.0x its input** in VRAM and `transform` **5.8x**
-   Nearly all of `fit`'s peak is the sort inside `torch.nanquantile`, which double-buffers the sorted values *and* their int64 indices and so reaches ~12x the bytes handed to it.
-   `transform` stacked up full-size `where`/`full_like`/`zeros_like` buffers on the way to an output that only ever needed to exist once.

**How**

-   `fit` reduces a block of columns at a time (per-column statistics are independent, so grouping cannot change one), `transform` fills its output a block of rows at a time (every step is elementwise).
-   Elementwise work is spelled in place — `a.sub_(b)` is the same kernel as `a - b` — and `full_like` sentinels become 0-dim broadcasts, so outputs stay **byte-identical** (and that's a Claude promise!).

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?

H100 80GB, 1M x 2,000 float32, against `cb23da5f`. Transient VRAM is the peak *above* what was live on entry — what decides whether a shape fits:

| stage | input | transient VRAM | of input | median wall time |
| :--- | :--- | :--- | :--- | :--- |
| fit | 5.33 GB | **69.33 → 2.14 GB** | **13.0x → 0.4x** | 215.2 → 198.8 ms |
| transform | 8.00 GB | **46.01 → 8.04 GB** | **5.8x → 1.0x** | 110.0 → 93.7 ms |

Outputs are byte-identical at every torch from 2.5.0 to 2.13.0. `transform` is at its floor: 1.0x is the output itself.

-   `scripts/bench_squashing_scaler.py` gates byte-identity, transient VRAM and wall time in one process, and runs 28 correctness cases with every blocking budget forced to 1 so the blocked paths are exercised at small shapes too.
-   `tests/test_torch_preprocessing/`, `test_inf_handling.py`, `test_pipeline_consistency.py`: 205 passed, 39 skipped.

<details><summary>Details: the one cost, and why it is bounded</summary>

Blocking the sort is not free. It costs ~5% of `fit`'s wall time on torch ≤2.9 while saving ~8% on ≥2.10, and no block size wins on both — bisected across nine releases:

| torch | fit: reference | working tree | change |
| :--- | :--- | :--- | :--- |
| 2.5.0 | 215.7 ms / 69.33 GB | 226.3 ms / 2.14 GB | +4.9% |
| 2.9.0 | 214.9 ms / 69.33 GB | 221.7 ms / 2.14 GB | +3.2% |
| 2.10.0 | 214.8 ms / 69.33 GB | 198.7 ms / 2.14 GB | −7.5% |
| 2.11.0 | 215.0 ms / 69.33 GB | 198.7 ms / 2.14 GB | −7.6% |
| 2.13.0 | 215.2 ms / 69.33 GB | 198.7 ms / 2.14 GB | −7.6% |

The reference column barely moves across nine releases, so torch did not speed up the unblocked sort — it sped up the blocked one, and that is where the whole swing lives. The VRAM columns are identical everywhere, so the memory win does not depend on the torch underneath; only its wall-time cost does.

`_FIT_UNBLOCKED_BYTES` (512 MB) bounds that cost: below it `fit` takes the old single-pass route unchanged, so the ~5% is only ever paid where the alternative is a 69 GB allocation. That threshold is above anything the test suite or a CPU/MPS caller reaches.

</details>

<details><summary>Details: block sizes</summary>

Both budgets were swept on an H100 at 1M x 2,000 rather than guessed, and the sweeps are recorded beside each constant. `_FIT_BLOCK_BYTES` = 128 MB is the best of both axes (2.14 GB / 198.5 ms; 512 MB and above lose badly on time *and* memory). `_TRANSFORM_BLOCK_BYTES` = 32 MB sits at the small end because time is flat across the whole range while memory is not (32 MB → 1.01x / 94.9 ms, versus no blocking → 2.25x / 91.6 ms).

</details>

---

## Checklist

---

### DIFF
--- a/src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
+++ b/src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
@@ -14,85 +14,37 @@ Original skrub attribution:
   All rights reserved.
   SPDX-License-Identifier: BSD-3-Clause
 
-The state is returned explicitly from ``fit`` rather than stored on the
-instance, matching the rest of ``preprocessing/torch``.
-
-Memory
-------
-
-Both halves are written to keep their device footprint down, because at the shapes
-this runs at they are dominated by their own temporaries rather than by their input.
-Measured with ``scripts/profile_squashing_scaler.py`` on an H100:
-
-* ``fit`` is essentially one ``torch.nanquantile`` call, whose internal sort peaks at
-  roughly **12x the bytes it is handed** (radix sort double-buffers both the values
-  and their int64 indices). So the reduction is done in blocks of columns of a fixed
-  byte size, which turns that peak from a multiple of the input into a constant. It is
-  only done above ``_FIT_UNBLOCKED_BYTES``, because blocking the sort costs a few
-  percent of fit's wall time on older torch and is not worth paying where 12x already
-  fits.
-* ``transform`` has to produce a full-size output, so its floor is 1x; what it must
-  not do is stack up further full-size buffers on the way there. It allocates the
-  output once and fills it a block of rows at a time, so every mask and temporary it
-  needs is the size of a block rather than of the whole tensor.
-
-The elementwise rewrites are bit-for-bit equivalent to the straightforward spelling
-(``a.sub_(b)`` is the same kernel as ``a - b``, writing to a different place), and the
-blocking is over axes the maths is independent along -- per column for a per-column
-reduction, per row for an elementwise map. Neither is a change to the arithmetic,
-and ``scripts/bench_squashing_scaler.py`` gates that: it fails unless the outputs stay
-byte-identical.
+The state is returned explicitly from `fit` rather than stored on the
+instance, matching the rest of `preprocessing/torch`.
 """
 
 from __future__ import annotations
 
 import torch
 
-# Bytes of input one ``fit`` column block is allowed to cover. The sort inside
-# ``torch.nanquantile`` peaks at ~12x the bytes it is given, and that multiple -- not
-# the input size -- is what put a 10M x 2,000 fit out of reach on an 80 GB card.
-# Blocking by a fixed *byte* size rather than a fixed number of blocks makes fit's
-# peak a constant (~13x this) at every shape, instead of a constant multiple of an
-# input that keeps growing.
-#
-# Swept on an H100, fit on [666667, 1, 2000] (5.33 GB), transient VRAM against median
-# wall time: 32 MB -> 0.52 GB / 212.0 ms, 64 MB -> 1.08 GB / 202.6 ms, 128 MB ->
-# 2.14 GB / 198.5 ms, 256 MB -> 4.28 GB / 210.1 ms, 512 MB -> 6.97 GB / 243.6 ms,
-# 1 GB -> 13.94 GB / 248.0 ms. 128 MB is the best of both: large blocks lose badly on
-# time as well as memory, and the smallest ones start paying for the extra launches.
+# Bytes of input one `fit` column block covers. The sort inside
+# `torch.nanquantile` needs a large multiple of what it is given -- it double-buffers
+# both the sorted values and their int64 indices -- so bounding the block by bytes is
+# what bounds fit's peak at every shape. Sits at the knee: larger blocks cost time as
+# well as memory, smaller ones start paying for the extra launches.
 _FIT_BLOCK_BYTES = 128 * 1024 * 1024
 
-# Below this much input, ``fit`` does not block at all. Blocking the sort is not free:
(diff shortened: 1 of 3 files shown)


src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
```diff
@@ -14,18 +14,53 @@
...
+# large enough to be a problem.
+_FIT_UNBLOCKED_BYTES = 512 * 1024 * 1024
+
+# Bytes of output `transform` produces at a time. Its output has to exist in full;
+# none of its intermediates do. Small end of the useful range, because wall time is
+# flat across that range while what the intermediates cost scales with the block.
+_TRANSFORM_BLOCK_BYTES = 32 * 1024 * 1024
+
+
+def _scalar(value: float, like: torch.Tensor) -> torch.Tensor:
```
priorphil, line 43: I don't think we need this, torch.where supports `torch.where(x < 0, float("nan"), x)`

src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
```diff
@@ -173,31 +247,59 @@ def transform(
...
+        # buffer holds "is not NaN, and in a zero column".
+        finite_in_zero_column = torch.isnan(x)
+        finite_in_zero_column.logical_not_()
+        finite_in_zero_column &= zero_mask
+        out.masked_fill_(finite_in_zero_column, 0.0)
+        del finite_in_zero_column
+
+        # In-place form of `z / torch.sqrt(1.0 + (z / b) ** 2)`: the same kernels in
+        # the same order, needing one temporary rather than a chain of them.
+        denominator = out / b
```
priorphil, line 290: Given that this is harder to read, how much of that do we really need since this is blocked now? This should allocate only ~2x the memory for the chain since they can reuse the memory via the caching allocator?

src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
```diff
@@ -146,6 +157,69 @@ def fit(self, x: torch.Tensor) -> dict[str, torch.Tensor]:
...
+        )
+
+    def _column_statistics_block(
+        self,
+        x: torch.Tensor,
+        qs: torch.Tensor,
+        quantile_dtype: torch.dtype,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """min, max and quantiles for one block of columns."""
+        x_finite = _replace_inf_with_nan(x)
```
priorphil: That sounds confusing to me - x_finite still seems to contain nans, so it's not finite. I realize this was there before but maybe we can fix it while at it.

priorphil (dismissed):

src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
arthur-priorlabs: Renamed to `x_masked` in 388db0c1

src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
arthur-priorlabs: You're right about VRAM; this in-place block doesn't buy much memory (8.10 → 8.04 GB).
However, it buys −17.9% `transform` wall time (115.4 → 94.8 ms) at 1M x 2k.
In principle I'd tend towards keeping the in-place logic, seeing as the out-of-place version is clearly documented in comment form.

src/tabpfn/preprocessing/torch/torch_squashing_scaler.py
arthur-priorlabs: Not with the in-place syntax: `torch.where(x < 0, float("nan"), x, out=y)` fails; removed `_scalar` where possible in ff6e3497.

arthur-priorlabs:
@priorphil hope this addresses your comments

priorphil:
SG, thanks!

PR #1224, opened 2026-08-31
Deprecate passing an InferenceConfig object as inference_config

LeoGrin:
I've added a future warning but kept the option for now, could also delete it directly if you think it's better and safe (no usage in tabpfn-extensions, a few minor internal usages that are easy to change, and unknown usage by external users).

## Claude summary

Stacked on #1220.

`inference_config` takes two shapes with opposite semantics:

```python
inference_config={"POLYNOMIAL_FEATURES": "all"}   # merge: unnamed fields keep the checkpoint's value
inference_config=InferenceConfig(...)             # replace: unnamed fields take a class default
```

This deprecates the second with a `FutureWarning`. Behavior is unchanged.

## Why

The class defaults are a backwards-compatibility artifact, not a configuration — the class docstring says so, and points at `get_default()` for the real per-version values. Passing an object adopts the artifact for every field it does not name.

`PREPROCESS_TRANSFORMS` is the only required field, so the other 23 default silently. Historically those failed loudly: `MAX_NUMBER_OF_FEATURES` and `MAX_NUMBER_OF_SAMPLES` reject data against a pretraining limit. `SOFTMAX_TEMPERATURE`, added in #1220, is the first that is silently applied and numerically consequential — a checkpoint declaring 0.42 runs at 0.9 when the caller passes a config object that mentions no temperature.

It also weakens two guards in #1220. An object always carries a temperature, so `_resolve_softmax_temperature_override` reads it as named. An object that says nothing about temperature can therefore trip the "given twice" error quoting a value the caller never wrote, and disarm the check that rejects ensembles whose checkpoints declare different temperatures. Both go away with the object branch.

## Migration

Naming every key in a dict *is* wholesale replacement, so `dataclasses.asdict(config)` reproduces the object form byte-identically. Tested, including the parts that could round-trip badly: `PREPROCESS_TRANSFORMS` deep-converts to dicts and pydantic revalidates them, `REGRESSION_Y_PREPROCESS_TRANSFORMS` stays a tuple, and the two private `_*_DEFAULT_OUTLIER_REMOVAL_STD` fields pass through.

Callers who want a pinned, checkpoint-independent config keep it. The replacement just becomes visible at the call site instead of implied by the type.

## Scope

No object-form callers in this repo outside tests. `ClassifierModelSpecs`/`RegressorModelSpecs(inference_config=...)` is a different field — the checkpoint side, where replacing wholesale is correct — and is unaffected.

I checked the downstream code we maintain: a handful of sweep and eval scripts pass objects to pin a known config, each a one-line `asdict` away. Nothing needs the object form for something the dict form cannot express.

## Tests

The warning fires for an object and not for a dict or `None`; the object still replaces wholesale; a dict keeps the fields it does not name; `asdict` matches the object form.

https://claude.ai/code/session_01DEKRW8D4Z9uBHkLovuj2at


bejaeger (dismissed):
Thanks for catching this. I think deprecating makes sense to avoid surprises.

bejaeger (approved):

LeoGrin:
@oscarkey could we get a stamp here, it seems the reapprove workflow didn't work here and Ben can't reapprove as he merged to main

oscarkey (approved):

PR #1226, opened 2026-09-01
Add operating system-specific install instructions to the README.

### LINKED ISSUE #158: [docs] Add supported operating systems, CPU vs GPU install

oscarkey:
- On macOS: update suggested minimum PyTorch version from dev version of 2.13 to release version of 2.13
- On Linux: previous instructions installed a CUDA version, add notes for ROCm and CPU-only
- On Windows: previous instructions installed a CPU-only verison, add notes for CUDA and ROCm

Closes #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,26 +24,26 @@
 ```bash
 pip install tabpfn
 ```
-
-TabPFN supports Python 3.10+. `pip install tabpfn` installs a compatible
-PyTorch build automatically. If you want a smaller CPU-only install, need a
-PyTorch build for a specific accelerator or CUDA version, or want platform
-notes for Windows and WSL, choose the matching command from the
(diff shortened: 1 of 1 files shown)


TuanaCelik (approved):
lgtm!

PR #1225, opened 2026-09-01
Detect dates by dtype only, and refuse a native datetime column with a clear error

alanprior:
Two things bundled into one small PR:

1. **Dates are recognized only from a genuine `datetime64` dtype**, never by running `pd.to_datetime()` against every string/object column. `_is_date_like_pandas_series`/`_all_parse_as_dates` and the never-consumed `FeatureModality.DATE` are gone: a string column that merely looks like a date is classified like any other string.
2. **A genuine `datetime64` column is unsupported, and is now refused up front.** `handle_datetime_columns` runs on the raw DataFrame at fit and at predict and raises a `TabPFNUserError` naming each such column by index and label and telling the user to preprocess it themselves. Before, such a column mixed with any other dtype failed inside validation with numpy's opaque dtype-promotion error.

One accepted edge case: a genuine `datetime64` column declared categorical skips the check, so it still hits that validation error.

## Issue
Please link the corresponding GitHub issue. If an issue does not already exist,
please open one to describe the bug or feature request before creating a pull request.

This allows us to discuss the proposal and helps avoid unnecessary work.

## Motivation and Context

Simplifies date detection ahead of #1227, which adds an opt-in calendar expansion for such columns, and turns an opaque crash into a clear error along the way.

---

## Public API Changes

-   [x] No Public API changes
-   [ ] Yes, Public API changes (Details below)

---

## How Has This Been Tested?

---

## Checklist

-   [x] The changes have been tested locally.
-   [ ] Documentation has been updated (if the public API or usage changes).
-   [x] A changelog entry has been added (see `changelog/README.md`), or "no changelog needed" label requested.
-   [x] The code follows the project's style guidelines.
-   [x] I have considered the impact of these changes on the public API.

---

PR #1227, opened 2026-09-01
Expand a date into calendar features behind TRANSFORM_DATES, through a fitted DateTransformer

alanprior:
Stacked on #1225, which this branch's base points at: review that one first, and merge it first.

#1225 refuses a genuine datetime column with an error that tells the user to set `TRANSFORM_DATES` or preprocess the column themselves. This PR makes the first of those true. Four parts:

1. **`DateTransformer` owns temporal columns, as `date_transformer_`.** Built at fit from `TRANSFORM_DATES` and the declared categoricals, kept on the estimator exactly like `ordinal_encoder_`, and reapplied at predict through `convert_dates`, rather than each of eight call sites re-deriving what to do. With the flag off it refuses a point in time with #1225's error, at fit and at predict alike. Nothing reads a date as one plain number anywhere.
2. **`TRANSFORM_DATES` expands a date into calendar features** via `skrub.DatetimeEncoder`, off by default. `transform` reapplies exactly the encoders `fit_transform` fitted, so predict cannot come out a different width than the model was fit on: a position that is no longer a date degrades to `NaN` features, and a date at a position that was not one at fit is refused, since no encoder was fit for it. Expansion changes the column count, so the conversion also reports its resolved feature names and the declared categorical indices moved to where those columns ended up. The calendar features themselves are then typed by the usual heuristics, with no hint from the transformer: a cyclical pair from a narrow window can land as categorical, which is the same trade `MIN_UNIQUE_FOR_NUMERICAL_FEATURES` already makes everywhere else.
3. **`period` and `timedelta64` columns.** A period is a point in time like any other: refused, or expanded from the instant it starts at. A duration becomes its length in seconds, flag or no flag, and even when declared categorical: that loses nothing, and leaving it alone only crashes validation. Both aborted a `fit` before (`Unknown dtype: period[D]`, or numpy's dtype-promotion error).
4. **`n_features_in_`/`feature_names_in_` keep describing the caller's frame.** Expansion widens `X` before validation, so `validate_data` was recording TabPFN's internal view: a 2-column input fit with the flag on came out with `n_features_in_ = 14` and a `feature_names_in_` full of names the caller never wrote. New `capture_input_shape` sets or checks both on the raw input, before conversion, reading only its shape and labels; the existing entry points check values only, via `check_X_y`/`check_array`. Each predict path runs the shape check, then `convert_dates`, then value validation, in that order. `get_embeddings` had drifted from the others while nobody was looking, hand-rolling `fix_dtypes` plus `ordinal_encoder_.transform` with the declared categorical indices where every other path called `clean_data_transform` with the fitted schema; it now calls the same helper as the rest. In the same spirit, the out-of-memory message stops reading its feature count off `n_features_in_` and takes both counts off the `X` it is handed, so it names the width the transformer really ran on. The message is about memory, and expansion is exactly when the two numbers diverge.

One accepted edge case carried over from #1225: a genuine datetime column declared categorical is left alone by the transformer, so it still hits numpy's promotion error in validation.

Also of note: for a DataFrame with non-string column labels, the schema now carries the real labels (`input_0`) rather than positional stand-ins (`input_f0`), so a warning names a column the caller can find in their own frame.

## Issue

Follow-up to #1225, and supersedes the structural half of #1206.

## Motivation and Context

Calendar expansion needs per-column fitted state, reapplied identically at predict, which is what the transformer object holds. Since #1225 already refuses an unexpanded date, there is exactly one thing left for it to do.

---

## Public API Changes

-   [ ] No Public API changes

Additive: `InferenceConfig.TRANSFORM_DATES`, the `date_transformer_` fitted attribute, and `tabpfn.preprocessing.datetimes` (`DateTransformer`, `DateConversion`, `convert_dates`).

Changed, in `tabpfn.validation`: `ensure_compatible_fit_inputs` returns `(X, y, target_name)` rather than also the feature names and count, `ensure_compatible_fit_inputs_sklearn` returns `(X, y)`, and `ensure_compatible_predict_input_sklearn` no longer checks the input against `n_features_in_`/`feature_names_in_` (`capture_input_shape` does, on the raw input). New: `capture_input_shape`.

---

## How Has This Been Tested?

---

## Checklist


alanprior:
Moved to https://github.com/PriorLabs/TabPFN/pull/1236

PR #1229, opened 2026-09-01
Use a single importance ordering for every estimator

LeoGrin:
Before this PR, above 100K rows we compute feature importance on a different subset of 100K rows for each estimators. This adds diversity but:
- makes the machinery to have topK features + balanced repartitions of other features more complex, as the top k features are different for each estimator
- can we quite costly. See in the claude summary below, but for 2K features for instance, 8 estimators --> 8 orderings --> 8 x 7s on a good cpu.

So IMO we should just use one shared feature importance, computed once on a 100K subset.

I don't have a strong opinion on the exact number of rows to use for subsampling, would also be fine with e.g 200K (it's sublinear, seems to be rougly 50% more expensive than 100K).

Also tell me if you think this needs more evaluation.

## Claude summary

Stacked on #1141.

## Evaluation

Evaluated internally across the affected large/wide datasets, A/B against `main` with the same checkpoint, seed and splits. Headline: **accuracy unchanged within noise, fit time down 26-37%** on the datasets measured so far. Numbers and method in [RES-2773](https://linear.app/priorlabs/issue/RES-2773).

## Summary

#1141 restores the shared balanced pool for importance-based feature subsampling, but only below `FEATURE_IMPORTANCE_MAX_SAMPLES` (100k rows). Above it, `_collect_importance_orderings` fits one LightGBM model per independent 100k subsample, so estimators get *different* orderings. Since the consumer keys its pools by ordering, that splits the ensemble into `n_orderings` non-cooperating groups — and once `n_samples >= n_estimators * max_samples` every estimator has a private pool again, exactly the pre-#913 behaviour.

That regime is the one that ships: `auto` selects `gini_feature_importance` only when `n_samples > AUTO_FEATURE_SUBSAMPLING_IMPORTANCE_MIN_SAMPLES` (100_000), and `_collect_importance_orderings` takes its single-ordering branch only when `n_samples <= max_samples` (100_000). The two conditions are exactly complementary, so **under `auto` the single-ordering path is unreachable**.

Measured, 28 features / top-4 / budget 8 / 6 estimators, sized so one shared pool covers the tail exactly (mean features never seen by any estimator, 50 seeds):

| dataset | orderings | main | #1141 | this PR |
|---|---|---|---|---|
| 50k | 1 | 7.74 | 0.00 | 0.00 |
| 150k | 2 | 6.76 | 5.22 | 0.00 |
| 300k | 4 | 4.86 | 4.46 | 0.00 |
| 600k | 6 | 3.74 | 3.74 | 0.00 |

## Change

Fit one ordering, subsampling rows to `max_samples` when the dataset is larger. `_subsample_features_importance_based` collapses to a single shared pool, which makes it structurally identical to `_subsample_features_balanced` and `_subsample_features_constant_and_balanced`. Net -16 lines in `ensemble.py`.

Also drops up to `n_estimators - 1` LightGBM fits.

## Cost

`_collect_importance_orderings` ran `min(n_estimators, n_samples // max_samples + 1)` LightGBM fits; this PR runs one. Each fit is on `max_samples` (100k) rows, so the saving is `(k - 1)` fits of:

| features | one ordering @ 100k rows |
|---|---|
| 100 | 0.65s |
| 200 | 0.96s |
| 500 | 1.98s |
| 1000 | 3.64s |
| 2000 | 7.01s |

So ~2s saved per dropped ordering at 500 features, up to ~7s at 2000. With `n_estimators` capped at 32 by `MAX_AUTO_SCALED_N_ESTIMATORS`, the worst case removed is ~31 x 7s. The stratified `train_test_split` in front of each fit is negligible (0.02s at 200k rows, 0.10s at 1M).

Measured on 18 cores, synthetic Gaussian features, `n_estimators=50` trees, `n_jobs=-1`; real data with native categoricals will move the constants but not the shape.

## Why not keep the distinct orderings

The multi-ordering design (#908) predates the pool (#913, one day later): #908 filled the non-top budget with `rng.choice(...)`, so distinct orderings cost nothing. #913 added the coverage guarantee without reconciling the two, scoping its comment to estimators *sharing* an ordering. Coverage and per-subsample diversity want opposite things, and every extra ordering subtracts proportionally from coverage.

Keeping both is possible — put the pool in global feature-index space so it is ordering-agnostic — but it is more machinery, and it needs `scale_n_estimators_for_feature_coverage` to provision `n_total_features` draws rather than `n_total_features - top_k`. Happy to go that way instead if the diversity is worth it; there are no benchmarks on record either way.

## Follow-ups (not in this PR)

- **The 100k cap looks far too low.** LightGBM is sublinear in rows, so `FEATURE_IMPORTANCE_MAX_SAMPLES` is buying much less than it costs in ranking quality (time / LightGBM peak memory on top of X):

  | rows | 200 features | 500 features |
  |---|---|---|
  | 100k | 0.94s / 368 MB | 1.98s / 898 MB |
  | 200k | 1.64s / 701 MB | 3.41s / 1438 MB |
  | 500k | 2.39s / 739 MB | 5.50s / 1688 MB |
  | 1M | 3.70s / 1043 MB | 9.17s / 2113 MB |
  | 2M | 6.45s / 1474 MB | 16.86s / 2913 MB |

  20x the rows costs ~7-8x the time and ~4x the memory. At 500 features a single fit on all 1M rows (9.2s) is cheaper than the 8 x 100k fits this PR removes (~15.8s) while using 10x the data, and ranking fidelity to a full-data fit rises from 85.7% at 100k to 97.8% at 1M on synthetic data. Cost tracks rows x features (~3 bytes/cell at scale), so a cell budget would be better calibrated than a row count.
- `scale_n_estimators_for_feature_coverage` computes `ceil(n_features / max_features)`, ignoring that top-K is pre-committed in every estimator; the tail needs `ceil((n_features - k) / (size - k))`.

https://claude.ai/code/session_017V5aMgmDFAJxP5ymimD3vG

### DIFF
--- a/changelog/1229.fixed.md
+++ b/changelog/1229.fixed.md
@@ -1,0 +1,8 @@
+- Fixed importance-based feature subsampling losing its feature-coverage
+  guarantee on datasets above `FEATURE_IMPORTANCE_MAX_SAMPLES` (100k rows),
+  where one importance ordering was fit per row subsample and each estimator
+  ended up drawing its non-top features from a private pool. A single ordering
+  is now fit and shared by every estimator, so the balanced round-robin covers
+  all features on both sides of that threshold. Results change for fits using
+  `gini_feature_importance` (directly or via `"auto"`) on datasets larger than
+  100k rows.
--- a/changelog/1229.fixed.md
+++ b/changelog/1229.fixed.md
@@ -1,8 +1,1 @@
-- Fixed importance-based feature subsampling losing its feature-coverage
-  guarantee on datasets above `FEATURE_IMPORTANCE_MAX_SAMPLES` (100k rows),
-  where one importance ordering was fit per row subsample and each estimator
-  ended up drawing its non-top features from a private pool. A single ordering
-  is now fit and shared by every estimator, so the balanced round-robin covers
-  all features on both sides of that threshold. Results change for fits using
-  `gini_feature_importance` (directly or via `"auto"`) on datasets larger than
-  100k rows.
+Above 100k rows, importance-based feature subsampling gave every estimator its own feature ranking and its own sampling pool, so some features ended up in no ensemble member at all even when the combined feature budget covered them. A single shared ranking now restores full coverage, and replaces up to `n_estimators` LightGBM fits with one. Results change for `gini_feature_importance` fits (directly or via `"auto"`) above 100k rows.
--- a/src/tabpfn/preprocessing/ensemble.py
+++ b/src/tabpfn/preprocessing/ensemble.py
@@ -178,7 +178,7 @@ class TabPFNEnsemblePreprocessor:
             c.preprocess_config.max_features_per_estimator for c in self.configs
         ]
 
-        importance_feature_orders: list[np.ndarray] | None = None
+        importance_feature_order: np.ndarray | None = None
         needs_subsampling = any(
             s < n_total_features for s in max_features_per_estimator
         )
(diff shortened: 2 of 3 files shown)


bejaeger (approved):
LGTM!

PR #1230, opened 2026-09-02
Move the number of estimators into the inference config

bejaeger:
`n_estimators="auto"` resolved to a hard-coded `DEFAULT_N_ESTIMATORS = 8`, so a checkpoint could not say how many estimators it wants to be run with. It now takes `InferenceConfig.N_ESTIMATORS` from the checkpoint. Same treatment `SOFTMAX_TEMPERATURE` got in #1220.

The field is `int | "auto"`, defaulting to `"auto"`, and means exactly what the `n_estimators` argument means:

- an int is used as given and **never** raised by feature-coverage scaling — a checkpoint asking for 4 gets 4, the same guarantee a user passing 4 gets
- `"auto"` means `DEFAULT_N_ESTIMATORS`, raised on wide tables for feature coverage

No existing checkpoint carries the key, so they all resolve to `"auto"` and behave exactly as before.

| call | resolved `n_estimators_` |
|---|---|
| checkpoint declares nothing, or `"auto"` | 8 |
| checkpoint declares 3 | 3 |
| `n_estimators=5`, checkpoint declares 3 | 5 |
| `inference_config={"N_ESTIMATORS": 6}`, checkpoint declares 3 | 6 |
| `n_estimators=5` and `N_ESTIMATORS=6` | raises |
| two checkpoints declaring 3 and 4, or `"auto"` and 4 | raises |
| those two plus `n_estimators=2` | 2 |

On a 20-feature table seeing 2 features per estimator: a checkpoint's `"auto"` scales 8 → 10, while a checkpoint's `3` stays 3 and warns, same as an explicit `n_estimators=3`.

## Two things to look at

**An explicit `n_estimators` is not written back into `inference_config_`**, unlike the temperature. `save_tabpfn_model` persists that config and finetuning defaults to `n_estimators_finetune=2`, so writing it back would make every finetuned checkpoint declare `N_ESTIMATORS=2` and cap everyone who loads it. The argument is applied where the count is consumed instead. Pinned by `test__classifier_n_estimators__explicit__stays_out_of_the_config`.

**#1220's helpers are generalized rather than copied**, since two fields now need the same "checkpoints may disagree, the user breaks the tie" logic: one `OVERRIDABLE_FIELDS` table, one agreement check, one resolver. That renames `equals_ignoring_softmax_temperature` to `equals_ignoring_overridable_fields` and rewords one softmax error to name the argument (`` `softmax_temperature` was given twice ``). Behavior is unchanged. Happy to split the table out if you'd rather keep #1220 byte-identical.

Two smaller notes: `"auto"` counts as naming nothing wherever it comes from, so a hand-built `InferenceConfig` (whose `N_ESTIMATORS` is `"auto"`) no longer collides with an explicit `n_estimators` — this field avoids the wart the temperature has. And the agreement check sorts by `str`, since a field can now mix ints with a string.

## Tests

Full suite green (2062 passed), including `test_consistency.py` reference predictions and both finetuning suites. New: 17 tests in `test_inference_config.py` — resolution order, both raises, both ways to resolve the ensemble case, `"auto"` vs a count, and that a checkpoint's int survives coverage scaling while its `"auto"` does not — plus a guard that every `OVERRIDABLE_FIELDS` entry names a real config field and a real constructor argument on both estimators.


src/tabpfn/base.py
```diff
@@ -455,51 +466,82 @@ def initialize_model_variables_helper(
...
+    overrides: dict[str, float | int | None] = {}
+    for field, (_, argument) in OVERRIDABLE_FIELDS.items():
+        from_config: float | int | None = None
+        if isinstance(user_config, InferenceConfig):
+            # A hand-built config replaces the checkpoint's wholesale, so it carries
+            # a value for every field, this one included.
+            from_config = getattr(user_config, field)
+        elif isinstance(user_config, dict) and field in user_config:
+            from_config = user_config[field]
+        if from_config == "auto":
```
klemens-floege, line 511: Shall we get rid of "auto" already? Otherwise put both cases in here pls

```
 - [P2] Explicit "auto" overrides cannot resolve checkpoint disagreement — /private/tmp/tabpfn-
    pr1230.X8EkLO/src/tabpfn/base.py:511. inference_config={"N_ESTIMATORS": "auto"} is normalized to “no
    override,” so an ensemble whose checkpoints declare 3 and 4 estimators still raises. This prevents
    users from deliberately selecting automatic feature-coverage scaling. Preserve the distinction between
    “not supplied” and explicitly supplied "auto"
```

src/tabpfn/base.py
```diff
@@ -455,51 +466,82 @@ def initialize_model_variables_helper(
...
+
+    The `n_estimators` argument wins; left at `"auto"` the count comes from the
+    checkpoint's `InferenceConfig.N_ESTIMATORS`, which may itself be `"auto"`. Both
+    mean the same thing, so the result is passed straight to
+    `scale_n_estimators_for_feature_coverage`: an int is used as given, `"auto"`
+    resolves to `DEFAULT_N_ESTIMATORS` and may be raised for feature coverage.
+    """
+    if estimator.n_estimators != "auto":
+        return estimator.n_estimators
+    return estimator.inference_config_.N_ESTIMATORS
```
klemens-floege, line 544: Codex found another weird edge case

```
 - [P3] Warning guidance becomes incorrect for checkpoint-defined counts — /private/tmp/tabpfn-
    pr1230.X8EkLO/src/tabpfn/base.py:544. When a checkpoint declares N_ESTIMATORS=3, the scaler loses that
    provenance. On a wide table it tells the user to pass n_estimators="auto", although that is already
    their setting. With auto_scale_n_estimators=False, it additionally claims the behavior equals
    n_estimators=8, while the resolved result remains 3. The warnings should be source-aware or use
    source-neutral guidance.
```

klemens-floege (commented):
All in all really good, I think we have 2 more edge cases to iron out :)

src/tabpfn/base.py
bejaeger: I didn't want to change the default for v3.
Hm...this sounds like an edge case that we and the user can live with? if someone uses multiple model_paths that have different n_estimator settings, they cannot use "auto". I think this is fine?

src/tabpfn/base.py
bejaeger: Adjusted the messages

klemens-floege (approved):
LGTM!

PR #1233, opened 2026-09-02
Plan safe batched prediction across heterogeneous preprocessing shapes

eliott-kalfon:
## Before / after

| Before | After |
|---|---|
| One global batch padded differing preprocessed widths. | Full post-preprocessing shapes form stable, padding-free groups. |
| Predictions could depend on batch peers. | Results match serial prediction and retain input order. |
| Fine-tuning used padded collation. | Fine-tuning behavior is unchanged. |

## Tests

- L4 spot: 12 passed (stock v3 classifier + regressor)
- CPU batched APIs: 31 passed
- Ruff: passed


jmkuebler (dismissed):
Nice, thanks a lot for fixing this @eliott-kalfon! the approach looks sane to me and testing should cover if we in the future violate this again. 

On the downside, this could make the batched run slower ofc. Would it make sense to add a `logger.debug` that states how many heterogeneous groups are launched -- so it is easy to debug in case one doesn't see speedups from batching?

eliott-kalfon:
> Nice, thanks a lot for fixing this @eliott-kalfon! the approach looks sane to me and testing should cover if we in the future violate this again.
> 
> On the downside, this could make the batched run slower ofc. Would it make sense to add a `logger.debug` that states how many heterogeneous groups are launched -- so it is easy to debug in case one doesn't see speedups from batching?

I agree, it will make the batched run slower, as it adds sequential operations. The logger debug does make a lot of sense. I am also thinking of a UserWarning. Will stick to the debug for now

jmkuebler:
@eliott-kalfon the CI fix is merged. Could you undo your changes to tests/test_torch_preprocessing/test_torch_quantile_transformer.py and then we are good to go

eliott-kalfon:
> @eliott-kalfon the CI fix is merged. Could you undo your changes to tests/test_torch_preprocessing/test_torch_quantile_transformer.py and then we are good to go

Yes, accepted incoming changes and resolved conflicts

PR #1236, opened 2026-09-02
Refuse datetime columns by default, or expand them with TRANSFORM_DATES

alanprior:
Supersedes #1225 and #1227, which this branch grew out of, as one PR. Follow-up to #1206.

## What and why

A genuine `datetime64` column next to any other dtype crashed validation with an opaque numpy error, and a string column that happened to parse as a date was silently reclassified. Date handling is now explicit: a date is recognized only by its dtype, refused by default with an error naming it, and expanded into calendar features on request. Nothing is inferred from content, and nothing is guessed at predict time.

## How it works

- `DateTransformer` (`preprocessing/datetimes.py`) runs on the raw DataFrame before sklearn validation. It is fitted once, kept as `date_transformer_`, and reapplied at predict.
- Off, the default: a `datetime64`, tz-aware or `period` column raises `TabPFNValidationError` naming its index and label. A `timedelta64` column becomes its length in seconds.
- On: each date column becomes `skrub.DatetimeEncoder` features (year, day of year, seconds since the epoch, circular month/day/weekday pairs, and the time of day when the column carries one). Expansion widens the frame, so the transformer reports `feature_names_out_` and shifts declared categorical indices through `output_indices()`; the throwaway estimators used for temperature calibration and threshold tuning get the shifted indices too.
- `feature_names_in_` and `n_features_in_` keep describing the caller's frame: `extract_input_shape` reads them off the raw input, and `check_input_shape_matches` checks predict input against them. Fine-tuning checks `X_val` against the training set's shape the same way.
- Refused rather than guessed at: a date declared in `categorical_features_indices`; at predict, a date where fit saw none, a fitted date position that no longer holds dates, a column in another timezone than at fit, or an array after an expanding fit. A column label in `categorical_features_indices` is refused as well, where it used to be ignored.

## How to use

`TabPFNRegressor(inference_config={"TRANSFORM_DATES": True})`, likewise for the classifier. Predict input has to be a DataFrame carrying the same columns as datetimes, in the same timezone as at fit.

## Also in this PR

- `get_embeddings` cleans categoricals from the fitted schema like every other predict path.
- The out-of-memory error reports the width the model actually ran on, and the regressor now reports sizes at all.
- `FeatureModality.DATE`, which nothing consumed, is removed.

## Not covered

- The fine-tuning estimators validate their input directly and do not run the transformer, so a datetime column has to be converted before fine-tuning. Documented in the flag's docstring and the module.
- Only DataFrame columns are inspected. A numpy `datetime64` array fails as before, and a pandas `category` dtype is not read as a categorical declaration, since detection runs on a numpy array by then. Both are follow-ups.
- A string column listed in `categorical_features_indices` with more than `MIN_CARDINALITY_FOR_TEXT` distinct values is still classified as text, as on main. Also a follow-up.

---

## Public API Changes

-   [ ] No Public API changes

Additive: `InferenceConfig.TRANSFORM_DATES`, the `date_transformer_` fitted attribute, and `tabpfn.preprocessing.datetimes.DateTransformer`.

Changed, in `tabpfn.validation`: `ensure_compatible_fit_inputs` returns `(X, y, target_name)` rather than also the feature names and count; `ensure_compatible_fit_inputs_sklearn` returns `(X, y)`; `ensure_compatible_predict_input_sklearn` no longer checks the input against `n_features_in_`/`feature_names_in_` (`check_input_shape_matches` does, on the raw input). New: `extract_input_shape`, `check_input_shape_matches`, `validate_categorical_features_indices`. Removed: `FeatureModality.DATE`.

---

## Checklist


src/tabpfn/preprocessing/datetimes.py
```diff
@@ -0,0 +1,375 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from tabpfn.constants import XType
+
+__all__ = ["DateTransformer"]
+
+
+@dataclasses.dataclass
+class FittedDateColumn:
```
bejaeger: nit: can be `_FittedDateColumn` to make extra clear it's not used outside of this file.

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -148,35 +93,22 @@ def _format_names_for_warning(names: list[str]) -> str:
...
     *,
     declared_cat_indices: Sequence[int] | None = None,
 ) -> None:
-    """Warn about detected dates, then about any remaining free-text columns.
+    """Warn about any free-text columns.
 
-    Called before `_demote_dates_for_now` acts on `DATE`, so a date column
-    isn't `TEXT` yet and needs no special-casing to avoid a double warning.
+    A converted date is numeric by the time detection runs (see
+    `datetimes.DateTransformer`), so it is never reported here.
```
bejaeger: nit: seems like an unnecessary comment

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -256,37 +188,24 @@ def _detect_feature_modality(
         s.dtype, pd.CategoricalDtype
```
bejaeger, line 185: let's remove the second `isinstance(
        s.dtype, pd.CategoricalDtype)` to avoid converting cats to text.

src/tabpfn/validation.py
```diff
@@ -17,23 +17,94 @@
...
+    """The `feature_names_in_` and `n_features_in_` a raw fit input implies.
+
+    Read off the raw input, before date conversion (`DateTransformer`) or value
+    validation run: both attributes have to describe what the caller passed, not
+    TabPFN's internal, possibly wider, view of it. Only the shape and the column
+    labels are read here, never the values, so an `X` still holding a genuine
+    `datetime64` column is fine, unlike for the rest of validation.
+
+    Sets nothing: the caller assigns the two values it gets back, so what an
+    estimator records about its input is written where it can be read.
```
bejaeger: nit: can be removed

src/tabpfn/validation.py
```diff
@@ -17,23 +17,94 @@
...
+
+def check_input_shape_matches(X: XType, *, estimator: BaseEstimator) -> None:
+    """Check a predict input against the `feature_names_in_`/`n_features_in_` of fit.
+
+    The counterpart of `extract_input_shape`, and read-only likewise: it compares
+    and raises, writing nothing back onto `estimator`. Call it on the raw input,
+    before date conversion, for the same reason.
+
+    The labels are checked before the count, so a frame that is both renamed and
+    narrowed reports the name mismatch, as sklearn's own check would.
```
bejaeger: nit: make more concise

bejaeger (dismissed):
Great! Just some nits (and one fix for categoricals becoming strings) but then let's get this in! Two more comments:
- Let's do a round of "make docstrings more concise"
- We could think about setting an explicit modality for the transformed datetime column. In a follow-up, we could think about assigning them explicitly to NUMERICAL or CATEGORICAL (e.g. for circular features).

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -241,6 +158,10 @@ def _detect_feature_modality(
         # constant numeric column when predict sees an unseen string value.
         return FeatureModality.CONSTANT
 
+    if isinstance(s.dtype, pd.CategoricalDtype):
+        # An explicit category dtype settles it, whatever the values or their count.
+        return FeatureModality.CATEGORICAL
```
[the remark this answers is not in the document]
alanprior: Reverting

changelog/1236.fixed.4.md
```diff
@@ -0,0 +1 @@
+A string column listed in `categorical_features_indices` is now read as categorical whatever its number of distinct values. Above `MIN_CARDINALITY_FOR_TEXT` it used to be classified as text, silently, since declaring it also suppressed the free-text warning, and so was neither cast to category nor passed to the model as categorical.
```
bejaeger: is this still true for this PR or do we do it in a follow-up?

bejaeger (dismissed):
Just one comment, otherwise: let's get this in!

changelog/1236.fixed.4.md
alanprior: As discussed, it's there but let's revert this and leave for the next PR

alanprior:
@bejaeger DONE! Thanks!!! 🤯 📆

PR #1240, opened 2026-09-04
Add TRANSFORM_TEXT: expand string columns into numeric features via skrub's StringEncoder

alanprior:
### What & why
The text half of #1211, without any change to modality detection. With `inference_config={"TRANSFORM_TEXT": True}` a text column is expanded into tf-idf features over its character n-grams, reduced by a truncated SVD (`skrub.StringEncoder`). Off by default, the estimators behave exactly as on `main`; the only visible change is that the free-text warning now points at the flag instead of saying the package has no text handling.

### How it works
- `preprocessing/text.py::TextTransformer` mirrors `DateTransformer` and runs right after it on the raw DataFrame, before validation flattens dtypes away. A column is text when it has a string dtype (pandas `string` in any storage, or a pyarrow string), more than `MIN_CARDINALITY_FOR_TEXT` distinct values, does not parse as numbers, and is not declared categorical. `object` and `category` columns are never expanded.
- `base.expand_dates_and_text` runs both transformers and returns the expanded input, the fitted transformers, the output column labels and the declared categorical positions in the expanded input, so neither estimator needs to know which transformer ran last.
- The estimators keep `text_transformer_` and a new `categorical_features_indices_` (declared indices in expanded coordinates); the tuning estimators, which fit on the already-expanded array, read the latter.
- At predict, each expanded position is cast to `string` and re-encoded with the encoder fitted on the training column, so the width holds. A numeric column at such a position is refused; an all-missing one is not, since it holds no numbers. An array after an expanding fit is refused.
- At predict, a string is encoded by the character n-grams it shares with the training column, so an unseen sentence in the same language lands near its neighbours. A missing value becomes the empty string, which shares none, and so an all-zero row, like an ID or a string in another script with no n-gram in common; the model never sees a missing value in an expanded column.

### How to use
```python
TabPFNClassifier(inference_config={"TRANSFORM_TEXT": True, "TEXT_N_COMPONENTS": 30})
```
Columns must carry pandas' `string` dtype (the default from pandas 3.0; `.astype("string")` on pandas 2).

### Design decisions
- No detection change: `FeatureModality.TEXT`, the `category` dtype handling and the declared-categorical string fix from #1211 are left for a separate PR, as asked.
- Default off, mirroring `TRANSFORM_DATES`.
- `TEXT_N_COMPONENTS` defaults to 30 because that is `StringEncoder`'s own default.
- The encoder is seeded on its own (`random_state=0`): the features a column turns into are a property of the data, not of the estimator's seed.
- The fine-tuning estimators do not run it, like the date transformer.

### Deferred follow-ups
- Text vs category detection (`category` dtype, declared string columns above the cutoff, ordered categoricals) in a separate PR.
- Expanding `object` columns.
- pyarrow is not a dev dependency, so the three pyarrow tests skip in CI unless it is added.


src/tabpfn/inference_config.py
```diff
@@ -151,6 +149,19 @@ class InferenceConfig:
...
 
+    TRANSFORM_TEXT: bool = False
+    """Whether a text column, a pandas `string` or pyarrow string column with more
+    than `MIN_CARDINALITY_FOR_TEXT` distinct values, is expanded into
+    `TEXT_N_COMPONENTS` numeric features via `skrub.StringEncoder` (tf-idf over
+    character n-grams, truncated SVD). Off, such a column is ordinal-encoded as a
+    high-cardinality category and `fit` warns about it. An `object` column is
+    never expanded. Not run by the fine-tuning estimators."""
+
+    TEXT_N_COMPONENTS: int = 30
```
bejaeger, line 160: nit: would still be more a fan of 32.

bejaeger (dismissed):
Looks great! Thanks a lot

src/tabpfn/inference_config.py
alanarazi7: Let us evaluate this thoroughly as I don't have results for the possible performance shift.

PR #1243, opened 2026-09-07
Default fitted-model loading to device="auto"

### LINKED ISSUE #1221: TabPFN inference extremely slow on large spatial dataset — RTX 4090
Hi, I am using TabPFNRegressor to generate maps from satellite data. The model is trained on 100,000 rows (12 features) and I need to run inference on 4,344 scenes, each with ~56,000 pixels.

Setup:

TabPFNRegressor(device="cuda", fit_mode="fit_with_cache")
PyTorch 2.11.0+cu128, RTX 4090 24GB, CUDA 12.8
Batch size: 1,000 (as recommended in the docs)
Model saved/loaded with save_fitted_tabpfn_model / load_fitted_tabpfn_model
tabpfn version: 8.4.0

Problem:
Inference is extremely slow — roughly 5 days for ~9% of scenes (400/4,344). Each scene has ~56,000 valid pixels, split into ~57 batches of 1,000. That means ~247,000 predict() calls total.

Questions:

Does fit_with_cache actually persist through save_fitted_tabpfn_model / load_fitted_tabpfn_model? Or does the KV cache need to be rebuilt after loading?
Is there a recommended approach for running TabPFN inference on a large number of small batches efficiently?
Would fit_mode="batched" be more appropriate for this use case?

Thanks

alanprior:
`load_fitted_tabpfn_model` and the `load_from_fit_state` class methods defaulted to `device="cpu"`, so a GPU-fitted estimator silently reloaded onto CPU and predicted far slower with nothing to signal it. Assigning to `.device` afterwards cannot fix that, since predict reads `devices_`. Reported in #1221.

The archive does not record the device a model was fitted on, so `"auto"` matches what the constructors already default to.

- `load_fitted_tabpfn_model` and both `load_from_fit_state` methods default to `device="auto"`, and the annotation widens to `DevicesSpecification`.
- The loader passes the spec through untouched. The old `str(device)` would have turned a sequence into the literal `"['cuda:0', 'cuda:1']"`.
- `save_fitted_tabpfn_model` renders the device spec for JSON. This also fixes saving an estimator constructed with a `torch.device`, which raises `TypeError: Object of type device is not JSON serializable` on `main` today, independently of this PR.
- The two tuning-config tests pin to `"cpu"`: they fit on CPU and assert a same-device round-trip, which the new default would otherwise turn into a cross-device one.
- `examples/save_and_load_model.py` fits with `device="auto"` rather than a hardcoded `"cuda"`, so it can run without a GPU. Nothing in the test suite or CI executes `examples/*.py`.

## Issue
Fixes #1221

## Motivation and Context

Loading a fitted model was the only entry point whose device default disagreed with the constructors', and the disagreement was silent: the CPU sample guard is reachable only from fit paths, so a large context predicting on CPU raises neither an error nor a warning. The reporter in #1221 diagnosed the device himself and applied the intuitive repair; the library did not honour it.

---

## Public API Changes

-   [ ] No Public API changes

`load_fitted_tabpfn_model`, `TabPFNRegressor.load_from_fit_state` and `TabPFNClassifier.load_from_fit_state` now default to `device="auto"` instead of `device="cpu"`. Code relying on the old default to keep a GPU free must pass `device="cpu"` explicitly.

---

## How Has This Been Tested?

`tests/test_save_load_fitted_model.py`, 28 passing with MPS both available and excluded, plus running `examples/save_and_load_model.py` end to end.

---

## Checklist

---


changelog/1243.changed.md
```diff
@@ -0,0 +1 @@
+`load_fitted_tabpfn_model` and the `load_from_fit_state` class methods now default to `device="auto"`, matching the estimator constructors, instead of `device="cpu"`. The archive does not record the device a model was fitted on, so a GPU-fitted estimator silently reloaded onto CPU and predicted far slower with nothing to signal it — and assigning to `.device` afterwards could not fix it, since predict reads `devices_`. Code that relied on the old default to keep a GPU free must now pass `device="cpu"` explicitly.
```
bejaeger: Let's make this and the other changelog more concise

bejaeger (dismissed):
LGTM! I think the changelog could be made more concise.

PR #1245, opened 2026-09-07
Take declared categoricals at face value: never TEXT, and fold in pandas category columns

bejaeger:
## What & why

A string column the user declared categorical could still end up as `FeatureModality.TEXT`. Two ways in:

- **`categorical_features_indices`**: above `MIN_CARDINALITY_FOR_TEXT` (30) distinct values, the declaration only silenced the free-text warning. The column was ordinal-encoded like text and never flagged categorical to the model.
- **pandas `category` dtype**: not a declaration at all. The dtype is lost when the frame becomes a numpy array before `detect_feature_modalities` runs, so a `category` column was read like any `object` column: category up to 30 distinct values, text above.

Now a declared string column is a categorical, whatever its cardinality.

## How

- `include_category_dtype_columns` (`base.py`): at fit, the positions of every `category` column are merged into the declared indices, before any date/text expansion moves columns. The merged list drives everything downstream and is what `categorical_features_indices_` exposes.
- `_detect_feature_modality`: a declared string-like column is `CATEGORICAL` whatever its cardinality. `MIN_CARDINALITY_FOR_TEXT` only sorts undeclared string columns into category or text.
- The `CategoricalDtype` check in `_detect_feature_modality` is removed: it could never fire in the estimator path (`X` is a numpy array by then), and the dtype's intent now travels in the declared indices.
- `_warn_on_text` loses its `declared_cat_indices` exemption: a declared column can no longer be text, so there is nothing to exempt. The message now says a declaration makes the column a categorical rather than just silencing the warning, and mentions the `category` dtype.
- Docstrings for `categorical_features_indices`, `categorical_features_indices_` and `MIN_CARDINALITY_FOR_TEXT` updated.

Unchanged: a declared *numeric* column is still only categorical up to `MAX_UNIQUE_FOR_CATEGORICAL_FEATURES`, as before. An undeclared high-cardinality string column is still `TEXT` and still warns.

## Is `TEXT` still needed?

Checked, and kept. `TEXT` is produced only in `modality_detection.py` and consumed in two places: the free-text warning, and by exclusion from `indices_for(CATEGORICAL)`, so an ID or free-text column is ordinal-encoded but not presented to the model as a categorical. That is the string analogue of the numeric rule that caps declared categoricals at 30 distinct values. After this PR it applies only to *undeclared* columns, which is where a cardinality heuristic belongs. Dropping it would mean either flagging every high-cardinality string column as categorical to the model, or losing the warning; neither follows from "a declared categorical is a categorical".

Empirically the label makes a small difference: on synthetic data with a 40- or 100-way string column, 4 seeds, `n_estimators=4`, TEXT vs CATEGORICAL differ by about ±0.01 AUC / R².

## Public API changes

- `categorical_features_indices_` now also contains the positions of pandas `category` columns.
- Behavior change: a declared string column with more than 30 distinct values is now presented to the model as categorical (before: ordinal-encoded, not flagged). A `category` dtype column with more than 30 distinct values no longer warns.

## How has this been tested?

- `test_modality_detection.py`: reported string column is categorical at any cardinality (unit); declared columns (index and `category` dtype) are `CATEGORICAL` and do not warn, undeclared ones are `TEXT` and do (detection-level); estimator-level test that a `category` column lands in `categorical_features_indices_` and the schema, merges with user-given indices, and that only the undeclared column is named in the warning.
- `test_text.py`: the `category` case of the dtype test now asserts the column is a declared categorical rather than a warned-about text column.
- `tests/test_preprocessing`, `test_classifier_interface.py`, `test_regressor_interface.py`, `test_finetuning_*.py`: all pass locally. `TestExpansion::test__text_column__becomes_numeric_features` fails identically on `main` (a skrub SVD warning under `simplefilter("error")`), unrelated.

## CI note

The failing `highest` jobs are an upstream break, not this PR: joblib 1.6.0 (released 2026-08-31) dropped its vendored `cloudpickle`, which skrub 0.10.0 imports. skrub 0.10.1 (2026-09-01) fixes it, but `[tool.uv] exclude-newer = "7 days"` keeps it out of the resolution until 2026-09-08. Rerunning CI then should go green; `main` would fail the same way on a fresh run today.

## Checklist

- [x] Documentation has been updated (docstrings).
- [x] A changelog entry has been added (see `changelog/README.md`).


src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -182,14 +175,14 @@ def _detect_feature_modality(
             return FeatureModality.CATEGORICAL
         return FeatureModality.NUMERICAL
 
-    is_string_like = pd.api.types.is_string_dtype(s.dtype) or isinstance(
-        s.dtype, pd.CategoricalDtype
-    )
-    if is_string_like:
-        if n_unique <= min_cardinality_for_text:
+    # A pandas `category` column never arrives here as such: `X` is a numpy array
```
alanprior, line 182: @bejaeger could we perhaps add such assertion, so this behavior is enforced and maintained? it feels like removing comments / changing the flow could kill this behavior.

src/tabpfn/base.py
```diff
@@ -415,6 +416,30 @@ def create_inference_engine(  # noqa: PLR0913
...
+    `categorical_features_indices`, so the two are merged here, before any column
+    moves, and the merged list drives everything downstream: the declared
+    categoricals are taken at face value, whatever their cardinality.
+
+    Returns:
+        The sorted union of both, or `None` when neither names a column.
+    """
+    declared = set(categorical_features_indices or ())
+    if isinstance(X, pd.DataFrame):
+        declared.update(
```
alanprior: @bejaeger nit: if possible, I would define an intermediate variable names `typed` or similar. It's anyway 4 lines, I think it's a bit nicer

src/tabpfn/classifier.py
```diff
@@ -208,8 +209,9 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
     """The transformer that expanded every text column before validation."""
 
     categorical_features_indices_: list[int] | None
-    """`categorical_features_indices` as positions in the validated input, where
-    an expanded date or text column has moved everything after it down."""
+    """`categorical_features_indices`, plus every pandas `category` column of the
+    fit input, as positions in the validated input, where an expanded date or text
+    column has moved everything after it down."""
```
alanprior: @bejaeger I know this is old and probably mine but I'm unsure what "where an expanded date or text column has moved everything after it down." means

src/tabpfn/classifier.py
```diff
@@ -305,8 +307,10 @@ def __init__(  # noqa: PLR0913
             categorical_features_indices:
                 The indices of the columns that are suggested to be treated as
                 categorical. If `None`, the model will infer the categorical columns.
-                If provided, we might ignore some of the suggestion to better fit the
-                data seen during pre-training.
+                A column with pandas' `category` dtype counts as listed here. A
+                string column declared this way is read as categorical whatever
+                its cardinality, never as text; for a numeric one, we might ignore
+                the suggestion to better fit the data seen during pre-training.
```
alanprior, line 313: @bejaeger I think we should stop ignoring the numerical as well. But we can make this a TODO / discuss

alanprior: This would be a bigger change because it will kill some of the current config params, like the MIN_TO_INFER or so

src/tabpfn/regressor.py
```diff
@@ -900,10 +904,13 @@ def _initialize_dataset_preprocessing(
         self.feature_names_in_, self.n_features_in_ = extract_input_shape(X)
 
         validate_categorical_features_indices(self.categorical_features_indices)
+        categorical_indices = include_category_dtype_columns(
```
alanprior: @bejaeger nit: could we unify 

validate_categorical_features_indices and include_category_dtype_columns?

it will alleviate the code duplication across classifier/regressor and who knows who else needs

tests/test_preprocessing/test_modality_detection.py
```diff
@@ -277,12 +277,6 @@ def test__detected_categorical_without_reporting():
     assert result == FeatureModality.CATEGORICAL
 
 
-def test__detect_for_categorical_with_category_dtype():
```
alanprior, line 280: @bejaeger actually could be cool to put this back and see it raises an assertion - that the original function can't get any categorical values or so

alanprior (dismissed):
I like it. Most of my comments are style related. There's only one decision we could take now or later, about reported indices for numerical features

src/tabpfn/classifier.py
LeoGrin: Agree! Last time we chatted with Ben the plan was to move MIN_UNIQUE_FOR_NUMERICAL_FEATURES to 0 for the next version.

alanprior: Good! Longer run I will consider removing this altogether, but this might break older versions.

PR #1246, opened 2026-09-07
Accept numpy arrays of strings as input

alanprior:
## What & why

`fix_dtypes` refused a numpy array of strings (dtype kind `U`) with a bare `ValueError` ("String dtypes are not supported"), at fit and at predict, while the same strings in an `object` array or a DataFrame were accepted and encoded. A model fitted on a DataFrame refused a numpy string array at predict but accepted the same array cast to `object`. The guard dates from the original V2 code; nothing depended on it, since the modality detector had already labelled the columns by the time it fired.

## How

- A unicode array is wrapped in a DataFrame the way an `object` array already is; the existing dtype conversion and ordinal encoding take it from there.
- The dtype-kind constants are named by what `fix_dtypes` does with them: `OBJECT_OR_STRING_DTYPE_KINDS = "OVU"` (object, void, unicode) takes the pandas introspection path; `BYTES_DTYPE_KINDS = "Sa"` is refused, now with a message that says "Byte string dtypes". Each letter is spelled out where the constants are defined.
- Scope is fixed-width unicode arrays only. Byte string arrays and numpy 2's `StringDType` are left as they were.

## Public API Changes

- [ ] No Public API changes
- [x] Yes, Public API changes (Details below)

A numpy array of strings is accepted by `fit` and `predict` where it was refused before. It behaves exactly like the same strings passed in a DataFrame.

## How Has This Been Tested?

`tests/test_preprocessing/test_data_cleaning.py`: `fix_dtypes` on a unicode array equals `fix_dtypes` on its `object` twin; a bytes array is refused; both estimators fit and predict on a unicode array and predict the same as on the `object` array, whether fitted on the array or on a DataFrame. The full files `test_data_cleaning`, `test_sklearn_estimator_checks`, `test_input_immutability`, `test_estimators`, `test_classifier_interface` and `test_regressor_interface` pass locally.

## Checklist

### DIFF
--- a/changelog/1246.fixed.md
+++ b/changelog/1246.fixed.md
@@ -1,0 +1,1 @@
+numpy arrays of strings are accepted as input at fit and predict, like `object` arrays and DataFrames.
--- a/src/tabpfn/preprocessing/clean.py
+++ b/src/tabpfn/preprocessing/clean.py
@@ -40,9 +40,10 @@ NUMERIC_DTYPE_KINDS = "?bBiufm"
 # converts those through its own units rather than numpy's raw integers.
 FAST_CONVERTIBLE_DTYPE_KINDS = "?bBiuf"
 OBJECT_DTYPE_KINDS = "OV"
-# numpy's fixed-width unicode strings; pandas reads them as string columns.
-UNICODE_DTYPE_KINDS = "U"
-BYTES_DTYPE_KINDS = "Sa"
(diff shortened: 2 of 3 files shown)


LeoGrin (approved):
Thanks, LGTM!

nit: we could maybe also support the new numpy StringDType? but not necesseraly here

PR #1250, opened 2026-09-08
Remove the dead cache_trainset_representation flag

jmkuebler:
## Issue

RES-2743 (base of #1219).

## Motivation and Context

`cache_trainset_representation` has been dead since `InferenceEngineCacheKV` was removed in #1057. Every architecture's `get_architecture` accepts it and immediately `del`s it; the KV cache is owned by the inference engine and selected at call time through `forward()`. The flag survived only because the `ArchitectureModule` protocol still declared it, which forced every implementation, every loader, and every test call site to carry it.

This removes it end to end:

- `ArchitectureModule.get_architecture(config)` no longer takes the kwarg; the four architectures drop the parameter and their "accepted for interface compatibility but ignored" notes.
- `load_model_criterion_config`, `load_model` and `_build_model` lose the parameter.
- The opt-in built-model cache (`TABPFN_MODEL_CACHE_SIZE`) no longer gates on the flag. The gate assumed a `fit_with_cache` build mutates the module, which it does not, so that fit mode now uses the cache like every other one.
- `initialize_tabpfn_model` loses its `fit_mode` argument, whose only job was computing the flag.
- Tests: every call site, the three fake `ArchitectureModule`s, the "mutating build is never cached" test, and the `load_model` signature tripwire.

#1219 is stacked on top and shrinks to keying the cache on device and dtype.

fomo-fitting side: PriorLabs/fomo-fitting#3776 drops the parameter from every architecture there. It is a draft until this merges and the tabpfn pin is bumped.

---

## Public API Changes

-   [ ] No Public API changes

- `ArchitectureModule.get_architecture` no longer accepts `cache_trainset_representation`. External architectures implementing the protocol must drop the parameter.
- `tabpfn.model_loading.load_model` and `load_model_criterion_config` no longer accept `cache_trainset_representation`.
- `tabpfn.base.initialize_tabpfn_model` no longer accepts `fit_mode`.
- With `TABPFN_MODEL_CACHE_SIZE` set, `fit_mode="fit_with_cache"` now uses the built-model cache. The default (cache off) is unchanged.

---

## How Has This Been Tested?

- ruff check / format clean.
- Cache off (shipping default): `tests/test_classifier_interface.py` and `tests/test_regressor_interface.py` pass on cpu and cuda; the touched unit files (`test_model_cache`, `test_model_loading`, `test_config`, `test_inference_config`, `test_inference`, `test_architectures/`, `test_save_load_fitted_model`) pass.
- Cache on, the interface suites fail with dtype/device collisions, as they did before this change (see #1219, which fixes that and is stacked on this PR).

---

## Checklist

### DIFF
--- a/changelog/1250.breaking.md
+++ b/changelog/1250.breaking.md
@@ -1,0 +1,1 @@
+Remove the dead `cache_trainset_representation` argument from `ArchitectureModule.get_architecture`, `load_model` and `load_model_criterion_config`, and the `fit_mode` argument from `initialize_tabpfn_model`. Every architecture ignored the flag since `InferenceEngineCacheKV` was removed; external architectures implementing the protocol must drop the parameter.
--- a/changelog/1250.changed.md
+++ b/changelog/1250.changed.md
@@ -1,0 +1,1 @@
+`fit_mode="fit_with_cache"` now uses the opt-in built-model cache (`TABPFN_MODEL_CACHE_SIZE`), where it previously rebuilt the model on every fit.
--- a/src/tabpfn/architectures/interface.py
+++ b/src/tabpfn/architectures/interface.py
@@ -143,20 +143,13 @@ class ArchitectureModule(Protocol):
         """
         ...
 
-    def get_architecture(
-        self,
-        config: ArchitectureConfig,
-        *,
-        cache_trainset_representation: bool,
-    ) -> Architecture:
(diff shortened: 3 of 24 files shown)


jmkuebler:
cc @priorphil  as you authored #1057  any reason to keep `cache_trainset_representation`?

oscarkey (dismissed):
sweeeet

PR #1253, opened 2026-09-08
Add majority_downsample row-subsampling mode via SAMPLE_SUBSAMPLING_METHOD

psinger-prior:
## Summary

Adds `InferenceConfig.SAMPLE_SUBSAMPLING_METHOD`, which makes the per-estimator row-subsampling rule explicit, and a new `"majority_downsample"` mode for it. The new mode keeps every row outside the single most frequent target value in every estimator's context and fills the rest of the `SUBSAMPLE_SAMPLES` budget from that majority group. On heavily imbalanced classification this puts the whole minority class in every context; on zero-inflated regression it downsamples only the zero spike. The default `"auto"` reproduces the previous behavior exactly (stratified for classification, balanced for regression).

## What changed

```diff
 TabPFNEnsemblePreprocessor.__init__
   _get_subsample_feature_indices
+  _resolve_sample_subsampling_method      # "auto" -> stratified | balanced
   _get_subsample_indices_for_estimators
-    if y given: _subsample_rows_stratified else _subsample_rows_balanced
+    _subsample_rows_by_method
+      balanced            -> _subsample_rows_balanced
+      stratified          -> _subsample_rows_stratified
+      majority_downsample -> _subsample_rows_majority_downsample
```

- New config option `SAMPLE_SUBSAMPLING_METHOD` with values `auto`, `balanced`, `stratified`, `majority_downsample`, plus a matching `SampleSubsamplingMethod` enum exported from `tabpfn.preprocessing`.
- New sampler: groups rows by exact target value, keeps all non-majority groups whole, and draws the majority group through the existing round-robin pool so majority rows are covered evenly across estimators. Non-majority rows are identical across estimators; the majority subset differs per estimator.
- Guardrails: no unique most frequent value (balanced classes, or a continuous target with no repeats) warns and falls back to the `auto` behavior. A budget that cannot hold all non-majority rows plus one majority row raises. Classification warns if the downsampled majority ends up smaller than another class.
- `stratified` is rejected for regressors; `majority_downsample` works for both tasks.
- Resolution happens only when `SUBSAMPLE_SAMPLES` is an int or float, so the method is ignored when subsampling is off or explicit index lists are given.
- Tensor targets from the differentiable-input path are detached once in the dispatch instead of at call sites.

## Results

Laptop-scale runs on MPS, 4 estimators, 10k rows of context per estimator, two seeds each (shown as seed 0 / seed 1), same pool and budget for both arms. Default TabPFN settings otherwise.

**Zero-inflated regression** (freMTPL2 loss cost, 3.7% nonzero, 100k-row pool, 20k test rows). Context holds ~370 nonzero rows with `auto` and all ~3,700 with `majority_downsample`.

| arm | gini | gini in exposure deciles | Tweedie skill (mean-rebalanced) |
|---|---|---|---|
| auto | 0.28 / 0.38 | 0.20 / 0.29 | 0.03 / -3.99 |
| majority_downsample | 0.45 / 0.59 | 0.28 / 0.40 | 0.10 / 0.14 |

**Imbalanced classification** (credit card fraud, 0.17% positive, 200k-row pool, 50k test rows). Context holds 18 positives with `auto` and all ~345 with `majority_downsample`.

| arm | ROC AUC | average precision | log loss |
|---|---|---|---|
| auto | 0.970 / 0.966 | 0.69 / 0.78 | 0.0038 / 0.0037 |
| majority_downsample | 0.982 / 0.977 | 0.77 / 0.82 | 0.0063 / 0.0054 |

Ranking improves consistently in both settings. Calibration gets worse before correction because the model sees a shifted prior: regression means come out 3 to 7 times too high and fraud probabilities about 3 times too high. Rescaling regression predictions to the training mean, or multiplying class probabilities by the original-to-context class ratio, restores the level. With that correction applied, `majority_downsample` beat `auto` on every metric in these runs (fraud log loss 0.0032 vs 0.0038, Brier 0.00056 vs 0.00068).

## Review guide

- Start with `_compute_majority_downsample_group_counts` and `_subsample_rows_majority_downsample` in `src/tabpfn/preprocessing/ensemble.py`. That is the whole new algorithm; everything else is plumbing.
- The `SAMPLE_SUBSAMPLING_METHOD` docstring in `src/tabpfn/inference_config.py` is the user-facing contract, including the prior-shift caveat.
- Mechanical: the enum, the export, and the two estimator call sites that forward the config value.
- Design choice: only the single largest group is downsampled. A smallest-to-largest water-filling variant was considered and dropped for being harder to explain; the simple rule matches the name.
- Design choice: ties and no-repeat targets fall back with a warning rather than raising, since a fixed config is often applied across datasets where the mode is simply not applicable.

## Testing

- New unit tests cover the count allocation (binary, multiclass, ties, budget edge cases, a randomized invariant that counts never exceed group sizes), the sampler (no duplicate rows, even majority coverage, string labels, zero-inflated targets, fast path with 200k distinct values), the resolver, dispatch errors, torch target detaching, and end-to-end classifier and regressor fits.
- `tests/test_preprocessing/test_ensemble.py`, `tests/test_finetuning_classifier.py`, and `tests/test_finetuning_regressor.py` pass locally. Ruff check and format are clean.
- The benchmark numbers above come from ad hoc scripts outside the repo and are not part of CI.

## Follow-ups

- Prior correction inside TabPFN: the mode shifts the target prior, and the results above show the correction is what makes it a strict improvement rather than a ranking-for-calibration trade. For classification this is a multiply-and-renormalize on the averaged probabilities next to `balance_probabilities`; for regression it is a rescaling of the predicted mean. Left for a separate PR so this one stays about sampling.


src/tabpfn/preprocessing/ensemble.py
```diff
@@ -207,9 +215,14 @@ def __init__(  # noqa: PLR0913
             cat_indices = (
                 self.feature_schema.indices_for(FeatureModality.CATEGORICAL) or None
             )
+            y_for_importance = (
+                y_train
+                if isinstance(y_train, np.ndarray)
+                else y_train.detach().cpu().numpy()
```
LeoGrin: Astra says this crashes the differentiable classifier fitting:

> P2] bfloat16 targets crash differentiable classifier fitting — [ensemble.py:728(https://github.com/PriorLabs/TabPFN/pull/1253/files). With numeric SUBSAMPLE_SAMPLES, default "auto" now calls .numpy() on bfloat16 targets, raising TypeError: Got unsupported ScalarType BFloat16. Confirmed with actual fits; the previous balanced path succeeds. Cast the detached metadata view to float32 before conversion.

LeoGrin (dismissed):
LGTM thanks! One very small bug I think, and a few nits:

- Wondering if we could name `majority_downsample` differently for classification and regression, and make the doc about each mode specific to the relevant task. Maybe `majority_downsample` for classif and `most_frequent_value_downsample` for regression? I understand the code is basically shared in both cases so might be a bit confusing, but from the outside I feel like `majority_downsample` might be a bit confusing for regression. But not a strong opinion and very nit, feel free to ignore.

- > Prior correction inside TabPFN: the mode shifts the target prior, and the results above show the correction is what makes it a strict improvement rather than a ranking-for-calibration trade. For classification this is a multiply-and-renormalize on the averaged probabilities next to balance_probabilities; for regression it is a rescaling of the predicted mean. Left for a separate PR so this one stays about sampling.

Happy to have this as a follow up, but I think steady state we should have the correction on by default right? WDYT?

src/tabpfn/classifier.py
```diff
@@ -1281,6 +1285,11 @@ def fit_with_differentiable_input(self, X: torch.Tensor, y: torch.Tensor) -> Sel
             ),
             constant_feature_count=self.inference_config_.FEATURE_SUBSAMPLING_CONSTANT_FEATURE_COUNT,
             subsample_samples=self.inference_config_.SUBSAMPLE_SAMPLES,
+            sample_subsampling_method=SampleSubsamplingMethod(
+                self.inference_config_.SAMPLE_SUBSAMPLING_METHOD
+            ),
+            y_train=y,
+            task_type=self.estimator_type,
```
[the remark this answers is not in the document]
psinger-prior: I think this was rather a bug before this change, differentiable classification should also to stratified sampling

src/tabpfn/preprocessing/ensemble.py
psinger-prior: good find, fixing

psinger-prior:
> Happy to have this as a follow up, but I think steady state we should have the correction on by default right? WDYT?

I would like to make it a separate setting, next to the balanced adjustment. And then we can turn it on by default for this one, yes.

For the naming I would stick to the current one to not have separate ones and the docstring should be clear.

PR #1255, opened 2026-09-09
Count distinct values for all columns of a numeric array at once in modality detection

Innixma:
`detect_feature_modalities` built a `pd.Series` and called `nunique` for every column. When the array is numeric or bool, which it is whenever the input has no object columns, only the distinct-value count decides a column's modality, so it is now counted for all columns at once: sort each column, count adjacent unequal pairs, count NaN once. `-0.0` equals `0.0` and `inf` equals `inf`, as under `nunique(dropna=False)`.

The per-column early exit is kept: the first 1024 rows are counted for every column, columns that already clear every threshold keep that count (it lands in the same bucket as the full count), and only the remaining columns are counted in full, so tall data does not pay for a full sort of high-cardinality columns. Object arrays go through the per-column path as before, and the numeric decision (constant / categorical / numerical) moved into one shared helper used by both paths.

On a 22k-column float frame with a few hundred rows this takes modality detection from about 0.4 s to 0.02 s and roughly halves `fit`. A randomized test compares the resulting schema with the per-column decisions on float, float32, int and bool arrays with NaN, signed zeros, infinities, all-missing and constant columns, row counts on both sides of the prefix, random thresholds and random declared-categorical indices; a second test pins the count against `nunique(dropna=False)`.

Two smaller items in the same spirit:

- `_is_numeric_pandas_series` walked every value of an object column in Python (or coerced the whole column) to decide whether it is numeric; a frame with a single non-numeric column arrives as one object array, so every numeric column paid for that. `pd.api.types.infer_dtype` now settles the common case in C: a kind whose non-missing values are all numbers (`integer`, `floating`, `mixed-integer-float`, `boolean`, `decimal`, `empty`) makes the column numeric; `string` and `mixed` kinds still take the existing path, since a spelled-out number counts. A parametrized test checks the shortcut against the value walk on object columns of floats, ints, bools, decimals, bytes, spelled numbers, words, mixed values, complex numbers and timestamps.
- `coerce_nullable_dtypes_to_numpy` evaluated two pandas dtype predicates per column; they now run once per distinct dtype and the columns are selected by membership. Same columns, same order, with a test over eleven dtypes.

### DIFF
--- a/src/tabpfn/preprocessing/modality_detection.py
+++ b/src/tabpfn/preprocessing/modality_detection.py
@@ -7,8 +7,8 @@ from __future__ import annotations
 import math
 import warnings
 from collections.abc import Sequence
-from typing import TYPE_CHECKING
 
+import numpy as np
 import pandas as pd
 
 from tabpfn.errors import TabPFNUserError
@@ -20,9 +20,6 @@ from tabpfn.preprocessing.datamodel import (
     FeatureSchema,
     build_input_feature_names,
 )
-
-if TYPE_CHECKING:
-    import numpy as np
 
 _EARLY_EXIT_PREFIX_ROWS = 1024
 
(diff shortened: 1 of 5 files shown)


jmkuebler:
@arthur-priorlabs since you worked quite a bit on preprocessing?

src/tabpfn/preprocessing/modality_detection.py
```diff
@@ -192,9 +197,115 @@ def _detect_feature_modality(
...
+) -> np.ndarray | None:
+    """Distinct values per column of a numeric or bool array, NaN counted as a value.
+
+    `None` for anything else (an object array is parsed column by column). Mirrors
+    the per-column early exit: a column whose first `_EARLY_EXIT_PREFIX_ROWS` rows
+    already hold `decided_at` distinct values keeps that prefix count, which lands
+    in the same bucket as the full count; only the other columns are counted in
+    full.
+    """
+    if not isinstance(X, np.ndarray) or X.ndim != 2 or X.dtype.kind not in "biuf":
```
arthur-priorlabs, line 261: Why are we bothering to check the number of distinct values in a bool array?

arthur-priorlabs (dismissed):
LGTM, thanks!

src/tabpfn/preprocessing/modality_detection.py
Innixma: (Claude-written) The count is what decides CONSTANT vs CATEGORICAL for a bool column (1 vs 2 distinct values), the same as `nunique` did on main, so it stays. You are right that sorting to get it is silly, and it was even slower than sorting the same column as float. e6e845c counts bool columns with `any`/`all` instead: 100k x 50 bools go from 49 ms to 2 ms, with a test against `nunique`.

PR #1257, opened 2026-09-10
Build models from a checkpoint without the random parameter initialisation

Innixma:
Stacked on #1255.

`_build_model` constructs the architecture and then loads the checkpoint with a strict `load_state_dict`, which overwrites every parameter and persistent buffer. The construction still ran every layer's `reset_parameters`, drawing random weights only to discard them: about a quarter of a second per build on a full-size checkpoint, paid at every `fit` and, under bagging, at every child.

The architecture is now built inside a small context manager that turns the `torch.nn.init` functions into no-ops for the duration and restores them afterwards, so the tensors are allocated but not filled before the load. The result is the same model: the architectures register no non-persistent buffers, and the tensors they initialise themselves are parameters, so all of them come from the checkpoint. A test loads small v2 and v3 checkpoints and checks that every named parameter and buffer equals the source model's exactly; a second test checks the patch is scoped to the block.

On an 83.5M-parameter checkpoint the build goes from 287 ms to 19 ms, most of the remainder being the state-dict copy; the reference-prediction consistency tests pass unchanged. One caveat for reviewers: the context manager patches module-level functions, so a model constructed concurrently in another thread of the same process during that window would also skip its initialisation. The library builds models on the calling thread, and the parallel paths use separate processes, so no such window exists in the code as it stands.

### DIFF
--- a/changelog/1257.changed.md
+++ b/changelog/1257.changed.md
@@ -1,0 +1,1 @@
+Building a model from a checkpoint no longer runs torch's random parameter initialisation before the weights are loaded, which cuts the per-fit model construction from about a quarter of a second to a few tens of milliseconds on a full-size checkpoint. The loaded model is unchanged.
--- a/src/tabpfn/model_loading.py
+++ b/src/tabpfn/model_loading.py
@@ -17,6 +17,7 @@ import urllib.request
 import warnings
 import zipfile
 from collections import OrderedDict
+from contextlib import contextmanager
 from dataclasses import asdict, dataclass
 from enum import Enum
 from importlib import import_module
@@ -43,7 +44,7 @@ from tabpfn.inference_config import (
 from tabpfn.settings import settings
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Iterator
 
     from sklearn.base import BaseEstimator
 
@@ -1053,6 +1054,52 @@ def load_model(
     return result
 
 
+#: The `torch.nn.init` functions that `_skip_parameter_init` turns into no-ops.
+_PARAMETER_INIT_FUNCTIONS = (
+    "uniform_",
+    "normal_",
+    "trunc_normal_",
+    "constant_",
+    "ones_",
+    "zeros_",
+    "eye_",
+    "orthogonal_",
+    "xavier_uniform_",
+    "xavier_normal_",
+    "kaiming_uniform_",
+    "kaiming_normal_",
+)
+
+
+@contextmanager
+def _skip_parameter_init() -> Iterator[None]:
+    """Build a model without torch's random parameter initialisation.
+
+    Every parameter and persistent buffer of a model built here is overwritten by the
+    strict `load_state_dict` that follows, so the initialisation is wasted work: the
+    layers' `reset_parameters` draw random weights that are discarded, about a quarter
+    of a second per build on a full-size checkpoint, paid at every `fit`. For the
+    duration of the block the `torch.nn.init` functions return their tensor untouched;
+    they are restored afterwards. A model built inside it must be loaded before use.
+    """
+    originals = {
+        name: getattr(torch.nn.init, name) for name in _PARAMETER_INIT_FUNCTIONS
+    }
+
+    def leave_uninitialised(
+        tensor: torch.Tensor, *_args: Any, **_kwargs: Any
+    ) -> torch.Tensor:
+        return tensor
+
+    for name in originals:
+        setattr(torch.nn.init, name, leave_uninitialised)
+    try:
+        yield
+    finally:
+        for name, original in originals.items():
+            setattr(torch.nn.init, name, original)
+
+
 def _build_model(
     resolved: str,
     identity: tuple[int, int],
@@ -1081,10 +1128,11 @@ def _build_model(
         "Keys in config that were not parsed by architecture config: "
         f"{', '.join(unused_model_config.keys())}"
     )
-    model = architecture.get_architecture(
-        model_config,
-        cache_trainset_representation=cache_trainset_representation,
-    )
+    with _skip_parameter_init():
+        model = architecture.get_architecture(
+            model_config,
+            cache_trainset_representation=cache_trainset_representation,
+        )
 
     # A checkpoint may carry criterion state for a task it is not being loaded for
     # (save_tabpfn_model writes it for regressors), so it is always kept out of the
--- a/tests/test_model_loading.py
+++ b/tests/test_model_loading.py
@@ -858,3 +858,53 @@ def test__load_ckpt_without_softmax_temperature__uses_legacy_default(
     )
 
     assert inference_config.SOFTMAX_TEMPERATURE == DEFAULT_SOFTMAX_TEMPERATURE == 0.9
+
+
+@pytest.mark.parametrize("architecture_name", ["tabpfn_v2", "tabpfn_v3"])
+def test__load_model__skips_the_random_init_and_matches_the_checkpoint(
+    tmp_path: Path, architecture_name: str
+) -> None:
+    """A model built without parameter initialisation matches the checkpoint exactly."""
+    if architecture_name == "tabpfn_v2":
+        config = _get_minimal_v2_config()
+        source = tabpfn_v2.get_architecture(config, cache_trainset_representation=False)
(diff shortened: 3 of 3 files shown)


oscarkey:
hey! Sorry for the drive-by comment, I just happened to be discussing this in the office yesterday and then spotted your PR. I'm just wondering if you also tried PyTorch's [suggested approach](https://docs.pytorch.org/tutorials/recipes/recipes/module_load_state_dict_tips.html#using-torch-device-meta) of using the meta device? That won't affect other threads in the process and won't miss future initialisation functions, but maybe it doesn't work for us.

Innixma:
Below is from Claude, but TLDR it was actually a bit slower (and adds 750ms cold start time) and also unsafe to do without adding guards around it. Do we have concerns about other threads in the process for this scenario? I'm unsure how we would test all the cases.

Claude:

Thanks, good pointer. I tried it. Same 52 small datasets (23 binary, 16 multiclass, 13 regression), a single estimator fit once per dataset through the AutoGluon wrapper on one GPU, every model build timed in-process:

| build | fit total, 52 fits | per-build median | first build (cold, incl. checkpoint load) |
|---|---|---|---|
| skip-init (this PR) | 15.2 s | 20.9 ms | 549 ms |
| meta + `load_state_dict(assign=True)` | crashed on the first regression dataset | | |
| meta + assign + guards | 15.9 s | 25.3 ms | 1285 ms |

Predictions are bit-identical between skip-init and the guarded meta build on all 52 datasets. The build is about 2 s of the 15 s either way, and the 0.7 s gap between the two runs is within run-to-run noise (a repeat of the skip-init run the day before came out at 14.6 s).

The bare meta variant fails with `Tensor.item() cannot be called on meta tensors` inside `FullSupportBarDistribution.__init__`: one of our architectures keeps a plain tensor attribute that aliases its regression-borders buffer. `assign=True` swaps the buffer for the checkpoint tensor, but the alias still points at the meta tensor, and `_resolve_regression_borders` hands that to the criterion. The "guards" variant walks the modules after loading, re-points any meta attribute at the loaded tensor of the same name and shape, and asserts nothing on meta remains. That works, but it is a name-matching heuristic and a second mechanism to maintain. `assign=True` also makes the parameters alias the cached state dict, so on the CPU path any in-place mutation of a model would corrupt the cache for later estimators.

In a microbenchmark the meta build was ~2x faster (about 15 ms vs 30 ms per build), but that did not survive the real path. I agree with both of your points in favour of meta (no process-wide patch, no list of init functions to keep current); given no end-to-end gain and the two new failure modes, I'll stick with skip-init here. Worth revisiting if host memory during builds becomes the constraint, since assign avoids the second copy of the weights.

oscarkey:
ah thank you this is very useful! If we don't need to get this in urgently, I suggest we fix the architecture so it works with the meta device. From a very quick look, I think we need to make [this line](https://github.com/PriorLabs/TabPFN/blob/7065a12c2b931ede021d20936c9e51ab9d1f1d0b/src/tabpfn/architectures/tabpfn_v3_5.py#L2118) an `@property`. This will then mean we don't need the guard function, which I'd hope will shave off a few ms and bring us close to the patching approach. Although, I'm not sure what's going on with the cold-start time.

For me, I'd prefer to use the recommended approach even if it is a bit slower, as I'm worried that patching creates several tricky latent bugs:
- Anything that takes a reference to one of the random functions while the patch is applied will keep the patched one forever (although you could have the patches raise an error if used after the patch is removed)
- Anything that takes a reference before the patch is applied doesn't get the patch
- If anyone does use TabPFN with multi-threading, we'll potentially silently and randomly break their results

For people who want to go really fast, we can help them use the `TABPFN_MODEL_CACHE_SIZE` option.

What do you think?

Innixma:
Tried your suggestion (from Claude, numbers measured here):

Making `regression_borders` on the v3.5 architecture an `@property` is indeed all the architectures need. With that change, a `torch.device("meta")` build followed by `load_state_dict(assign=True)` leaves no tensor on the meta device for the v2.5, v3 and the small v3.5 test checkpoints, the borders equal the checkpoint's, and predictions through the AutoGluon wrapper are identical to the skip-init build to the last bit. No guard function needed.

The cold-start time is real and I can now place it: `nn.Embedding.reset_parameters` calls `init.normal_`, and on a meta tensor that dispatches through the decomposition path (`torch._refs.normal_` -> `torch._compile`), which lazily imports `torch._dynamo` together with sympy, triton and `torch.distributed`, 864 modules in this environment. A normal `fit`/`predict` never imports dynamo, so this is a pure addition, paid once per process. The meta context itself costs about a millisecond; it is the init call on a meta tensor that triggers it.

| | baseline (init then load) | skip-init (this PR) | meta + `@property` |
|---|---|---|---|
| warm build, v3 classifier (53M params), median of 7 | 194 ms | 22 ms | 30 ms |
| warm build, v3 regressor (58M) | 204 ms | 16 ms | 28 ms |
| warm build, v2.5 classifier (7M) | 30 ms | 6 ms | 10 ms |
| cold first build in a fresh process, incl. checkpoint read | ~0.3 s | 0.12 s | 1.8 to 2.6 s |
| cold-start overhead vs skip-init (the dynamo import) | | | +1.7 to +2.5 s |
| total: first fit of a 200-row task in a fresh process, one build, AutoGluon wrapper, 3 interleaved runs each | 2.6 to 3.8 s | 2.4 to 2.5 s | 4.3 to 5.3 s |
| second checkpoint's first fit in the same process (regressor, its own build) | 0.39 to 0.41 s | 0.14 s | 0.15 s |

The total row varies by over a second between runs of the same variant because the first fit reads the checkpoint over NFS; the second checkpoint's row, measured after the process is warm, isolates the build itself: baseline 0.4 s, skip-init and meta both 0.15 s, so once dynamo is loaded the two approaches are equivalent.

Warm builds are close, as you expected: meta is 8 to 12 ms behind skip-init and still six to seven times faster than the baseline. The cold start decides it. Once a process builds each checkpoint once and shares the network across fits (which is what the AutoGluon wrapper does), the meta approach costs about 2 s per process where skip-init saves 0.2 s per build, so it ends up slower than not optimising the build at all. The venv here sits on NFS, which inflates the import; it stays seconds on a local disk.

So: the property change is worth making on its own for the aliasing bug, but the meta device cannot replace skip-init while torch routes meta initialisation through dynamo, unless a process imports dynamo anyway (`torch.compile` users). On the threading concern for the patch: the library builds on the calling thread and the parallel paths are separate processes, and the patched functions could raise if called after the block ends to make a leaked reference visible; happy to add that if you think it is worth the lines.

oscarkey (commented):
noooo that's so close to working apart from the dynamo import!

Okay let's do the approach in this PR then. As this could be very tricky to debug, I suggest we play extra safe by:
- raising an exception if the fake function is called when the patch has been removed
- having the fake check it's on the same thread as the thread that added entered the context manager, and passing through to the real function if not

if that makes sense to you?

jmkuebler:
@oscarkey @Innixma not sure I completely got the conclusion from this one. but i was check [RES-2875: check pytest runtimes with new default more carefully](https://linear.app/priorlabs/issue/RES-2875/check-pytest-runtimes-with-new-default-more-carefully)

and actually removing this initialization would also half our pytest time on GPU 😅  so I think we should really land sth like this

Innixma:
> @oscarkey @Innixma not sure I completely got the conclusion from this one. but i was check [RES-2875: check pytest runtimes with new default more carefully](https://linear.app/priorlabs/issue/RES-2875/check-pytest-runtimes-with-new-default-more-carefully)
> 
> and actually removing this initialization would also half our pytest time on GPU 😅 so I think we should really land sth like this

I haven't really looked too much into this post 3.5 release.

To be honest I don't know the risks that could occur with this change, so I'm probably not the best person to move it forward into a merge. I only created this because I observed it could be used to speed up the model and it didn't negatively impact any of the logic I was running on the benchmarks.

@oscarkey @jmkuebler Would you be ok with adding a commit to this PR that addresses the remaining concerns?

jmkuebler:
Okay, since this PR is in a Stack, I used the diff here and applied Oscars comments on top. https://github.com/PriorLabs/TabPFN/pull/1304

If it is merged, let's close this one here. 
@Innixma note that using `TABPFN_MODEL_CACHE_SIZE=4` makes things even faster than just skipping the init and should be used for repeated evals within the same process

jmkuebler:
#1304  merged, so closing this one

PR #1263, opened 2026-09-11
Encode numeric categorical columns with a numpy ordinal encoder

Innixma:
Stacked on #1255.

`EncodeCategoricalFeaturesStep`'s ordinal path wrapped sklearn's `OrdinalEncoder`. On a small table almost all of a call is sklearn's input validation rather than the encoding, and the step fits once per ensemble member, so on a categorical-heavy table of a few hundred rows the step was the largest single CPU cost of `fit`. Profiled on a 598 x 31 table with 25 categorical columns and 88 ensemble preprocessing pipelines: 0.54 s of a 0.85 s fit, about 6 ms per pipeline, of which the encoder's own arithmetic is well under a millisecond.

Numeric arrays now go through `NumericOrdinalEncoder`, a `OneToOneFeatureMixin` transformer that does one `np.unique` and one `np.searchsorted` per column and validates nothing. It gives the same output as `OrdinalEncoder(handle_unknown="use_encoded_value", unknown_value=np.nan)`: sorted distinct values become `0..k-1` as float64, NaN stays NaN, a value unseen at fit becomes NaN, and `categories_` lists NaN last for a column that had NaN at fit so the per-column category count that the `ordinal_shuffled` permutation reads is unchanged. Object arrays keep sklearn's encoder. The `ColumnTransformer` wrapper, output names, schema carry-over and the one-hot path are untouched.

On the table above the step goes from 3.97 ms to 1.20 ms per `fit_transform` and from 1.59 ms to 0.27 ms per `transform`; preprocessed outputs are identical. Tests compare the new encoder with sklearn's on plain, missing-value, unknown-at-transform, float32, all-NaN and constant columns, check the encoder choice by dtype, and check the whole step's output, including the shuffled variant, against the step with sklearn's encoder inside. The four `test_sklearn_estimator_checks` failures for `KDITransformerWithNaN` in this environment are present on the base branch as well.


changelog/1263.changed.md
```diff
@@ -0,0 +1 @@
+Faster ordinal encoding of categorical columns: numeric arrays are encoded with a numpy-based encoder instead of sklearn's `OrdinalEncoder`, whose input validation dominated the step on small tables and is paid once per ensemble member. On a 598 x 31 table with 25 categorical columns the step goes from 3.97 ms to 1.20 ms per fit and from 1.59 ms to 0.27 ms per transform. Preprocessed outputs are unchanged.
```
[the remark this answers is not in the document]
Innixma: Shortened to a user-facing summary in 9c64514.

changelog/1255.changed.md
arthur-priorlabs, line 1: This looks like a fragment from another PR; did you forget to restack?

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -93,6 +94,58 @@ def _get_all_cat_indices_after_onehot(
...
+            missing = np.isnan(column)
+            categories = np.unique(column[~missing])
+            if missing.any():
+                categories = np.append(categories, np.nan)
+            self.categories_.append(categories)
+        return self
+
+    def transform(self, X: np.ndarray) -> np.ndarray:
+        """Map each value to the position of its category; NaN where there is none."""
+        X = np.asarray(X, dtype=np.float64)
```
arthur-priorlabs, line 125: Same as above.

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -93,6 +94,58 @@ def _get_all_cat_indices_after_onehot(
...
+    `categories_` lists NaN last for a column that held NaN at fit, as sklearn's does,
+    so the category count per column is the same. sklearn's encoder spends most of a
+    call on input validation, which on a small table dwarfs the encoding itself; this
+    one assumes a numeric 2D array and validates nothing.
+    """
+
+    def fit(self, X: np.ndarray, y: None = None) -> NumericOrdinalEncoder:
+        """Record each column's sorted distinct values."""
+        del y
+        X = np.asarray(X, dtype=np.float64)
```
arthur-priorlabs, line 112: Footgun when `X.dtype == np.int64`: we'll get collisions at high values. Note that `np.issubdtype(np.int64, np.number)` is True, cf. `_ordinal_encoder` below.
May be worth adding a test for as well.

tests/test_preprocessing/test_encode_categorical_features_step.py
```diff
@@ -679,3 +684,115 @@ def test__encode_categorical__ordinal__one_row_fit_learns_every_category() -> No
...
+    encoder = NumericOrdinalEncoder()
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        expected_fit = reference.fit_transform(X)
+        expected_test = reference.transform(X_test)
+
+    np.testing.assert_array_equal(encoder.fit_transform(X), expected_fit)
+    np.testing.assert_array_equal(encoder.transform(X_test), expected_test)
+    assert encoder.transform(X_test).dtype == expected_test.dtype
+    assert [len(c) for c in encoder.categories_] == [
```
arthur-priorlabs, line 734: doesn't check the actual order; while we might want to keep it as an implementation detail, that could silently change predictions

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -93,6 +94,58 @@ def _get_all_cat_indices_after_onehot(
...
+            if missing.any():
+                categories = np.append(categories, np.nan)
+            self.categories_.append(categories)
+        return self
+
+    def transform(self, X: np.ndarray) -> np.ndarray:
+        """Map each value to the position of its category; NaN where there is none."""
+        X = np.asarray(X, dtype=np.float64)
+        encoded = np.full(X.shape, np.nan, dtype=np.float64)
+        for index, categories in enumerate(self.categories_):
```
arthur-priorlabs, line 127: We instantiate the output from `X.shape`, but it fill by iterating `self.categories_`, which would make an oversized `X` have NaN columns instead of errorring like sklearn does

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
arthur-priorlabs, line 273: We're calling `np.unique` twice here.

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
```diff
@@ -93,6 +94,58 @@ def _get_all_cat_indices_after_onehot(
...
+            known = categories[~np.isnan(categories)]
```
arthur-priorlabs, line 128: Can't we compute this once and for all at fit time instead?

src/tabpfn/preprocessing/steps/encode_categorical_features_step.py
arthur-priorlabs, line 271: Not touched by this PR, but since we're doing perf improvements for categoricals: having `i in categorical_features` computes linear tests inside the for loop, where a `set()` conversion would make it cheaper.

arthur-priorlabs (changes requested):
The main concern is large integers being cast to float64; the rest is perf/nit comments.

PR #1270, opened 2026-09-14
Correct the target prior shift of majority_downsample row subsampling

psinger-prior:
## Summary

`SAMPLE_SUBSAMPLING_METHOD="majority_downsample"` now corrects the target prior shift it introduces, automatically and with no new setting. The sampler keeps every row outside the most frequent target value and subsamples that value, so every estimator's context over-represents the rare targets; predicted probabilities and regression means inherited that shift. The fix multiplies the predicted distribution by the training-to-context prior ratio, which is exact under the label-shift assumption and free. Nothing changes for any other sampling method. Follow-up to #1253; supersedes the closed #1261, which wrapped the same correction in a five-mode `prediction_scaling` option that turned out to be more API than the feature needs.

## What changed

```diff
 TabPFNEnsemblePreprocessor
   _get_subsample_indices_for_estimators
+  row_sampling_distribution_        # per-row inclusion probabilities from the budget

 TabPFNClassifier.logits_to_probabilities
   temperature -> softmax / average
+  apply downsample_correction_weights_   (training prior / context prior, per class)
   balance_probabilities (if set)

 TabPFNRegressor._reduce_accumulated_logits
   average -> log
-  / temperature
+  temper_and_correct_logits(logits, T, downsample_correction_log_weights_)
```

- **Preprocessor.** Under `majority_downsample` with a numeric budget, records the inclusion probability of every training row: one for non-majority rows, the budget ratio for majority rows. A tied majority falls back with the sampler and records no shift.
- **Classifier.** Exact per-class weights, training prior over context prior, applied to the averaged probabilities before optional balancing. For binary tasks the correction is monotone and preserves ranking.
- **Regressor.** Per-bar log weights on the raw-space bar distribution, from a histogram of the training target weighted by the inclusion probabilities. Because the weights come from the sampling design rather than realized context counts, sparse bars cannot inflate them. Applied after the ensemble temperature; every output type follows since it acts on the shared logits.
- **Tuning.** Holdout folds carry their own correction: the classifier passes each fold clone's weights per sample into the objective, ones where a fold was too small to be subsampled, and the regressor keeps fold logits and weights separate so the sweep composes `logits / T + log_w` exactly as prediction does.
- **Guards and docs.** Batched prediction raises under this sampler since the correction is per-dataset state. The old class-prior warning in the sampler is gone, since the correction handles what it warned about.

## Results

Same laptop-scale setups as #1253: 4 estimators, 10k rows of context per estimator, seed 0, MPS.

| dataset | metric | majority_downsample before | with correction | default sampler |
|---|---|---|---|---|
| credit card fraud (0.17% positive) | log loss | 0.00627 | 0.00320 | 0.00378 |
| credit card fraud | avg precision | 0.773 | 0.773 | 0.691 |
| credit card fraud | mean predicted rate (base 0.0015) | 0.0048 | 0.0014 | 0.0015 |
| freMTPL2 loss cost (3.7% nonzero) | Tweedie skill | -0.09 | 0.04 | -0.81 |
| freMTPL2 loss cost | cost ratio | 2.98 | 0.56 | 0.11 |
| California housing (continuous) | RMSE | no shift, correction off | 0.4161 | 0.4161 |

A 3-seed x 3-budget sweep on the insurance data had the corrected arm best or within noise of best at every budget, worst case skill -0.02 against -1.35 uncorrected. For regression the correction assumes the model adopts the context prior fully, which holds only approximately; the level stayed off by up to a factor of two, in either direction depending on target transform, though far closer than without it.

## Review guide

- Start with `src/tabpfn/downsample_correction.py`. Four short functions carry all the math; the estimators only wire them in.
- Then the preprocessor block in `src/tabpfn/preprocessing/ensemble.py` that derives `row_sampling_distribution_`. It reuses the sampler's own count allocation, so the probabilities match what the sampler actually did.
- In the classifier, `_compute_downsample_correction` and the weights step in `logits_to_probabilities`. The preprocessor is now built before the tuning step, because the correction needs the context prior and tuning must see corrected probabilities. The two were independent otherwise.
- In the regressor, `_compute_downsample_correction`, the `apply_downsample_correction` flag on `_compute_aggregated_logits`, and the holdout fold tuple gaining a fourth element. `find_regression_optimal_temperature` accepts three- or four-tuples so existing callers are untouched.
- Mechanical: batched-predict guards, docstring edits, the removed warning and its test, one existing regressor test updated for the four-tuple.
- Design choice: weights from inclusion probabilities rather than realized context counts. An earlier version histogrammed context rows per bar and blew up on sparse bars, 28x cost ratio on insurance under the default sampler. The design-based version is exact for the single downsampled group and identity for uniform samplers.
- Design choice: no opt-out switch. Every run showed the uncorrected output worse on calibration with identical or better ranking. An inference-config boolean is trivial to add later if asked.
- Not certain about: the regression correction's residual level error. It is real and documented; a holdout-fitted level alignment would fix it but belongs in `tuning_config` as its own feature.

## Testing

- `tests/test_downsample_correction.py`, 33 tests: the weight math including an exactness check that the class weights recover the training posterior under synthetic label shift, the inclusion-probability plumbing and its fallbacks, both estimators end to end, differentiable-input paths including bfloat16 targets, batched-predict guards, and the tuning composition: fold weights differ from the parent's under an absolute budget, folds without subsampling are scored with weight one, prediction composes temperature then correction, and a spy on the temperature sweep confirms it receives `logits / T + log_w` for every candidate temperature.
- Existing suites pass locally: ensemble preprocessing, inference tuning, classifier and regressor interfaces, save/load, consistency, finetuning, batched shape safety. The only failures on this machine are checkpoint downloads gated on license acceptance and are unrelated.
- Benchmark numbers come from ad hoc scripts outside the repo and are not part of CI.

## Breaking changes / follow-ups

- Predictions change only for `SAMPLE_SUBSAMPLING_METHOD="majority_downsample"`, which now returns corrected probabilities and distributions. There is no flag to restore the old output.
- `predict_proba_batched` and `predict_batched` raise under that sampler.
- Follow-up: a holdout-fitted level alignment in `tuning_config` for regression, where the analytic correction leaves a residual.


psinger-prior:
Decided to just put the scaling in for the majority downsample, after some tinkering in now closed PR to enable different types of rescalings.

src/tabpfn/classifier.py
```diff
@@ -241,6 +246,11 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator):
     """The ensemble configurations used during fit.
     Stored for reuse in prompt tuning."""
 
+    downsample_correction_weights_: np.ndarray | None
```
LeoGrin, line 249: Agree with Claude here:

> fit_from_preprocessed never resets this attribute, so weights from a                                                                                                                                          previous fit() survive into the finetuning path. forward() goes through                                                                                                                                           logits_to_probabilities, which picks them up via getattr, so in the usual                                                                                                                                         loop (fit() to evaluate, fit_from_preprocessed() + forward() to train)                                                                                                                                              every training step's loss is computed on probabilities reweighted for a                                                                                                                                           context prior the batch never had. Silent skew, classifier only today

LeoGrin: (I think we also need better tests for finetuning 🤔)

LeoGrin (dismissed):
LGTM! One small thing

PR #1274, opened 2026-09-14
Add the decoder attention readout to the multitask architecture

LeoGrin:
Generated by Claude Code

Adds `ManyClassDecoder.attention_weights` to the `tabpfn_v3_5` architecture, the same per-train-row readout the v3 decoder has had since 8.3.0. `tabpfn_extensions.interpretability.get_decoder_readout` calls it on whatever decoder the fitted model holds, so without it the readout raises `AttributeError` on a v3.5 model, and the extensions compat job fails as soon as v3.5 is the default (#1273). The decoder is otherwise the same retrieval attention as v3's, so the method is a direct port.

<details>
<summary>How this was checked</summary>

- New test mirrors the v3 one: the weights are non-negative, sum to 1 over train rows, and collapsed by class label reproduce the fused forward's logits, with and without softmax scaling.
- tabpfn-extensions `tests/test_interpretability_decoder_readout.py` run as the compat job does (`uv run --with <this branch> --all-extras pytest`) with `TABPFN_MODEL_VERSION=v3.5`: 7 passed. The same tests fail with `AttributeError` on main under that setting.

</details>


eliott-kalfon (approved):
LGTM, thanks for ensuring compatibility @LeoGrin

PR #1275, opened 2026-09-14
Precision fixes in the multitask architecture: fp16 stays finite, fp64 stays fp64

LeoGrin:
Generated by Claude Code

Three precision problems in `tabpfn_v3_5`, all surfaced by making it the default in #1273. With `inference_precision=torch.float16` the model returned near-uniform predictions on torch 2.5 through 2.8, on CPU and CUDA, because two fp16 overflows produced NaN logits that `_nan_safe_output` turned into zeros. With `inference_precision=torch.float64` the cell embedding silently ran in fp32, so a float64 forward was not chunk-invariant on GPUs whose fp32 GEMMs depend on the batch size. Each op now computes at the precision the caller asked for, widening half dtypes to fp32 where fp16 cannot hold the values.

<details>
<summary>The three changes</summary>

- `_DtypeMatchingRMSNorm` computes in fp32 for fp16 inputs on any device: the residual stream at the last ICL blocks has entries whose squares exceed the fp16 range, so `F.rms_norm` in fp16 got an infinite variance and returned zeros. Torch releases before 2.14 take that mean in fp16 on CPU and CUDA alike.
- `_batched_scaled_dot_product_attention` computes in fp32 for fp16 on CPU: the softmax-scaling layer inflates the ICL queries to about 1e4, so `q.k^T` exceeds the fp16 range, and the CPU flash kernel of older torch returns NaN. CUDA's kernels accumulate in fp32 and stayed finite, so this one is CPU-only.
- The Fourier cell embedding and the metadata projection compute at `promote_types(dtype, float32)` instead of a hard `.float()`: half dtypes still widen to fp32, fp64 stays fp64. Nothing changes for fp32, bf16 or fp16 users.

TabPFN-3 and TabPFN-3.5-Fast were not affected by the overflows: their activations stay inside the fp16 range.

</details>

<details>
<summary>How this was checked</summary>

- Unit tests per op at the magnitudes seen in the shipped checkpoint. On torch 2.5.0 the two overflow tests fail before the change and pass after; the fp64 test checks the embedding against a hand-computed fp64 reference.
- `test__predict_proba_batched__fp16_matches_fp32` and the regressor twin, which fail in the `lowest-direct` CI jobs of #1273 with 3.5 as default, pass on torch 2.5.0 on CPU with this change, and on an H100 with torch 2.5.0+cu124 and 2.14+cu130.
- H100, torch 2.5.0, before the norm fix: 3.5 fp16 vs fp32 mean |dproba| 0.44 with no attention NaN; after: 0.008. TabPFN-3: 0.021, 3.5-fast: 0.007.
- `test__kv_cache_chunking__matches_unchunked[*-cuda]` failed on the T4 CI runner for 3.5 with fp32-sized differences in a float64 run; H100 and CPU were exact. The embedding change is aimed at that; the CI GPU job on #1273 is the check.
- v3.5 architecture tests on torch 2.14: 108 passed.

</details>

<details>
<summary>Speed</summary>

TabPFN-3.5, 4 estimators, 5,000 train and 1,000 test rows, `predict_proba`, best of 3:

| Mode | H100 before | H100 after | A100 before | A100 after |
| --- | --- | --- | --- | --- |
| default `"auto"` | 0.54 s | 0.53 s | 1.03 s | 1.00 s |
| forced `torch.float16` | 0.48 s | 0.55 s | 0.89 s | 1.03 s |

The default is unchanged: under autocast the residual is fp32, so the fp16 paths never run. Forced fp16 pays about 15 percent, from the fp32 copy of the residual the norm now makes on each of its 48 calls. Forced fp16 was returning near-uniform predictions on torch before 2.14 and is lossy on 3.5 on every torch, so correctness is preferred over that 15 percent. A hand-rolled norm that reduces in fp32 and multiplies in fp16 would recover most of it if it ever matters.

</details>

Stacked on #1274; needed by #1273.


jmkuebler (dismissed):
thanks Leo! Good as a workaround, but we should generally ask whether we want to support forced FP16 precision... I also found that above torch 2.9.0 this issues wouldn't arise. So ideally mid term we leave that kind of house-keeping to torch itself.

brendan-priorlabs (approved):

PR #1278, opened 2026-09-15
CI: 60 minutes for the compatibility matrix

LeoGrin:
Generated by Claude Code

Raises the job timeout of the compatibility matrix from 30 to 60 minutes. The macOS 15 job already needs 25 to 30 minutes for the PR test set, so single runs and merge-queue runs fail on runner speed alone: #1274 was rejected by the queue this way while its own CI was green. With TabPFN-3.5 as the default (#1273) the same job takes about 47 minutes.

We need to make the tests faster soon. The bump only removes a flaky failure so the stack can merge. It sits between #1274 and #1275 and replaces the timeout commit that #1275 carried.

<details><summary>Measured macOS 15 job times, PR test set</summary>

| Run | Time |
|---|---|
| #1274 PR run, first head | 28 min |
| #1274 PR run, final head | 25 min |
| #1274 merge-queue run | killed at 30 min |
| #1275 PR run | 24 min |
| #1275 earlier PR run | killed at 30 min |
| #1273 PR run, 3.5 default | 47 min |

The test step itself is about 19 of the 24 minutes on a good run. Downloading and caching the models takes another 3.5 minutes because the model cache does not hit.

</details>

<details><summary>Speedups to follow, ranked by payoff</summary>

1. Run pytest with two workers. `pytest-xdist` is already a dev dependency, and every job runs the suite serially on a 3-core runner.
2. Load the fast 3.5 checkpoint in CI for tests that do not pin a version. Its forward pass on the macOS runner is 3 to 4 times cheaper than full 3.5 and cheaper than TabPFN-3.
3. Fix the model cache miss.
4. Trimming the v3/v3.5 and cpu/mps parametrization saves only about one minute of 19 and is not worth the coverage.

</details>


alanprior (approved):
Approving, but I think there are optimization opportunities.

See https://github.com/PriorLabs/TabPFN/pull/1279 for example

alanprior (changes requested):
Ok, we decided to merge a fix that fixes latency rather than this one, solving (hopefully) the core issue cc @oscarkey's

jmkuebler (approved):

alanprior (approved):

PR #1281, opened 2026-09-15
Run unpinned tests on the fast TabPFN-3.5 checkpoint

jmkuebler:
Stacked on #1275.

Tests that do not pin a model version now run on the fast TabPFN-3.5 checkpoint. A session-scoped autouse fixture in `tests/conftest.py` sets `settings.tabpfn.model_version` to `V3_5_FAST` unless `TABPFN_MODEL_VERSION` is set in the environment, so `TABPFN_MODEL_VERSION=v3.5 pytest` still runs everything on the full model. Tests that pin `ModelVersion.V3_5` or its filename are unaffected and keep covering the full model.

<details><summary>Local measurements, M5 Pro, regressor interface PR tests, stack branch</summary>

| Default model | Result | Sum of test call times |
|---|---|---|
| full v3.5 | 116 passed | 116s |
| fast v3.5 via this fixture | 116 passed | 73s |

The slowest test in the file, batched regression on MPS, goes from 8.3s to 3.6s, under the 4.6s TabPFN-3 takes on main. Tests pinned to `V3_5` still take full-model time, e.g. 4.3s for the demo-dataset r2 test versus 1.5s for the fast variant.

</details>

This alone brings the macOS job back to roughly TabPFN-3 timings, which already brushed the old 30 minute limit, so it complements rather than replaces the timeout bump in #1278 and the xdist follow-up.


oscarkey (dismissed):
sgtm!

jmkuebler:
@oscarkey if you could please reapprove, this got left out as we skipped the line with #1275  but would land that afterwards.

oscarkey (approved):

PR #1288, opened 2026-09-17
Stop the regression border remap from losing bucket mass to float32

jmetzen-pl:
🤖 Generated by Claude Code

**Summary.** The bar-distribution border remap computed each destination bucket's mass by
differencing a cumulative sum in float32, whose resolution near a CDF of 1 is ~6e-8, so
any smaller bucket became exactly zero and `-inf` after `log()`: infinite NLL for a
target landing in one. `_cdf` also pinned the CDF to 0/1 at the outer borders, which
discarded the half-normal tail mass `FullSupportBarDistribution` puts there. The
differencing now runs in float64, the outer buckets are read as the tails they are, and
each destination bucket is differenced from whichever end of the distribution it is
nearer to, so the two halves of the grid are treated alike. Tracked internally as PRI-361.

This is the first of two stacked PRs. The follow-up moves the regressor's ensemble
pooling into log space so that masses below float32's floor survive the cast back; it
is split out because it touches the prediction path and deserves its own review.

Two behaviour changes, both detailed under **Blast radius**: point predictions move,
by round-off for well-behaved targets and measurably for skewed ones; and with
`eval_metric="nll"`, temperature tuning previously discarded every candidate whose
holdout loss came out infinite and silently kept the default temperature, so a tuned
NLL temperature may now differ.

## Blast radius

Measured on this branch's head against `main`, on a synthetic 8-member ensemble
assembled the way `predict` assembles one: a 5000-bucket shared grid, members
alternating `target_transform=None` (an identity remap) and a fitted
`SafePowerTransformer`, with each member's `borders_t` produced by the real
`transform_borders_one`. The per-member head is a Gaussian rather than a trained
model, so read the magnitudes as indicative. 512 test rows; deltas are in units of the
target's standard deviation.

| targets | shared borders outside the member grid | rows with infinite NLL, `main` -> PR | max d(mean) | max d(q10) | max d(mode) |
|---|---|---|---|---|---|
| `N(0, 1)` | 1 of 5001 | 8 -> 0 | 9e-7 | 1e-6 | 4e-3 |
| lognormal | 707 of 5001 | 11 -> 5 | 3e-3 | 5e-2 | 1.5 |
| `t(3)`, shifted | 2 of 5001 | 9 -> 2 | 1e-5 | 1e-6 | 5e-3 |

**Who is affected.** `TabPFNRegressor` only. `translate_probs_across_borders` has
exactly two callers, both in `regressor.py`: `predict` and the batched decode path.
The classifier and the finetuning paths never call it. Everything downstream of the
pooled logits moves with them: `predict(output_type=...)` for `mean`, `median`,
`mode` and `quantiles`; the `logits` handed back by `output_type="full"`; and any NLL,
CRPS or sampling a user computes from those.

**Point predictions are not unchanged**, contrary to what this PR claimed before it was
measured. For well-behaved targets they move by ~1e-6 sd, which is round-off. Where a
member's warped grid ends well inside the shared grid -- skewed targets, 707 of 5001
shared borders below the member's grid in the lognormal row above -- that member's tail
mass is now spread along the half-normal instead of piled into the one bucket
straddling its border, and the mean and the quantiles move with it. `mode` moves most,
being an argmax over 5000 buckets where near-ties flip.

**NLL is corrected, not uniformly improved.** For a single member on a grid 7% narrower
than the shared one, of 1630 rows finite under both, 746 come out lower and 752 higher;
the worst single-row increase is +1.94 nats. Rows that used to land in the over-filled
border bucket were being scored too generously.

**Infinite NLLs are reduced, not eliminated.** The recovered mass is still cast back to
float32, whose smallest subnormal is ~1e-45, so 5 of 512 rows survive as `inf` in the
lognormal case. That is what the stacked follow-up fixes.

**ROCm has no float64 `erfc` kernel.** On an MI250X it raises HIP error 209,
`hipErrorNoBinaryForGpu`, under both torch 2.6.0 and 2.7.1, so `predict` would crash
outright on an AMD GPU rather than just cost the upcast. The tail helper therefore
computes `erfc(z)` as `erfcx(z) * exp(-z^2)`: ROCm does ship `erfcx` in float64, and the
identity agrees to 5e-14 relative, down to subnormal results.

**Cost.** Every row below is `main` against this branch's head, the two measured in
the same job on the same machine. The remap alone, one member, best of 5:

| device | rows x buckets | `main` | this PR | ratio |
|---|---|---|---|---|
| CPU, Apple M-series | 2000 x 5000 | 54 ms | 80 ms | 1.5 |
| MPS | 2000 x 5000 | 25 ms | 82 ms, on the CPU | 3.3 |
| MI250X, ROCm 6.2.4 | 2000 / 10000 / 40000 x 5000 | 3.0 / 14.5 / 57.9 ms | 6.4 / 26.3 / 100.9 ms | 1.7-2.1 |
| A100-SXM4-40GB, cu128 | 2000 / 10000 / 40000 x 5000 | 2.7 / 13.1 / 48.8 ms | 7.4 / 27.4 / 102.7 ms | 2.1-2.7 |
| RTX PRO 6000 Blackwell, cu128 | 2000 / 10000 / 40000 x 5000 | 1.3 / 8.7 / 36.8 ms | 4.4 / 18.6 / 71.7 ms | 1.9-3.4 |

End to end, with a v2 checkpoint and 4 estimators, which is the share a user pays.
Half of that ensemble remaps onto the grid it is already on and now returns early, so
these are well below the ratios above:

| case, train / test | `main` | this PR |
|---|---|---|
| CPU, 600 / 400 | 1.144 s | 1.150 s |
| MPS, 1000 / 2000 | 0.880 s | 0.987 s |
| MI250X, 2000 / 1000 and 10000 | 0.173 / 0.623 s | 0.177 / 0.659 s |
| A100, 2000 / 1000 and 10000 | 0.166 / 0.566 s | 0.170 / 0.612 s |
| RTX PRO 6000, 2000 / 1000 and 10000 | 0.065 / 0.262 s | 0.068 / 0.291 s |

The FP64 throughput of the card turns out not to matter: RTX PRO 6000 runs FP64 at 1/64
of FP32 and MI250X at 1/1, yet the two pay the same on the remap itself, because these
kernels (cumsum, gather, erfc) are bandwidth-bound and the upcast doubles the bytes.
What matters is how fast the rest of `predict` is, since that sets the remap's share:
10000 test rows cost +6% on MI250X, +8% on A100 and +11% on the RTX, whose model forward
is quickest. MPS is worse at +12%, because float64 does not exist there and the remap
falls back to the CPU.

Peak memory is at or below `main` wherever the batch is large enough to chunk. A
40000 x 5000 remap allocates 2216 MiB against `main`'s 2289, and end-to-end on 10000 test
rows `predict` peaks at 1651 MiB on the MI250X against 1696, and 1628 against 1673 on
both NVIDIA cards. A single unchunked batch still pays for the dtype: 420 -> 767 MiB at
2000 x 5000, and 336 -> 499 MiB end-to-end on 1000 test rows.

**`halfnormal_with_p_weight_before`** is bit-identical for a Python float and for a
float32 tensor, which covers every caller inside TabPFN, since checkpoint borders are
float32. It moves by <=2e-8 relative for a float64 tensor, which reaches a user who
builds a float64 `FullSupportBarDistribution` of their own.

**One dtype change for direct callers of `translate_probs_across_borders`** (not part
of the public API, but importable). It now returns `logits.dtype`. On `main`, a float64
`frm` against float32 `logits` promoted the result to float64 by accident. The
regressor is unaffected, because sklearn preserves float32 through `inverse_transform`
and `borders_t` is therefore float32, but a direct caller passing float64 borders now
gets float32 back.

<details>
<summary><b>Full detail</b> — mechanisms, diff, measurements, tests</summary>

## Issue

No GitHub issue. Tracked internally as PRI-361; the bug is described in full below so
this PR stands on its own.

## Motivation and Context

`translate_probs_across_borders` computes each destination bucket's mass by differencing
a cumulative sum. A bucket's mass is therefore the difference of two CDF values, so its
resolution is that of the CDF near 1, not near 0: about 6e-8 in float32, which is what
the function ran in. Any bucket holding less than that landed on exactly zero.
`regressor.py` then takes `.log()` of the pooled probabilities, so those buckets became
`-inf` and `FullSupportBarDistribution`'s NLL was `inf` for any target landing in one.

This only bit density and likelihood evaluation, and there it did real damage, because
the usual workaround of dropping non-finite rows from a mean removes exactly the rows
where the model put least mass near the target, unevenly across the models being compared.

Three mechanisms are involved, plus one thing that had to be fixed to measure any of it.

**1. float32 cancellation.** `torch.cumsum(probs) - probs` already loses mass below the
running total's resolution, and differencing that degraded CDF loses more; the
`clamp_min(0.0)` then turns the negative rounding residue into an exact zero rather than
leaving it detectably negative.

**2. The remap discarded the distribution's tails.** `FullSupportBarDistribution` puts a
half-normal on each outer bucket, so half of an outer bucket's mass lies beyond the grid
border. `_cdf` pinned the CDF to 0 at `frm[0]` and 1 at `frm[-1]`, which drops both
half-tails at any precision, so every destination bucket outside the source range was
exactly zero whenever `to` extended past `frm`, and the mass was piled into the single
destination bucket straddling the border. This happens when a member's y-transform gives
a narrower grid than the pooled `znorm_space_bardist_` grid.

**3. Cancellation against a CDF near 1.** Even in float64, a CDF value near 1 has
rounded away anything below ~1e-16 of the total, and in the upper tail it is exactly 1.0.
Differencing it loses every bucket smaller than that in the upper half of the grid while
the same buckets in the lower half survive, so a distribution and its mirror image did
not remap to mirror images. An asymmetric zero pattern is exactly what biases a model
comparison toward whichever side reads low.

## What changed

```diff
 utils.translate_probs_across_borders
+  differencing runs in float64, result cast back to the caller's dtype
   _cdf  ->  _cdf_and_survival
-    prob_left_of_ys[ys <= borders[0]] = 0.0     # dropped the half-normal tails
-    prob_left_of_ys[ys >= borders[-1]] = 1.0
+    outer buckets read as the half-normal tails `forward` already uses (erfc)
+    survival function accumulated from the top, alongside the CDF
+  each bucket differenced from the nearer end: CDF below the median, survival above

 FullSupportBarDistribution.halfnormal_with_p_weight_before
+  reference half-normal built in `range_max`'s dtype, so the tail sigma is float64
```

- **The upcast.** `_TRANSLATE_CHUNK_BUDGET_ELEMENTS` counts elements, not bytes, so it is
  halved to keep this step's peak memory where it was.
- **Row-independent work.** `ys` is the destination grid, one grid for every row, but it
  was repeated to `(batch, len(to))` first, so the searchsorted, the int64 bucket index,
  the bucket share and the two tail masks were recomputed per row. Keeping them
  one-dimensional is bitwise identical and removes about half of the upcast's cost.
- **The identity case.** `REGRESSION_Y_PREPROCESS_TRANSFORMS` leaves every other member's
  target untransformed, so 16 of 32 remaps on a fitted 8-estimator regressor have `frm`
  bitwise equal to `to`. Those return the softmax directly, which is what the remap
  computes for them anyway.
- **The tails.** Written as a complementary error function, not a literal
  `1 - HalfNormal.cdf(...)`: the subtraction form cancels to exactly zero a few sigma
  out, which is the same defect one level down. Interior borders are untouched, because `CDF(borders[1]) == probs[0]` under
  either reading, and an identity remap still returns its input.
- **Nearer-end differencing.** A cumulative probability is accurate where it is small.
  The survival function is accumulated from the top with a reversed `cumsum`, both are
  pinned at the ends of `to`, and each bucket takes the CDF difference where the CDF is
  at most 0.5 and the survival difference otherwise. In the upper tail this reduces to
  the closed form `p_outer * (SF(a) - SF(b))`, so no separate tail code is needed, and
  the output still telescopes to 1.
- **`halfnormal_with_p_weight_before`** built its reference from a hardcoded
  `torch.tensor(1.0)`, so `icdf` ran in float32 and capped the returned scale at ~1e-8
  relative accuracy. It now follows `range_max`'s dtype and is bit-identical for float32
  and float callers. Without it the tail does not put exactly half the outer bucket's
  mass inside the grid, which the tail test checks to 1e-12.
- **MPS.** float64 does not exist there, so that device computes the differencing on the
  CPU. The conversions are staged (move, then cast) because a combined
  `.to(device=..., dtype=...)` off MPS converts on the source device and silently
  returns all zeros rather than raising.
- **`tests/test_settings.py`** lowered the process-global MPS memory cap and never put it
  back, so every later test in the session ran with a ~25 MB budget and unrelated MPS
  allocations OOM'd. `torch.mps.set_per_process_memory_fraction` has no getter, so a
  fixture re-initialises the settings with their default.

</details>

## Public API Changes

-   [x] No Public API changes

`halfnormal_with_p_weight_before`'s annotation widens to `float | torch.Tensor` to match
how the class already calls it. No signature changes otherwise.

## How Has This Been Tested?

Unit tests in `tests/test_utils.py`, `tests/test_architectures/test_bar_distribution.py`
and `tests/test_settings.py` pass locally (53 tests). Measured on this branch's head, on a
5000-bucket grid whose smallest source bucket holds 1e-12:

| destination grid | `main` | this PR | this PR, float64 caller |
|---|---|---|---|
| nested, 2.5% narrower | 1442 / 5000 | **0** | 0 |
| 3.1% wider | 1576 / 5000 | **118** | 44 (22 per side) |

Both `main` and the middle column are what a float32 caller sees, which is what the
regressor is. Of the 44 a float64 caller is left with, all are symmetric and
irreducible: they sit far enough into the tails that `erfc` itself underflows in
float64 and the true mass is below 1e-300. The other 74 are lost on the way back to
float32, whose smallest subnormal is ~1e-45; the stacked follow-up recovers them by
returning the log instead of the probability.

New tests cover: the upcast (with a companion test pinning the upcast as the cause), the
half-normal tails against the analytic half-normal, interior borders staying put, the
identity remap, the upper tail against the closed form, mirror symmetry of a distribution
and its flip with no tail bucket involved, the sum to 1 when the whole destination grid
lies inside one source tail, a degenerate zero-width outer bucket, the dtype-faithful
half-normal scale, and MPS/CPU parity through both the chunked and unchunked dispatch.

The end-to-end regressor tests were not run locally on the final head: they need the
v3.5 checkpoint download, which is license-gated in this environment. CI covers them.

## Known limitation

After this PR the recovered mass is still cast back to float32, which has no value below
~1e-45, so a head with a narrow enough kernel still loses its far buckets at that
boundary -- 5 of the 512 rows in the skewed case above. The stacked follow-up PR
addresses this by pooling the ensemble in log space.

## Checklist

-   [x] A changelog entry has been added (see `changelog/README.md`).


changelog/1288.fixed.md
```diff
@@ -0,0 +1 @@
+Regression density and likelihood evaluation no longer reports infinite NLL for targets in small buckets. The bar-distribution border remap computed each destination bucket's mass by differencing a cumulative sum in float32, where the resolution near a CDF of 1 is ~6e-8, so any smaller bucket became exactly zero and `-inf` after the `log()`. The differencing now runs in float64, the two outer buckets are read as the half-normal tails `FullSupportBarDistribution` actually puts there instead of being pinned to 0 and 1 (which dropped that mass at any precision whenever the destination grid was wider than the source), buckets inside a tail are computed in closed form rather than by differencing a CDF that has rounded to 1.0, and the ensemble is pooled in log space so no bucket mass is lost to float32 on the way in. Point predictions and CRPS are unchanged. With `eval_metric="nll"`, temperature tuning previously skipped every candidate whose holdout loss was infinite and silently fell back to the default temperature; that sweep now discriminates, so a tuned NLL temperature may differ.
```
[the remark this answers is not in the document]
jmetzen-pl: _Generated by Claude Code_

Fixed in d9c0176d — the fragment is now two user-facing sentences (finite NLL where it used to be infinite; a tuned `eval_metric="nll"` temperature may move), and the mechanism stays in the PR description.

src/tabpfn/architectures/shared/bar_distribution.py
```diff
@@ -501,13 +501,29 @@ def assert_support(self, *, allow_zero_bucket_left: bool = False) -> None:
...
-        s = range_max / torch.distributions.HalfNormal(torch.tensor(1.0)).icdf(
-            torch.tensor(p),
+        """Build a half-normal placing ``p`` of its mass below ``range_max``.
+
+        The reference half-normal is built in ``range_max``'s own floating
+        dtype. Hardcoding ``torch.tensor(1.0)`` would evaluate ``icdf`` in
+        float32 and cap the returned scale at ~1e-8 relative accuracy, which
+        is a float32-sized error in an otherwise float64 tail computation.
+        A float (or non-floating) ``range_max`` keeps the default dtype, so
+        float32 callers are unaffected.
```
jmetzen-pl: describe the behavior of this function, not the motivation for this behavior

src/tabpfn/regressor.py
```diff
@@ -1700,19 +1702,24 @@ def _compute_aggregated_logits(
                 unit="estimator",
                 disable=not self.show_progress_bar,
             ):
+                # Log-probabilities, not probabilities: float32 cannot hold a
+                # bucket mass below ~1e-45, so pooling probabilities throws away
+                # exactly the small buckets the float64 differencing recovered
+                # and turns them into `-inf` at the `.log()` below. Their logs
+                # are ordinary numbers, so the whole reduction stays in log
+                # space and nothing is lost on this boundary.
```
jmetzen-pl: this comment is too long and requires too much context (which float64 differencing?)

src/tabpfn/regressor.py
```diff
@@ -2365,6 +2377,34 @@ def to(self, device: DevicesSpecification) -> None:
...
+    average_before_softmax: bool,
+) -> torch.Tensor:
+    """Fold one estimator's log-probabilities into the running accumulator.
+
+    The ensemble pools its members as an arithmetic mixture by default, and
+    `log(mean_i p_i) = logsumexp_i(log p_i) - log(n)`, so the sum over members
+    is a running `logsumexp` rather than a running sum. Doing it this way keeps
+    the reduction in log space from end to end: a bucket mass of 1e-300 is not
+    representable in float32 but its log is, so nothing is lost to the dtype on
+    the way in, and `_reduce_accumulated_logits` subtracts the `log(n)`.
```
jmetzen-pl: this docstring is overly long for the small functionality this function offers

src/tabpfn/utils.py
```diff
@@ -354,10 +356,43 @@ def _map_to_bucket_ix(y: torch.Tensor, borders: torch.Tensor) -> torch.Tensor:
...
+
+    `FullSupportBarDistribution` replaces the two outer bars with half-normals
+    scaled so that half of the outer bucket's mass falls inside the bucket and
+    half falls beyond the outermost border. Distance is measured inwards-to-
+    outwards from the bucket's *inner* border, so this returns 1.0 at distance
+    0, 0.5 at one bucket width, and decays to 0 further out.
+
+    Written with `erfc` and not a literal `1 - HalfNormal.cdf(...)`: the
+    subtraction cancels to exactly zero a few sigma out, which is the very
+    underflow the tails are here to avoid.
```
jmetzen-pl: TODO: make docstring more concise

src/tabpfn/utils.py
```diff
@@ -373,28 +408,163 @@ def _cdf(logits: torch.Tensor, borders: torch.Tensor, ys: torch.Tensor) -> torch
...
+# the bar distribution's NLL is infinite for a target landing in it. float64
+# moves that resolution to ~1e-16, below anything a 5000-bucket grid puts in a
+# single bucket.
+#
+# The result is cast back to the caller's dtype, which keeps every downstream
+# dtype unchanged. That still bounds what survives: float32 cannot hold a mass
+# below ~1e-45, so a head with a narrow enough kernel keeps losing its far
+# buckets here even though the differencing found them. Fixing that needs the
+# pooling in `regressor.py` to stay in log space (`logsumexp` over members)
+# rather than a wider dtype on this boundary.
```
jmetzen-pl: shorten comment

src/tabpfn/architectures/shared/bar_distribution.py
jmetzen-pl: _Generated by Claude Code_

Shortened in c31c01a3.

src/tabpfn/regressor.py
jmetzen-pl: _Generated by Claude Code_

Shortened in c31c01a3.

src/tabpfn/regressor.py
jmetzen-pl: _Generated by Claude Code_

Shortened in c31c01a3.

src/tabpfn/utils.py
jmetzen-pl: _Generated by Claude Code_

Shortened in c31c01a3.

src/tabpfn/utils.py
jmetzen-pl: _Generated by Claude Code_

Shortened in c31c01a3.

LeoGrin (dismissed):
LGTM, thanks!!

src/tabpfn/utils.py
```diff
@@ -354,28 +356,107 @@ def _map_to_bucket_ix(y: torch.Tensor, borders: torch.Tensor) -> torch.Tensor:
...
-    prob_in_bucket = probs.gather(index=y_buckets, dim=-1) * share_of_bucket_left
-    prob_left_of_ys = prob_left_of_bucket + prob_in_bucket
+    cdf = prob_below_bucket + prob_in_bucket * share_of_bucket_left
+    survival = prob_above_bucket + prob_in_bucket * (1.0 - share_of_bucket_left)
+
+    # Read the two outer buckets as their half-normal tails. The whole bucket
+    # is overwritten, not just the part outside `borders`, because the tail
+    # redistributes mass within the bucket too; `CDF(borders[1])` stays
+    # `probs[..., 0]`, so interior buckets are untouched. Masking sizes the
+    # tail transients by the `ys` in the outer buckets, not by all of `ys`.
```
jmetzen-pl: this comment can be shortened

changelog/1288.fixed.md
```diff
@@ -0,0 +1 @@
+Regression density and likelihood evaluation reports far fewer infinite NLLs for targets that fall in a narrow bucket or in a distribution's tail, and the two tails of the predictive distribution are now treated alike. Point predictions shift with it: by round-off for well-behaved targets, and measurably where an ensemble member's target transform narrows its grid inside the shared one, because that member's tail mass is now spread along its half-normal instead of piled into the single bucket straddling its border. With `eval_metric="nll"`, temperature tuning previously discarded every candidate whose holdout loss came out infinite and silently kept the default temperature, so a tuned NLL temperature may now differ.
```
jmetzen-pl: this summary can be shortened

src/tabpfn/utils.py
jmetzen-pl: _Generated by Claude Code_

Shortened to three lines in 48f6d8e0. The sentence about masking sizing the transients by the outer-bucket `ys` is gone — it justified a memory choice, not the result — and the rest is tightened to the one non-obvious point: the whole outer bucket is overwritten, yet interior buckets do not move.

changelog/1288.fixed.md
jmetzen-pl: _Generated by Claude Code_

Shortened to three sentences in 48f6d8e0: fewer infinite NLLs, point predictions move (round-off usually, measurably for skewed targets), a tuned `eval_metric="nll"` temperature may differ. The mechanism behind each — which member grid narrows, where the tail mass used to pile up — is only in the PR description now.

jmetzen-pl:
Hi @LeoGrin, three commits landed on #1288 after your LGTM, so it's worth a second look at just these. Main reason was that the earlier implementation had a non-negligible performance penalty that is now (mostly) fixed and issues on ROCm. CI is green and the PR body's numbers are refreshed.

 - c121fb54: ROCm ships no float64 erfc: on an MI250X predict crashed outright with HIP error 209, not just ran slow.  Now computes it as erfcx(z) * exp(-z²), which agrees to 5e-14.
- 4dcf68b1: the remap was recomputing its row-independent half per row (the searchsorted, the bucket index, the tail masks). Hoisted out: bitwise identical, ~1.5–2× faster, and peak memory is now below main.
- c6d626fb: half the default ensemble remaps onto the grid it is already on, so that case now returns the softmax directly. Predictions move by ≤2.4e-6 sd.

Net effect: 10000 test rows on an A100 cost +8% over main, rather than the ~2–3× the remap itself costs. 

Let me know if you still agree to merging this.

PR #1168, opened 2026-08-10
Split FA3's KV sequence for short query chunks

jmkuebler:
## Motivation and Context

Stacked on #1165 (registry refactor); this is the first behaviour change on the FA3 path.

FA3 can do context parallelism, which is very helpful to saturate the hardware if there are few queries. FA3 has a heuristic to decide into how many chunks to split the KV, which is triggered by `num_splits=0`. Although today we do not compile, this heiristic would break compilation, so for now, I'd resort to a simple coarse and hand-designed heuristic. 
Therefore we just use 3 buckets. From LLM work I know that 32 splits is often pretty good so we do

The FA3 kernel only ever get's engaged in KV > 10K so we them implemented rule does:

```
If 10K < KV < 50K:
  if Q <=1024:
    num_splits = 32
  else: 
    num_splits = 1 (what we use today)
    
If KV > 50K:
   if Q <= 4K:
     num_splits = 32
   else:
     num_splits = 1
     
else:
 num_splits = 1
```

Measurements for different num_splits on H100 (ignore the rule column)

```
v3 ICL cross-attn: 8:1 MQA head_dim=64 batch=1 bf16 | rule: split when S_q <= 1024

MQA 8:1, S_kv=10,000
     S_q   splits=1  splits=32   ratio     accum   rel_err  rule
       1      0.184      0.086    2.13        0M  2.78e-03  split
      32      0.180      0.082    2.19        2M  2.84e-03  split
      64      0.180      0.082    2.18        4M  2.86e-03  split
     128      0.179      0.083    2.17        8M  2.87e-03  split
     256      0.181      0.095    1.90       16M  2.88e-03  split
     512      0.174      0.118    1.48       32M  2.88e-03  split
    1024      0.183      0.175    1.05       64M  2.89e-03  split
    2048      0.179      0.282    0.63      128M  2.88e-03  single
    4096      0.297      0.500    0.59      256M  2.88e-03  single

MQA 8:1, S_kv=20,000
     S_q   splits=1  splits=32   ratio     accum   rel_err  rule
       1      0.294      0.085    3.47        0M  2.77e-03  split
      32      0.288      0.086    3.33        2M  2.92e-03  split
      64      0.288      0.081    3.53        4M  2.87e-03  split
     128      0.286      0.090    3.16        8M  2.92e-03  split
     256      0.287      0.107    2.69       16M  2.90e-03  split
     512      0.276      0.140    1.98       32M  2.89e-03  split
[...]

PR #1291, opened 2026-09-18
Replace the FA3 attention backend with FlashAttention-4 (Hopper and Blackwell)

### LINKED ISSUE #1235: FlashAttention-4 backend for Blackwell
`fa3_backend.py` notes that FA3 is sm_90-only and that Blackwell will need FA4. FA4 is now available (`pip install flash-attn-4`, CuTeDSL, covering Hopper and Blackwell), and we'd like to contribute a backend for it. That means one new module in the shape of `fa3_backend.py`, implementing the `AttentionBackend` protocol from #1165, with no changes to the architectures themselves.

The shapes line up: v3's ICL head_dim of 64 and its fp16/bf16 dtypes are inside what FA4 supports on SM100 (`8 ≤ head_dim ≤ 128`, per `_validate_head_dims` in `flash_attn/cute/interface.py` as of `fa4-v4.0.0.beta28`), and `flash_attn.cute.flash_attn_func` takes q/k/v in the same `(B, S, H, D)` layout as FA3, so `run()` is close to identical. One caveat: that 64 comes from the v3 config defaults rather than a released checkpoint, so if the shipped models differ, worth knowing.

### Questions

Before we write anything:

1. **One backend or two?** FA4 covers Hopper too (head_dim up to 256 on sm_90). Separate `fa4_backend.py` beside FA3, or a single backend spanning both with FA3 eventually retired?
2. **Does a pre-release dependency belong in-tree?** FA4 is still on a weekly beta cadence. Registration is gated on importability, so a missing wheel just degrades to SDPA, but carrying a beta in-tree is your maintenance burden rather than ours. If that's the sticking point, we're equally happy to ship it out-of-tree as a separate package that registers itself, which is what #1165 makes possible.

### Benchmarking

`_FA3_MIN_SEQLEN_FOR_SPEEDUP` was measured on H100 with the FA3 kernels, and we'd rather not assume it carries over to a different kernel on different hardware. We're happy to benchmark FA4 against SDPA across a sequence-length sweep on v3's ICL shapes and post the crossover, including if it shows FA4 doesn't help. That needs no changes to the codebase, just the shapes.

tolleybot:
## Issue

Closes #1235.

## Motivation and Context

`fa3_backend.py` is Hopper-only and noted that Blackwell would need FlashAttention-4. FA4 is
now on PyPI (`flash-attn-4`, CuTeDSL, weekly betas) with kernels for sm_90 and sm_100+, so one
backend can serve both. Per the discussion on #1235, this PR replaces FA3 with FA4 rather than
adding it alongside.

What changes for a user:

- **Hopper:** FA4 matches or beats FA3 at every measured point (4–10% whole-model from 10k rows
  up, tie below) and, unlike FA3, is `pip install "tabpfn[fa4]"` rather than a 30–60 minute
  source build.
- **Blackwell:** attention dispatch to a fused kernel for the first time. 1.12–1.26× on
  `predict()` at ≥30k rows with ≥100 features.
- **No sequence-length threshold.** FA3's `_FA3_MIN_SEQLEN_FOR_SPEEDUP = 10_000` existed because
  its fixed `num_splits=32` made it 3–6× slower than SDPA at 1k rows. FA4 has no such penalty
  on either architecture (within noise from 100 rows, ahead from ~3k), so `is_preferred` is
  eligibility alone.
- **Wider head-dim eligibility.** FA4 accepts any multiple of 8 (8–256 on sm_90, 8–128 on sm_100),
  so it also takes the head_dim-16 feature-attention stages FA3 could not. On Blackwell this is
  where all the gain is: torch's SDPA picks cuDNN attention there, which already runs the
  head_dim-64 ICL calls at FA4 speed.
- **bf16 stays on SDPA on Blackwell**, with a one-time `UserWarning`. FA4's bf16 forward on sm_100
  is 18–23% slower than SDPA from ~10k rows (fp16 is not; Hopper is fine in both). TabPFN
  defaults to fp16 autocast on CUDA, so this only affects `inference_precision=torch.bfloat16`.

Measurements (H100 and GB200, torch 2.14.0+cu130, flash-attn-4 4.0.0b30) are in
[#1235](https://github.com/PriorLabs/TabPFN/issues/1235#issuecomment-5671366925) and
summarised in `fa4_setup.md`.

**Split-KV on Hopper** (asked on the issue): FA4 implements it on sm_100/110 only, so on sm_90 a short-Q call against a long KV at batch 1, the cached-prediction shape, fills only a handful of SMs: 9.9 ms for (1, 16, 8, 64) × (1, 1M, 1, 64) against 0.4 ms for FA3's 32-way split. The backend now splits outside the kernel on such calls (`_split_kv_plan`, `_fa4_split_kv`): KV chunks folded into the batch dimension for one launch, partial outputs combined with the LSE FA4 returns. Cached prediction (`fit_with_cache`, then predict 16–256 rows) on H100: 27–28 ms at 100k and 300k rows, against 23–25 ms with FA3 and 37–92 ms with SDPA or unsplit FA4.

### Implementation

- `shared/fa4_backend.py` (new): same shape as `fa3_backend.py`. Lazy cached import of
  `flash_attn.cute.flash_attn_func` so `is_available()` is one cheap check at registration;
  ROCm guard; per-architecture head-dim table; bf16 gate on cc ≥ 10. Three FA4 4.0.0b30
  specifics it handles: `flash_attn_func` returns `(out, lse)` unconditionally; split-KV is
  sm_100/110 only (sm_12x accepts `num_splits=1` only); the kernel launches one grid entry per
  batch element, so batches above 65,535 are chunked.
- `shared/scaled_dot_product_attention.py`: consult order `(FA4, torch-MPS, MLX)`.
- `shared/fa3_backend.py`, `shared/fa3_setup.md`: removed.
- `shared/fa4_setup.md`: install, dispatch conditions, measured speedups on both GPUs, bf16 rule.
- `pyproject.toml`: `fa4` and `fa4-cu13` extras (`flash-attn-4>=4.0.0b30`; the pre-release floor
  lets pip resolve the beta without `--pre`; `[cu13]` selects the CUDA 13 CuTeDSL libs). New
  `blackwell` pytest marker; `hopper` now means sm_90 + flash-attn-4.
- Tests: FA4 twins of the FA3 dispatch and equivalence tests, plus long-KV cross-attention,
  batch > 65,535, `is_preferred == eligibility`, and the bf16 gate/warning with a mocked compute
  capability so it runs anywhere. FA3-only tests removed.

**`torch.compile`:** FA4's `flash_attn_func` is not Dynamo-traceable (FA3's was), so with `enable_torch_compile=True` each FA4 call is a graph break inside `flash_attn/cute/`. Measured on v3 `predict()` on H100: within noise at 30k rows, ~4% at 100k, with FA4 still 1.17× faster than SDPA under compile. `enable_torch_compile` is off by default.

One thing found on the way, unrelated to this change: FA3's split-KV path crashed the process at
batch ≥ 65536 for any `num_splits > 1` (CUDA grid limit), which `fa3_attn_func` used for
`seq_q ≤ 1024` since #1168, so `test__fa3_batch_heads_above_cuda_max_grid` had been failing
since then without CI noticing. Moot once FA3 is gone.

---

## Public API Changes

-   [x] No Public API changes

`fa3_backend` was an internal module. Dispatch stays automatic and falls back to SDPA when
`flash-attn-4` is not installed; there is no user-facing switch. Installing FA3 no longer has any
effect.

---

## How Has This Been Tested?

- `tests/test_architectures/test_attention_backends.py`,
  `tests/test_architectures/test_shared/test_attention_backends.py`,
  `tests/test_architectures/test_attention_dispatch.py`: 27 pass on an H100 (sm_90) and on a
  GB200 (sm_100), Python 3.11 / 3.12, torch 2.14.0+cu130, flash-attn-4 4.0.0b30.
- FA4 matches SDPA within the FA3 tolerance (`atol=rtol=5e-3`) on the six canonical shapes ×
  fp16/bf16, on (1, 256, 100k, 8, 1, 64), and on batch=70,000, on both architectures.
- On the GB200 the bf16 gate was checked against the real device: bf16 declined with one
  warning, fp16 preferred at every sequence length.
- Whole-model and kernel-level benchmarks on both GPUs: see #1235. Cached prediction with few test rows measured separately (table above).
- `ruff check` / `ruff format` clean at the pinned 0.15.12; `mypy` clean on the changed files.

---

## Checklist


jmkuebler:
Thank you for starting this contribution! I will also run a few test on my end.

> Split-KV on Hopper (asked on the issue): not supported by FA4. The sm_90 forward asserts
not is_split_kv in b30, in b31 (current), and on main
([interface.py:1319](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L1319));
only sm_100/110 implement it, and there is no open issue or PR for sm_90. In the sweep this costs
2–9% against FA3 on the short-Q / long-KV shape at n_train=1M only; FA4 is ahead below that
(1.10× at 100k) and it does not show whole-model. _num_splits_for already asks FA4's heuristic
on sm_100/110, so if sm_90 support lands it is a one-line change.

could you please share the numbers for this experiment? in particular which "short-Q" ranges did you check?

PR #1293, opened 2026-09-19
Pin all GitHub Actions to commit SHAs and update outdated versions

eddiebergman:
This PR addresses some tag vs. SHA issues in the repo, which extended into some other good github action hygiene that are more "pressing", namely scoping permissions and access to SECRETS a bit more finely.

* Test failures seem to be that (rightfully) a fork does not have access to SECRETS required to download models, not sure how this is normally handled.

* Worth considering *Settings → Actions → General → "Require actions to be pinned to a full-length commit SHA"* so there's not a mismatch of tags and SHAs around the place.

---

# Claude >>

## Issue
N/A (CI maintenance).

## Motivation and Context
- Pin the remaining tag-referenced actions to full commit SHAs (`actions/checkout`, `actions/setup-python`, `astral-sh/setup-uv`, `scientific-python/action-towncrier-changelog`), so every `uses:` in `.github/workflows` is now SHA-pinned with a version comment.
- Bump outdated pins: `astral-sh/setup-uv` v7.3.0 to v10.1.0, `jlumbroso/free-disk-space` v1.3.1 to v2.0.0, `anthropics/claude-code-action` v1.0.224 to v1.0.231.
- Fix misleading comments: `actions/checkout` was pinned to the v7.0.1 SHA but commented `# v4.3.1`.

Note: `setup-uv` and `free-disk-space` cross major versions. The `with:` inputs in use look unchanged, but CI on this PR is the check. Also note no uv version is pinned anywhere in the repo, so the installed uv is still "latest" (unchanged by this PR).

---

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
Verified each SHA against the tag via the GitHub API. Not yet exercised by CI (this PR is the first run).

---

## Checklist

-   [ ] The changes have been tested locally.

CI-only change: "no changelog needed" label requested.


.github/workflows/ci.yml
```diff
@@ -329,7 +329,7 @@ jobs:
       # because its install footprint comfortably fits.
       - name: Free runner disk space
         if: matrix.repo-name == 'tabpfn-time-series'
-        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+        uses: jlumbroso/free-disk-space@ceedf095f4ec1a097402bc6bd80831f2e1a6fde6 # v2.0.0
```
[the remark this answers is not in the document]
eddiebergman: Outdated information in cursors view of the web.

https://github.com/jlumbroso/free-disk-space/releases/tag/v2.0.0

.github/workflows/ci.yml
```diff
@@ -20,7 +23,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4.3.1
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
```
eddiebergman, line 28: `persist-credentials: false`, nothing else (as far as I can tell needs git credentials stored)

.github/workflows/ci.yml
```diff
@@ -60,20 +65,18 @@ jobs:
             python-version: "3.14"
             dependency-set: highest
     runs-on: ${{ matrix.os }}
-    env:
-      HF_TOKEN: ${{ secrets.HF_TOKEN }}
-      TABPFN_TOKEN: ${{ secrets.TABPFN_TOKEN }}
```
eddiebergman, line 65: Moved to explicit steps, rather than all steps having access

.github/workflows/release-create-pr.yml
```diff
@@ -22,37 +22,38 @@ jobs:
...
+      BASE_REF: ${{ inputs.base_ref }}
+      BASE_BRANCH: ${{ inputs.base_branch }}
+      RELEASE_BRANCH: release/v${{ inputs.version }}
     steps:
       - name: Guard base_ref, must be a full 40-char commit SHA
         run: |
-          echo "${{ inputs.base_ref }}" | grep -Eq '^[0-9a-f]{40}$' || {
-              echo "base_ref must be a full 40-char commit SHA (got: ${{ inputs.base_ref }})"
+          echo "${BASE_REF}" | grep -Eq '^[0-9a-f]{40}$' || {
+              echo "base_ref must be a full 40-char commit SHA (got: ${BASE_REF})"
```
eddiebergman, line 36: The workflow itself is only exposed to trusted actors but good hygiene, as `${{ ref }}` is an exploitable pattern.

https://www.kenmuse.com/blog/the-hidden-danger-in-git-ref-names/#how-a-workflow-turns-this-into-execution

.github/workflows/release-publish.yml
```diff
@@ -46,10 +48,10 @@ jobs:
           PY
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@bec219d24cd3e171d82865faccec33120bb574f4 # v10.1.0
         with:
           version: "latest"
-          enable-cache: true
+          enable-cache: false
```
eddiebergman, line 54: Just a pre-caution to protect against poisoning in the published release artifact.

*if* another action allowed the cache to become poisoned with a bad package/code, then it's possible to inject arbitrary code into the published artifact that people would download. As far as I can tell, this isn't possible currently, but just some safety layers :)

.github/dependabot.yml
```diff
@@ -17,3 +17,7 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 3
+    # Match the pip quarantine above: only propose releases that have been public
+    # for a while, so a compromised release is likely caught before we adopt it.
+    cooldown:
+      default-days: 8
```
eddiebergman, line 23: Good to have explicit SHAs for versions, but the window for the linked _Trivy action_ compromise that was big, lasted about 8 hours and I was lucky to happened duing US time to not get bitten by it.

8 hours is pretty fast to detect + fix, so just like Python packages, good to have this extra cooldown

https://www.reddit.com/r/devsecops/comments/1rzjuev/your_security_scanner_was_the_attack_vector_trivy/

.github/workflows/nightly-notebook-check.yml
```diff
@@ -28,15 +28,17 @@ jobs:
...
+          persist-credentials: false
 
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        uses: astral-sh/setup-uv@bec219d24cd3e171d82865faccec33120bb574f4 # v10.1.0
```
oscarkey, line 41: Claude points out that this new version of uv disables cache pruning, which blows up the cache size: https://github.com/astral-sh/setup-uv/pull/967
This will cause our github actions cache to overflow. Would you be able to update all of the setup-uv calls so they only save the cache on merges to main?

oscarkey (changes requested):
thanks! Good idea

Re the tests, they work on fork like this: the CI running on main populates the github actions cache with the required checkpoints, so the forks don't need to downlaod the checkpoints, so don't need the tokens. That didn't work for this PR because the cache was overflowing, because we release 3.5 and it has larger checkpoint files. I manually fixed that for now and also made [this](https://github.com/PriorLabs/TabPFN/pull/1297) to improve the situation.

Re "Require actions to be pinned to a full-length commit SHA", we should turn this on the the org level. Just need to work out what happens to existing workflows that violate it.

Re the changelog, you can apply the "no changelog needed" label for PRs like this that users of the public package don't need to know about.

.github/workflows/nightly-notebook-check.yml
eddiebergman-priorlabs: CLAUDE: 
Done in dcc6a2c: every caching `setup-uv` step now sets `prune-cache: true` (v10 turned pruning off by default, which is the actual cause of the size blow-up) and `save-cache: ${{ github.ref == 'refs/heads/main' }}`, so PR runs only restore the cache that main produced. The two `release-publish` steps already had `enable-cache: false`, so they were left alone.

oscarkey: thank you! I'm wondering if we should set `prune-cache: false`, based on [this comment](<https://github.com/astral-sh/setup-uv/issues/745#issuecomment-3867334064>) from a PyPI dev, esp as we have quite a lot of traffic on this repo. That will be 10s of GBs in the cache, but that's okay if it's only on main.

eddiebergman-priorlabs: Update in 8b9e5a6: dropped the `prune-cache: true` again so main saves the full (unpruned) cache. Every caching `setup-uv` step now has just `save-cache: ${{ github.ref == 'refs/heads/main' }}`: only pushes to main write the cache, everything else restores it if present but never saves.

PR #1295, opened 2026-09-20
Run n_preprocessing_jobs > 1 on threads instead of processes

LeoGrin:
Generated by Claude Code

`n_preprocessing_jobs > 1` now runs the per-estimator preprocessing on threads instead of a loky process pool. The pool copied the training table into every worker and every transformed table back, so it was slower than `n_preprocessing_jobs=1` at every table size and paid 1.5 to 3 s per pool start; threads share the data and release the GIL inside numpy and scikit-learn, so they are never slower and 2x faster on large tables. The default stays 1 and the preprocessed outputs are bitwise unchanged.

## Issue

No issue. Discussed in Slack.

## Motivation and Context

joblib's default backend is loky. For eight preprocessing tasks that each take the whole table and return a transformed copy, that means the table is memmapped to a temp file on every call and eight transformed tables are pickled back through pipes, plus the pool spawn, which repeats across consecutive fits in one process. The docstring already recommended 1 because of this; the parameter name still invites the CPU count, and a caller who passes it pays seconds per fit for nothing.

<details>
<summary>Measurements</summary>

Fit only (`fit_mode="fit_preprocessors"`, no forward pass), 8 estimators, seconds per fit with the pool warm; each process spawn adds 1.6 to 2.7 s on top. Laptop, 18 cores, tabpfn 8.5.0 (same preprocessing code path as main).

| table | 1 job | 4 processes | 8 processes | 4 threads | 8 threads |
|---|---|---|---|---|---|
| adult, 32K x 14, 12 categorical | 0.21 | 0.40 | 0.45 | 0.19 | 0.20 |
| 100K x 30, 5 categorical | 0.34 | 0.48 | 0.48 | 0.23 | 0.28 |
| 500K x 30 | 1.7 | 3.7 | 3.8 | 0.79 | 0.70 |
| 1M x 30 | 3.3 | 14.5 | 14 | 1.45 | 1.3 |

Transfer alone, eight no-op tasks that return a float32 table of the input's shape: 1M x 30 rows cost 3.6 s through processes and 0.06 s through threads.

RTX Pro 6000, tabpfn 9.0.0, TabPFN-3.5 checkpoint, 8 estimators, 10K test rows, medians of 3: threads are 0.13 to 0.16 s faster on fit at 30K and 100K rows, the warm process pool 0.1 s slower; `low_memory` predict, where preprocessing overlaps the GPU forward, is 33.5 s with 8 threads against 33.7 s sequential at 100K rows, so no GIL contention shows. Predictions are bitwise identical across sequential, 8 threads and 8 processes in both fit modes.

This branch, 300K x 30 with 5 categoricals, CPU: fit 1.05 s with 1 job, 0.52 s with 4 threads, 0.47 s with 8 threads; the preprocessed `X_train` of all 8 ensemble members is bitwise equal to the 1-job result.

Few cores, to check that threads do not oversubscribe the BLAS threads inside each transform. Cluster CPU nodes, tabpfn 9.0.0, TabPFN-3.5 checkpoint, 8 estimators, threads, fit only, medians of 3. The 2-core row is a 4-core allocation pinned with `taskset -c 0-1`.

| cores | rows | 1 job | -1 (one thread per core) | 8 threads |
|---|---|---|---|---|
| 2 | 100K | 2.02 s | 1.70 s | 1.90 s |
| 2 | 1M | 15.9 s | 10.9 s | 10.8 s |
| 4 | 100K | 1.90 s | 1.53 s | 1.55 s |
| 4 | 1M | 12.0 s | 5.4 s | 5.3 s |
| 11 | 100K | 1.88 s | 1.45 s | 1.45 s |
| 11 | 1M | 12.3 s | 4.3 s | 4.1 s |

Threads win at every core count, 1.2x at 2 cores and 2.9x at 11 cores on 1M rows, and 8 threads on 2 cores is still faster than sequential. `-1` and `8` are the same because joblib never runs more threads than there are tasks. This PR keeps the default at 1; the numbers support `-1` as the default in a follow-up.

</details>

## Public API Changes

-   [x] No Public API changes

The parameter keeps its name, type and default. Its docstrings now describe threads and no longer warn against values above 1.

## How Has This Been Tested?

`tests/test_preprocessing/test_ensemble.py` and `tests/test_preprocessing/test_inf_handling.py` pass (141 passed, 10 skipped). The measurements above, including the bitwise comparison of preprocessed outputs across job counts, were run against this branch.

## Checklist

https://claude.ai/code/session_01Bdbxfjr1zmiLpYMftxAdXg


changelog/1295.changed.md
```diff
@@ -0,0 +1 @@
+`n_preprocessing_jobs > 1` now parallelises the per-estimator preprocessing over threads instead of processes, so it speeds up the fit instead of slowing it down. The default stays 1 and predictions are unchanged.
```
oscarkey:
```suggestion
`n_preprocessing_jobs > 1` now parallelises the per-estimator preprocessing over threads instead of processes, which we have found to be faster. Parallelisation is turned off by default, as previously.
```
maybe?

oscarkey (dismissed):
nice!!

oscarkey:
and yes I agree with sticking with 1 for now

PR #1297, opened 2026-09-21
CI: Only save the model cache when the CI runs on main, not in PRs.

oscarkey:
With the release of 3.5, the model cache is now quite big (22 GB, due to duplicates per platform). This meant we got stuck in a bad state, where cache entries created by PRs are continuously evicting cache entries on main. In particular, this breaks PRs from forks, which don't have the TABPFN_TOKEN, so can't download the model checkpoints themselves, so rely on the cache on main being populated.

The cache created by PRs is only useful for PRs that add new models. This happens very rarely, so just disable it to save total cache size.


jmkuebler (approved):
thanks!

PR #1299, opened 2026-09-21
feat(kv-cache): store an "auto" cache on the grid its attention backend left

ggprior:
Lets `kv_cache_precision="auto"` store the KV cache on the grid an attention backend already rounded the keys and values to, instead of always keeping the computed dtype.

**Why.** A registered attention backend can quantize the K/V of the call it takes and write the rounded values back, so the rest of the layer (e.g. the test→train half of the ICL stage) attends over the same numbers. A KV cache built from that call has to keep those values, otherwise test-time attention reads keys the training rows never saw. Backends choose calls by shape, e.g. only above a context-length threshold, so the right storage dtype is known only after the call has run. A caller setting `kv_cache_precision` up front has to either copy the backend's selection rule or pick a dtype that is wrong on one side of that threshold.

**How**
- `AttentionBackend` can declare `kv_grid_dtype`: the dtype whose grid (one scale per KV head) it leaves `k`/`v` on. This is optional, like `consumes_quantized_kv`.
- The SDPA wrapper records the backend each call ran on, via `recorded_attention_backends()`, a context-scoped recorder that does nothing unless active and does nothing while compiling.
- The v3 and v3.5 ICL layers record the train→train call when building a cache. `KVCacheEntry.grid_dtype` is set when every recorded call ran on a backend declaring the same grid and the cached slice is one KV head wide; a per-tensor scale then reproduces the per-head one.
- `KVCacheEntry.at_storage_dtype` applies the resolution: an explicit `"int8"`/`"fp8"` wins, otherwise `"auto"` uses `grid_dtype` if set and the computed dtype if not.

**Compatibility.** No behaviour change without a backend that declares `kv_grid_dtype`, and none of the in-tree backends do. The default (`None` → int8) is unchanged.

**Tests.** Recorder and grid resolution in `test_attention_backends.py`. In `test_tabpfn_v3.py` / `test_tabpfn_v3_5.py`: `"auto"` follows a backend that took the call, keeps the computed dtype when it declined, an explicit dtype overrides it, and several cached heads keep the computed dtype.


src/tabpfn/architectures/shared/attention_backends.py
```diff
@@ -224,6 +234,50 @@ def find_attention_backend(
...
+
+
+def record_attention_backend(backend: AttentionBackend | None) -> None:
+    """Note that a call ran on *backend*, if a recorder is active."""
+    if torch.compiler.is_compiling():
+        return
+    calls = _recorder.get()
+    if calls is not None:
+        calls.append(backend)
+
```
jmkuebler: is it on purpose to have this debugging logic on main with this PR?

jmkuebler (commented):
Thanks a lot @ggprior . The overall design makes sense to me, but IIUC the fallback then would be do FP16/BF16 usually and not INT8. 
We should make it such that if the attn backend does not expolicitly declare a quantized dtype that we still fall back to the current package default of `int8`.

So i'd say let's keep the existing auto behavior as well as the package default int8 and add a new option, say `adaptive` that sets quantized dtypes if enforced by the backend and otherwise falls back to int8

jmkuebler (dismissed):
Thanks you for working on this Georg! Looks good now

